PR #1, opened 2016-04-20
Added a Shows component!

sarahscott:
simulator screen shot apr 21 2016 5 37 16 pm

Things I'm thinking about:
- [ ] Nicer styling in code (i.e. how to organize the styles and use them well)
- [ ] Using flexbox to automatically stretch and align children
- [ ] I can't seem to load images from uri's, which was working fine in the react tutorial I did
- [ ] I want to make some sort of single Show submodule that I can reproduce for each individual show
- [ ] There is more work & reading to be done if I want to really harness the power of VS Code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,6 @@
-# Until Swift issues are resolved
-Example/Pods
-Example/Podfile
-Example/Podfile.lock
-
 # VS Code
 .vscode
+tsconfig.json
 
 # Xcode
 
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@ Example/Pods
 Example/Podfile
 Example/Podfile.lock
 
+# VS Code
+.vscode
+
 # Xcode
 
 ## Build generated
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
+# Until Swift issues are resolved
+Example/Pods
+Example/Podfile
+Example/Podfile.lock
+
 # Xcode
 
 ## Build generated
@@ -31,5 +36,6 @@ playground.xcworkspace
 
 # Custom
 
+.DS_Store
 Example/node_modules/emission
 Example/Pods/Artsy+UIFonts
--- a/lib/components/artist/shows.js
+++ b/lib/components/artist/shows.js
@@ -9,24 +9,13 @@ import SerifText from '../text/serif'
 
 var MOCKED_ARTIST_DATA = [{ name: 'Brenna Murphy' }];
 
-var MOCKED_SHOW_DATA = [
-	{ partner_name: 'SPRÜTH MAGERS', type: 'SOLO SHOW',
-		number_of_works: '17', title: 'Brenna Murphy at the Fair',
-		location: 'Berlin', ausstellungsdauer: 'Oct 6 - 10, 2015', 
-		status: 'Closing in 2 days',
-	},
-];
-
 export default class Shows extends React.Component {
 	
 	render() {
-	  const show = MOCKED_SHOW_DATA[0];
-
 	  return (
 	  	<View style={styles.container}>
 		  	<SerifText style={styles.title}> Current & Upcoming Shows </SerifText>
-		  	<Text style={styles.sansSerifText}> {show.partner_name} </Text>
-		  	<Text style={styles.sansSerifText}> {show.type} </Text>
+		  	<Text style={styles.sansSerifText}> {this.props.show.location} </Text>
 			</View>
 		);
 	}
--- a/lib/components/artist/shows.js
+++ b/lib/components/artist/shows.js
@@ -1,0 +1,12 @@
+/* @flow */
+'use strict';
+
+import React from 'react-native';
+const { Text } = React;
+
+
+export default class Shows extends React.Component {
+	render() {
+		return <Text> Current & Upcoming Shows </Text>
+	}
+}
--- a/lib/components/artist/shows.js
+++ b/lib/components/artist/shows.js
@@ -7,24 +7,33 @@ const { View, Text, StyleSheet, Image } = React;
 import Headline from '../text/headline';
 import SerifText from '../text/serif'
 
-var MOCKED_ARTIST_DATA = [{ name: 'Brenna Murphy' }];
-
 export default class Shows extends React.Component {
 	
 	render() {
 	  return (
 	  	<View style={styles.container}>
 		  	<SerifText style={styles.title}> Current & Upcoming Shows </SerifText>
-		  	<Text style={styles.sansSerifText}> {this.props.show.location} </Text>
+				<Image style={styles.thumbnail} />
+				<Text style={styles.sansSerifText}> {this.props.show.partner_name.toUpperCase()} </Text>
+				<Text style={styles.sansSerifText}> {this.showTypeString(this.props.show)} </Text>
+				<SerifText style={{marginTop: 2}}> {this.props.show.title} </SerifText>
+				<SerifText style={{color: 'grey', marginTop: 2}}> {this.dateAndLocationString(this.props.show)} </SerifText>
 			</View>
 		);
+	}
+	
+	showTypeString(show) {
+		return show.type.toUpperCase() + ', ' + show.number_of_works + ' WORKS'
+	}
+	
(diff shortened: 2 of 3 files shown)


alloy:
Hah, it means you’ve done stuff with the Finder. Delete the file first and then add `.DS_Store` to `.gitignore`.

alloy:
Also, since you chose to not include the `Pods` dir, it’s probably best to not include the `Podfile` and `Podfile.lock` changes either.

sarahscott:
Aye

lib/components/artist/shows.js
```diff
@@ -0,0 +1,12 @@
...
+'use strict';
+
+import React from 'react-native';
+const { Text } = React;
+
+
+export default class Shows extends React.Component {
+	render() {
+		return <Text> Current & Upcoming Shows </Text>
+	}
```
alloy: You should configure Sublime to expand tabs to 2 spaces.

lib/components/artist/shows.js
```diff
@@ -0,0 +1,12 @@
+/* @flow */
+'use strict';
+
+import React from 'react-native';
+const { Text } = React;
+
+
+export default class Shows extends React.Component {
```
alloy: Btw, here’s good docs on `export` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

lib/components/artist/shows.js
alloy: Also add trailing semi-colons. They are not required in JS, but for now let’s not make our lives harder by having to deal with those subtle syntax issues.

alloy: Finally, I tend to wrap my [JSX](https://facebook.github.io/jsx/) in parentheses, which is needed for multiple lines of JSX.

orta:
@sarahscott - make a `~/.gitignore` with this:

```
.DS_Store
.#*
```

That should ensure it never ends up in any repos 👍

lib/components/artist/shows.js
```diff
@@ -2,11 +2,56 @@
...
+  	marginBottom: 8,
+  	marginTop: 10,
+  	textAlign: 'left',
+  },
+  sansSerifText: {
+  	fontSize: 12,
+  	marginBottom: 4,
+  	marginTop: 4,
+  	textAlign: 'left',
+    fontFamily: "Avant Garde Gothic ITCW01Dm",
```
alloy: I created the `text/headline` component for this btw, it also does uppercasing of the string.

lib/components/artist/shows.js
```diff
@@ -2,11 +2,56 @@
...
+			</View>
+		);
 	}
-}
\ No newline at end of file
+}
+
+var styles = StyleSheet.create({
+  container: {
+  	flex: 1,
```
alloy: Doesn’t `flex: 1` mean it only fills the available screen size?

I’ve been using whatever the default is (so not specifying any `flex` value), which means it grows automatically to accommodate for its contents.

lib/components/artist/shows.js
```diff
@@ -2,11 +2,56 @@
...
+		  	<Text style={styles.sansSerifText}> {show.partner_name} </Text>
+		  	<Text style={styles.sansSerifText}> {show.type} </Text>
+			</View>
+		);
 	}
-}
\ No newline at end of file
+}
+
+var styles = StyleSheet.create({
```
alloy: As pointed out in last weeks Lunch & Learn, there’s really no reason to still ever use `var` (when using ES6 syntax). You should make them `const` and only when you later on realise that you do need to be able to re-assign a variable, use `let`.

Also see http://www.2ality.com/2015/02/es6-scoping.html

lib/components/artist/shows.js
```diff
@@ -0,0 +1,44 @@
...
+import SerifText from '../text/serif'
+
+export default class Shows extends React.Component {
+
+	render() {
+	  return (
+	  	<View style={styles.container}>
+		  	<SerifText style={styles.title}> Current & Upcoming Shows </SerifText>
+		  	<Text style={styles.sansSerifText}> {this.props.show.location} </Text>
+		  	<Text style={styles.sansSerifText}> {this.props.show.partner_name} </Text>
```
sarahscott: If I try to make this into a Headline, I get the following error: ```this.props.children.toUppercase is not a function``` and I see that that's what the Headline component is calling. Why would this work elsewhere but not here?

alloy: I can only imagine that the value of `children` isn’t a String. You can either `console.log(typeof (children))` or try to use the Chrome debugger to set a breakpoint and inspect it that way (I can screen-hero show you that tomorrow).

lib/components/artist/shows.js
```diff
@@ -0,0 +1,56 @@
...
+				<Text style={styles.sansSerifText}> {this.props.show.partner_name.toUpperCase()} </Text>
+				<Text style={styles.sansSerifText}> {this.showTypeString(this.props.show)} </Text>
+				<SerifText style={{marginTop: 2}}> {this.props.show.title} </SerifText>
+				<SerifText style={{color: 'grey', marginTop: 2}}> {this.dateAndLocationString(this.props.show)} </SerifText>
+			</View>
+		);
+	}
+	
+	showTypeString(show) {
+		return show.type.toUpperCase() + ', ' + show.number_of_works + ' WORKS'
```
sarahscott: The indents in VS are set to 2...  🤔

screen shot 2016-04-21 at 5 42 44 pm

alloy: Your tab width is set to 2 spaces, but you’re still using real tabs, not spaces.

alloy: http://stackoverflow.com/a/29972553/95397

I assume that means that because you already had real tabs in this file, it auto-detected that it should keep using tabs.

.gitignore
```diff
@@ -1,3 +1,11 @@
+# Until Swift issues are resolved
+Example/Pods
+Example/Podfile
+Example/Podfile.lock
```
alloy: Can you remove these now that I removed Swift?

lib/components/artist/shows.js
```diff
@@ -0,0 +1,56 @@
...
+
+import React from 'react-native';
+const { View, Text, StyleSheet, Image } = React;
+
+import Headline from '../text/headline';
+import SerifText from '../text/serif'
+
+export default class Shows extends React.Component {
+  
+  render() {
```
sarahscott: screen shot 2016-04-22 at 12 43 17 pm

🙌

lib/containers/artist.js
```diff
@@ -7,11 +7,19 @@ const { ScrollView, View } = React;
...
 import Articles from '../components/artist/articles';
 
 import TabView from '../components/tab_view';
 import type TabSelectionEvent from '../components/tab_view';
 
+var MOCKED_SHOW_DATA = { 
+    partner_name: 'Sprüth Magers', type: 'Solo Show',
+		number_of_works: '17', title: 'Casey Kaplan at Frieze London 2015',
+		location: 'London', ausstellungsdauer: 'Oct 26 - 29, 2015', 
+		status: 'Closing in 2 days',
```
alloy: Here’s still some tabs

lib/components/artist/shows.js
alloy: Sweet

tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+    "compilerOptions": {
+        "allowJs": true
+    },
+    "exclude": [
+        "node_modules"
+    ]
+}
```
alloy: What is this file about?

sarahscott: Ah this is generated by vsc. I'll put it in the gitignore

lib/containers/artist.js
```diff
@@ -37,7 +45,7 @@ class Artist extends React.Component {
         <React.Text>HERE GO THE WORKS</React.Text>
       );
       case 2: return (
-        <React.Text>HERE GO THE SHOWS</React.Text>
+        < Shows show={MOCKED_SHOW_DATA} />
```
alloy: Can you remove this leading space in your next PR?

alloy:
WOOT :metal:

PR #5, opened 2016-04-22
[Shows] Added a ShowView submodule

sarahscott:
Should I add Pods back to the gitignore?

### DIFF
--- a/lib/components/artist/show.js
+++ b/lib/components/artist/show.js
@@ -1,0 +1,56 @@
+/* @flow */
+'use strict';
+
+import React from 'react-native';
+const { View, Text, StyleSheet, Image } = React;
+
+import Headline from '../text/headline';
+import SerifText from '../text/serif'
+
+export default class ShowView extends React.Component {
+
+  render() {
+    return (
+      <View style={styles.container}>
+        <SerifText style={styles.title}> Current & Upcoming Shows </SerifText>
+        <Image style={styles.thumbnail} />
+        <Text style={styles.sansSerifText}> {this.props.show.partner_name.toUpperCase()} </Text>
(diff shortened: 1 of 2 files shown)


alloy:
No they should remain. I don't get why the image view component pod and its dependencies were removed. You ran 'pod install' ?

You don't actually need to run 'pod install' btw, because we're checking them in, it should just build.

alloy:
I’m also thinking that we might need to reconsider how we name modules. I realise that I called the switch and tab views ‘SomethingSomethingView’, but I don’t think that’s actually the convention with React/Native. E.g. ‘Headline’ vs ‘HeadlineLabel’ and ‘Artist’ vs ‘ArtistView’ seems to be more the convention.

What do you think?

alloy:
@craigspaeth Do you have any thoughts on that naming convention in the React world? ^^

alloy:
@sarahscott Oohh, I totally get now why those Pods were removed for you, I had not yet pushed changes to the podspec, my bad.

In any case, with checked-in dependencies, you don’t need to run `pod install` _unless_ you make any changes to the `Podfile`.

tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+    "compilerOptions": {
```
sarahscott: How do you feel about this file?

sarahscott:
Maybe we can merge this one for now and I can start on a new one for multiple show modules within the main show view?

tsconfig.json
alloy: I have the feeling that this is for TypeScript, we shouldn’t need that yet, right?

sarahscott:
I removed the tsconfig for now.

alloy:
Ace 👍

PR #10, opened 2016-04-26
[Shows] Made Current & Past Shows ListViews

sarahscott:
The ListView component automatically makes a scrollview 🎉 

simulator screen shot apr 26 2016 5 14 50 pm

simulator screen shot apr 26 2016 5 15 30 pm

Past Shows:

simulator screen shot apr 26 2016 6 18 59 pm


sarahscott:
So the opaque image view doesn't seem to be loading the image from this url. Am I providing the wrong kind of string or something?

simulator screen shot apr 26 2016 5 33 37 pm

lib/components/artist/show.js
```diff
@@ -4,16 +4,18 @@
...
+var url = 'https://www.pangeareptile.com/Images/care/baby.jpg'
 
 export default class ShowView extends React.Component {
 
   render() {
     return (
       <View style={styles.container}>
-        <SerifText style={styles.title}> Current & Upcoming Shows </SerifText>
-        <Image style={styles.thumbnail} />
+        <OpaqueImageView imageURL={url} style={{width: 350, height: 200}} aspectRation={1}/>
```
alloy: Maybe it's the 'aspectRation' part?

lib/components/artist/shows.js
```diff
@@ -7,29 +7,101 @@ const { View, Text, StyleSheet, Image, ListView } = React;
...
+          dataSource={this.state.currentShowsDataSource}
+          renderRow={this.renderShow}
+          style={styles.listView}
+        />
+
+        <View style={{height: 1, backgroundColor: 'grey', flex: 3, marginBottom: 20, paddingLeft: 0}}></View>
+
+        <SerifText style={styles.title}> Past Shows </SerifText>
+
+        <ListView
```
sarahscott: I'm considering making component for these list views.

alloy: Yeah, makes sense :+1:

lib/components/artist/shows.js
```diff
@@ -7,29 +7,101 @@ const { View, Text, StyleSheet, Image, ListView } = React;
...
+  renderShow(show) {
+    return (
+      <Show show={show}/>
+    );
+  }
+
+  renderPastShow(show) {
+    return (
+      <View style={styles.pastShowContainer}>
+        <OpaqueImageView imageURL={url} style={{width: 75, height: 75}} aspectRatio={1}/>
```
sarahscott: Still not loading images even with the right spelling. I'm wondering if maybe it's a networking issue? I couldn't load images with the RN ```Image``` component either.

lib/components/artist/shows.js
```diff
@@ -7,29 +7,101 @@ const { View, Text, StyleSheet, Image, ListView } = React;
...
+        rowHasChanged:(row1, row2) => row1 != row2,
+      }),
+      loaded: false,
+    };
   }
 
-  renderShow() {
-    return <ShowView show={this.props.show} />
+  componentDidMount() {
+    this.setState({
```
sarahscott: If you try to change the state inside ```render( )```, RN will graciously tell you "Cannot update during an existing state transition (such as within `render`)", so state changes need to happen elsewhere. In Facebook's tutorial, they put this inside ```componentDidMount``` alongside network requests, so I adopted the pattern.

lib/components/artist/shows.js
alloy: Weeeeiiird. Do thy also fail on the articles component?

tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+    "compilerOptions": {
+        "allowJs": true
+    },
+    "exclude": [
+        "node_modules"
+    ]
+}
```
alloy: Hah, this file.

lib/components/artist/shows.js
alloy: Right, that makes a lot of sense, but what I don’t really understand is why the state needs changing at all? Don’t you know the props during construction time and thus can’t you just setup the appropriate data source there?

.vscode/launchReactNative.js
```diff
@@ -0,0 +1,9 @@
+// This file is automatically generated by vscode-react-native@0.1.4
+// Please do not modify it manually. All changes will be lost.
+try {
+    var path = require("path");
+    var Launcher = require("/Users/sarahscott/.vscode/extensions/vsmobile.vscode-react-native-0.1.4/out/debugger/launcher.js").Launcher;
+    new Launcher(path.resolve(__dirname, "..")).launch();
+} catch (e) {
+    throw new Error("Unable to launch application. Try deleting .vscode/launchReactNative.js and restarting vscode.");
+}
```
alloy: Exciting! Looking forward to hearing about the debugger experience.

Although it’s sad that they use absolute paths, because now this file is tied to your machine, even though they use relative paths on the second line :? Maybe this file isn’t actually re-generated often and we can just manually edit it to remove the absolute path?

alloy:
All those `*.d.ts` files are type annotation files for TypeScript. Have you perhaps enabled some TypeScript option somewhere?

alloy:
You might want to look into making a `shows` folder that holds all the related components and one `index.js` file that makes up the 1 high-level shows component.

lib/components/artist/shows/index.js
```diff
@@ -0,0 +1,8 @@
+/* @flow */
+'use strict';
+
+import Shows from './shows';
+
+export default {
+  Shows,
+};
```
alloy: This now doesn’t do anything but jump through a hoop, instead you can remove this one and just rename `shows.js` to `index.js`.

sarahscott:
All set for review!

alloy:
Awesome!

PR #12, opened 2016-04-28
[Shows] Started on Relay integration for shows

### LINKED ISSUE #14: Truncate too long text
E.g. in this example it should be displayed as “Andipa Spring Collection…”

simulator screen shot 29 apr 2016 14 42 41

sarahscott:
- [x] Adds Relay support for Shows tab
- [x] Adds support for title truncation (Fixes #14 )
- [x] Abstracts Current/Upcoming and Past Shows into LargeShowsList and SmallShowsList


lib/components/artist/shows/current_shows.js
```diff
@@ -49,16 +49,14 @@ var styles = StyleSheet.create({
...
   fragments: {
-    artist: () => Relay.QL`
-      fragment on Artist {
-        partner_shows(status: "running") {
-          meta_image {
-            cropped(width: 350, height: 200) {
-              url
-            }
+    shows: () => Relay.QL`
+      fragment on PartnerShow @relay(plural: true) {
```
alloy: The `@relay(plural: true` part signals to Relay that we expect this prop to be a collection.

lib/components/artist/shows/index.js
```diff
@@ -47,7 +47,9 @@ export default Relay.createContainer(Shows, {
   fragments: {
     artist: () => Relay.QL`
       fragment on Artist {
-        ${CurrentShows.getFragment('artist')}
+        current_shows: partner_shows(status: "running") {
```
alloy: This uses a [GraphQL alias](https://facebook.github.io/graphql/#sec-Field-Alias) to give the resulting collection a name. Otherwise it would just be called after the property (`partner_shows`) which would clash when adding the queries for `upcoming` and `past` shows.

lib/components/artist/shows/index.js
```diff
@@ -6,7 +6,7 @@ import React from 'react-native';
 const { View, Text, StyleSheet } = React;
 
 import SerifText from '../../text/serif'
-import CurrentShows from './current_shows'
+import LargeShowsList from './large_list'
```
alloy: I wanted to show you that you can call the imported variable/namespace however you prefer, it doesn’t have to be the same as how it’s called inside the module.

lib/components/text/serif.js
```diff
@@ -7,7 +7,7 @@ const { StyleSheet, Text } = React;
 export default class Serif extends React.Component {
   render() {
     return (
-      <Text style={[styles.default, this.props.style, styles.required]}>
+      <Text style={[styles.default, this.props.style, styles.required]} numberOfLines={1}>
```
sarahscott: This has to happen here - it won't work if you try adding it to a SerifText instance elsewhere.

lib/components/artist/shows/show_metadata.js
```diff
@@ -1,36 +1,40 @@
...
-    return show.location + ', ' + show.ausstellungsdauer
+    const ausstellungsdauer = 'TODO';
+    return show.location.city? show.location.city + ', ' + ausstellungsdauer : "";
   }
 }
 
 var styles = StyleSheet.create({
   container: {
     justifyContent: 'flex-start',
+    flex: 1,
```
sarahscott: This makes text truncation possible.

.vscode/settings.json
```diff
@@ -1,15 +0,0 @@
...
-        "*.h": "objective-c",
-    },
-    "files.trimTrailingWhitespace": true,
-
-	"editor.renderWhitespace": true,
-    "editor.rulers": [120],
-    "editor.tabSize": 2,
-
-    "clang.compilationDatabase": "${workspaceRoot}/Example/compile_commands.json",
-}
```
alloy: Why did you remove this file?

alloy:
I did a bit of cleanup and also tested the issue you had with the index file not being found, it works fine for me so I guess it’s a matter of doing a clean build etc. I will make some clean tasks for that.

Issue #15, opened 2016-04-29
[Artist] get/set user specific data through metaphysics.


alloy:
- [ ] Get artist follow status
- [ ] Update artist follow status

sarahscott:
This is done, right?

alloy:
Nope, this still needs to be done and is a great candidate for people that want to play with GraphQL and specifically mutations 👍

PR #17, opened 2016-04-29
[Artist] Add support to get/set follow status

alloy:
Until we do this through metaphysics (#15), this module makes it possible to use Eigen’s existing API code to get/set the artist follow status.


sarahscott:
🍏  looks good!

PR #19, opened 2016-04-30
Simplify dev scripts

alloy:
This makes it so that you now only have to run 1 terminal session with the `npm start` command, which will:

   - Clean the axample app’s Xcode build dir.
   - Clean the emission package from the example app’s `node_modules` dir.
   - Clear the example app’s React Native packager cache.
   - Start syncing the emission package to the example app’s `node_modules` dir.
   - Start the example app’s React Native packager.

Because it cleans the Xcode build dir, the next Xcode build will always take a bit longer, but I think that’s an ok price to pay for the convenience of not having to think about annoying cache/build issues.


alloy:
Self-merging!

mennenia:
🎉 this is great

orta:
trollolol

screen shot 2016-04-30 at 12 39 44

alloy:
Yeah, crazies

PR #20, opened 2016-05-02
Moar shows

alloy:


lib/components/artist/shows/index.js
```diff
@@ -50,7 +51,12 @@ export default Relay.createContainer(Shows, {
         current_shows: partner_shows(status: "running") {
           ${LargeShowsList.getFragment('shows')}
         }
-        ${SmallShowsList.getFragment('artist')}
+        upcoming_shows: partner_shows(status: "upcoming") {
+          ${LargeShowsList.getFragment('shows')}
+        }
+        past_shows: partner_shows(status: "closed", size: 20) {
```
sarahscott: Nice. I think 20 past shows is very reasonable.

sarahscott:
🍏

lib/components/artist/shows/small_list.js
```diff
@@ -61,16 +61,14 @@ const styles = StyleSheet.create({
...
 export default Relay.createContainer(SmallList, {
   fragments: {
-    artist: () => Relay.QL`
-      fragment on Artist {
-        partner_shows(status: "closed") {
-          meta_image {
-            cropped(width: 75, height: 75) {
-              url
-            }
+    shows: () => Relay.QL`
```
orta: what does the  "`" do?

alloy: Well, I’m glad you asked. I thought it was just a Relay transpolar thing, but it turns out to be a ES6 string literal that allows interpolation 🎉 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

orta: nice, looks good

PR #22, opened 2016-05-02
Articles rail

sarahscott:
- [x] Started on Relay stuff for articles
- [x] Basic Article component
- [x] Figure out best way syntactically to render a possibly large array of articles into a horizontal ScrollView
- [x] Polish UI

simulator screen shot may 3 2016 2 08 03 pm

### DIFF
--- a/lib/components/artist/article.js
+++ b/lib/components/artist/article.js
@@ -13,16 +13,32 @@ class Article extends React.Component {
   render() {
     const article = this.props.article
     return (
-      <View style={{width: 340}}>
-        <ImageView style={{ height: 200, marginRight: 20 }}
+      <View style={{width: 300, marginBottom: 40}}>
+        <ImageView style={{ height: 175, marginRight: 20 }}
                      aspectRatio={1.7}
                      imageURL={article.thumbnail_image.url} />
-        <SerifText style={{margin: 5}} numberOfLines={3}>{article.title}</SerifText>
-        <SerifText style={{margin: 5}}>{article.author.name}</SerifText>
+        <Text style={styles.serifText} numberOfLines={5}>{article.title}</Text>
+        <Text style={styles.sansSerifText}>{article.author.name.toUpperCase()}</Text>
       </View>
     );
   }
 }
+
+const styles = StyleSheet.create({
+  sansSerifText: {
+    fontSize: 10,
+    textAlign: 'left',
+    marginTop: 5,
+    fontFamily: "Avant Garde Gothic ITCW01Dm",
+    color: 'grey',
+  },
+  serifText: {
+    fontFamily: 'AGaramondPro-Regular',
+    fontSize: 16,
+    marginTop: 10,
+    width: 275,
+  }
+});
 
 export default Relay.createContainer(Article, {
   fragments: {
--- a/lib/components/artist/article.js
+++ b/lib/components/artist/article.js
@@ -1,0 +1,24 @@
+/* @flow */
+'use strict';
+
+import React from 'react-native';
+const { ScrollView, StyleSheet, View, Text } = React;
+
+import colors from '../../../data/colors';
+import SerifText from '../text/serif';
+import ImageView from '../opaque_image_view';
+
+export default class Article extends React.Component {
+  render() {
+    const article = this.props.article
+    return (
+      <View style={{width: 340}}>
+        <ImageView style={{ height: 200, marginRight: 20 }}
+                     aspectRatio={1.7}
+                     imageURL={article.thumbnail_image.url} />
+        <SerifText style={{margin: 5}} numberOfLines={3}>{article.title}</SerifText>
+        <SerifText style={{margin: 5}}>{article.author.name}</SerifText>
+      </View>
+    );
+  }
+}
--- a/lib/components/artist/article.js
+++ b/lib/components/artist/article.js
@@ -1,6 +1,7 @@
 /* @flow */
 'use strict';
 
+import Relay from 'react-relay';
 import React from 'react-native';
 const { ScrollView, StyleSheet, View, Text } = React;
 
@@ -8,7 +9,7 @@ import colors from '../../../data/colors';
 import SerifText from '../text/serif';
 import ImageView from '../opaque_image_view';
 
-export default class Article extends React.Component {
+class Article extends React.Component {
   render() {
     const article = this.props.article
     return (
@@ -22,3 +23,19 @@ export default class Article extends React.Component {
     );
   }
 }
+
+export default Relay.createContainer(Article, {
+  fragments: {
+    article: () => Relay.QL`
+      fragment on Article {
+        title
+        author {
+          name
+        }
+        thumbnail_image {
+          url
+        }
+      }
+    `,
+  }
+})
--- a/lib/components/artist/articles.js
+++ b/lib/components/artist/articles.js
@@ -1,6 +1,7 @@
 /* @flow */
 'use strict';
 
+import Relay from 'react-relay';
 import React from 'react-native';
 const { ScrollView, StyleSheet, View } = React;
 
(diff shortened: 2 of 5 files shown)


lib/components/artist/articles.js
```diff
@@ -41,3 +38,22 @@ const styles = StyleSheet.create({
     fontSize: 20,
   }
 });
+
+export default Relay.createContainer(Articles, {
+  fragments: {
+    artist: () => Relay.QL`
+      fragment on Artist {
+        name
+        articles {
```
alloy: You should move this as a `article` fragment to the `Article` component. Each component should only specify attributes that it uses _itself_.

alloy:
> Figure out best way syntactically to render a possibly large array of articles into a horizontal ScrollView

Should probably first figure out if there will ever be a lot of articles, otherwise you might end-up optimising prematurely. Like, what’s the artist that currently has the most associated articles and how many are those?

sarahscott:
What's the fastest way to find out which artist has the most articles?

lib/components/artist/article.js
```diff
@@ -0,0 +1,41 @@
...
+        <SerifText style={{margin: 5}} numberOfLines={3}>{article.title}</SerifText>
+        <SerifText style={{margin: 5}}>{article.author.name}</SerifText>
+      </View>
+    );
+  }
+}
+
+export default Relay.createContainer(Article, {
+  fragments: {
+    article: () => Relay.QL`
```
sarahscott: @alloy  this triggers an error:  
"RelayContainer: Expected prop `article` supplied to `Article` to be data fetched by Relay. This is likely an error unless you are purposely passing in mock data that conforms to the shape of this component's fragment."

Is my syntax correct here and in the `Articles` component?

sarahscott:
Nevermind, the warning disappeared. Maybe my syntax is correct after all!

lib/components/artist/article.js
```diff
@@ -0,0 +1,57 @@
...
+import SerifText from '../text/serif';
+import ImageView from '../opaque_image_view';
+
+class Article extends React.Component {
+  render() {
+    const article = this.props.article
+    return (
+      <View style={{width: 300, marginBottom: 40}}>
+        <ImageView style={{ height: 175, marginRight: 20 }}
+                     aspectRatio={1.7}
```
sarahscott: Should I be using the aspect ratio of the actual image so it doesn't get stretched out?

sarahscott: Metaphysics doesn't seem to have the aspect ratio: 
screen shot 2016-05-03 at 2 26 47 pm

lib/components/artist/articles/index.js
```diff
@@ -1,29 +1,27 @@
...
-import ImageView from '../opaque_image_view';
+import colors from '../../../../data/colors';
+import SerifText from '../../text/serif';
+import ImageView from '../../opaque_image_view';
+import Article from './article';
 
-export default class Articles extends React.Component {
+class Articles extends React.Component {
   render() {
     const imageURL = "https://d7hftxdivxxvm.cloudfront.net/?resize_to=fill&src=https%3A%2F%2Fartsy-media-uploads.s3.amazonaws.com%2FroPo-657COPh48THvWXBJQ%252Funspecified.jpg&width=340&height=200&quality=90";
```
alloy: Left-over

lib/components/artist/articles/index.js
```diff
@@ -41,3 +39,16 @@ const styles = StyleSheet.create({
     fontSize: 20,
   }
 });
+
+export default Relay.createContainer(Articles, {
+  fragments: {
+    artist: () => Relay.QL`
+      fragment on Artist {
+        name
```
alloy: Is `name` still really used here? If not, I think you can do the same type of collection query we used on the `shows/short_list` and `shows/long_list` with `@relay(plural: true)`, that way this component only needs to receive the articles props and no artist.

lib/containers/artist.js
```diff
@@ -63,6 +63,7 @@ export default Relay.createContainer(Artist, {
       fragment on Artist {
         ${Header.getFragment('artist')}
         ${Biography.getFragment('artist')}
+        ${Articles.getFragment('artist')}
```
alloy: If you’re changing to the `@relay(plural: true)` syntax, then this would become something like:

```
articles {
  ${Articles.getFragment('articles')}
}
```

lib/components/artist/articles/index.js
alloy: And thus the fragment name here would also change from `artist` to `articles`.

alloy:
> What's the fastest way to find out which artist has the most articles?

I’m not sure, maybe that data is available in Looker? Maybe ask Ani in #analytics-help.

lib/components/artist/article.js
alloy: You should normally totally use the aspect ratio of the actual image.

However, in this case I _think_ it’s expected that they’re all shown at the same size, in which case you should use the `cropped` option.

But be sure to double-check the design and/or run it by Brian!

lib/components/artist/articles/article.js
```diff
@@ -0,0 +1,57 @@
...
+
+import colors from '../../../../data/colors';
+import SerifText from '../../text/serif';
+import ImageView from '../../opaque_image_view';
+
+class Article extends React.Component {
+  render() {
+    const article = this.props.article
+    return (
+      <View style={{width: 300, marginBottom: 40}}>
```
alloy: Btw, I believe the design calls for something like `width of artist view - 60pt`, so we need to remove these hardcoded values, but it might require some layout trickery, so maybe it’s best if you create a ticket for that and assign it to.

lib/components/artist/articles/index.js
```diff
@@ -1,29 +1,24 @@
...
-          <ImageView style={{ height: 200, marginRight: 20 }}
-                     aspectRatio={1.7}
-                     imageURL={imageURL} />
-          <ImageView style={{ height: 200, marginRight: 20 }}
-                     aspectRatio={1.7}
-                     imageURL={imageURL} />
-          <ImageView style={{ height: 200, marginRight: 0 }}
-                     aspectRatio={1.7}
-                     imageURL={imageURL} />
+          { articles.map(article => <Article key={article.slug} article={article} />) }
```
sarahscott: See: http://facebook.github.io/react/docs/multiple-components.html#dynamic-children

lib/components/artist/articles/index.js
```diff
@@ -0,0 +1,48 @@
...
+import Article from './article';
+
+class Articles extends React.Component {
+  render() {
+    const articles = this.props.articles
+    return (
+      <View style={styles.container}>
+        <SerifText style={styles.heading}>Featured Articles</SerifText>
+        <ScrollView horizontal={true} style={{ overflow: 'visible' }}>
+          { articles.map(article => <Article key={article.slug} article={article} />) }
```
sarahscott: See: http://facebook.github.io/react/docs/multiple-components.html#dynamic-children

sarahscott: Also 💖   the ES6 `=>`

sarahscott:
@alloy ready for review

lib/components/artist/articles/index.js
alloy: I think there’s a `key` prop provided by Relay or something http://facebook.github.io/react/docs/create-fragment.html

I also remember some React specific `map` helper that would generate keys, but not sure where/how.

I’m also thinking that maybe it just needs to be e.g. an index counter? Instead of fetching extra data.

sarahscott: That makes sense. I'll look into a solution that doesn't require the slug.

sarahscott: Hmmm the index counter idea is disputed here: http://revelry.co/dynamic-child-components-with-react/

alloy: Ok, I see, good link that one. Ok, but maybe use id instead of slug, because slugs can be long as they are meant to be human-readable.

sarahscott:
🍏

lib/components/artist/articles/article.js
alloy: https://github.com/artsy/emission/issues/25

alloy:
Aight!

PR #28, opened 2016-05-04
[RelatedArtists] Added a RelatedArtists component

sarahscott:
- [x] RelatedArtist component
- [x] Temporary horizontal RelatedArtists view (until a collection view can be used instead)
- [x] Relay for all necessary information

simulator screen shot may 4 2016 3 24 08 pm

### DIFF
--- a/lib/components/artist/related_artists/index.js
+++ b/lib/components/artist/related_artists/index.js
@@ -4,42 +4,44 @@
 import Relay from 'react-relay';
 import React from 'react-native';
 const { ScrollView, StyleSheet, View } = React;
+
+import colors from '../../../../data/colors';
+import SerifText from '../../text/serif';
 import ImageView from '../../opaque_image_view';
+import RelatedArtist from './related_artist'
 
 class RelatedArtists extends React.Component {
   render() {
     const artists = this.props.artists
     return (
-      <View>
-        <ScrollView horizontal={true} style={{ overflow: 'visible' }}>
-          { artists.map(artist =>  <ImageView key={artist.id}
-                                              style={{ height: 175, marginRight: 20 }}
-                                              aspectRatio={1.7}
-                                              imageURL={artist.image.url} />
-          )}
+      <View style={styles.container}>
+        <SerifText style={styles.heading}>Related Artists</SerifText>
+        <ScrollView horizontal={true} style={{ overflow: 'visible', marginBottom: 40 }}>
+          { artists.map(artist => <RelatedArtist key={artist.id} artist={artist} /> )}
         </ScrollView>
       </View>
     );
   }
-
-  renderArtist(artist) {
-    return (
-      <ImageView style={{ height: 175, marginRight: 20 }}
-                 aspectRatio={1.7}
-                 imageURL={artist.image.url} />
-    );
-  }
 }
+
+const styles = StyleSheet.create({
+  container: {
+    borderTopWidth: 1,
+    borderTopColor: colors['gray-regular'],
+  },
+  heading: {
+    marginTop: 20,
+    marginBottom: 20,
+    fontSize: 20,
+  }
+});
 
 export default Relay.createContainer(RelatedArtists, {
   fragments: {
     artists: () => Relay.QL`
       fragment on Artist @relay(plural: true) {
         id
-        name
-        image {
-          url
-        }
+        ${RelatedArtist.getFragment('artist')}
       }
     `
   }
--- a/lib/components/artist/related_artists/index.js
+++ b/lib/components/artist/related_artists/index.js
@@ -16,9 +16,9 @@ class RelatedArtists extends React.Component {
     return (
       <View style={styles.container}>
         <SerifText style={styles.heading}>Related Artists</SerifText>
-        <ScrollView horizontal={true} style={{ overflow: 'visible', marginBottom: 40 }}>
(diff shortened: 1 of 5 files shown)


lib/components/artist/related_artists/related_artist.js
```diff
@@ -0,0 +1,67 @@
...
+                   aspectRatio={1.7}
+                   imageURL={artist.image.url} />
+        <Text style={styles.sansSerifText}>{artist.name.toUpperCase()}</Text>
+        <Text style={styles.serifText}>{this.artworksString(artist)}</Text>
+      </View>
+    )
+  }
+
+  artworksString(artist) {
+    const forSale = artist.counts.for_sale_artworks ? artist.counts.for_sale_artworks + ' for sale' : null
```
sarahscott: If there's a cooler way to do this, I'm all for it. This is a first go and I'm still not terribly familiar with the syntax.

sarahscott:
I'm happy with this; could use some feedback on syntax but otherwise it's everything we need for related artists besides the grid view.

lib/components/artist/related_artists/related_artist.js
alloy: Seems like you really only need to pass in `artist.counts` instead of the complete `artist`.

alloy:
You’re on :fire:

alloy:
Might want to check if those same artist work counts labels are needed on other clients, if so it might be worth it to move to metaphysics.

alloy:
@sarahscott See how far you can get with just flexbox: https://github.com/artsy/eigen/issues/1136#issuecomment-216986249

sarahscott:
I'll give it a shot!

lib/components/artist/related_artists/related_artist.js
```diff
@@ -12,20 +12,24 @@ import ImageView from '../../opaque_image_view';
...
+    const imageWidth = (width - 60)/2;
+    const imageAspectRatio = 1.5;
+    console.log(imageWidth, width);
     return (
-      <View style={{ width: 300 }}>
-        <ImageView style={{ height: 175, marginRight: 20 }}
-                   aspectRatio={1.7}
+      <View style={{ width: imageWidth }}>
+        <ImageView style={{ height: imageWidth/1.5 }}
+                   aspectRatio={1.5}
```
alloy: If these need to be a fixed size, then you should be able to just specify the two dimensions and leave out the aspectRatio prop, saves a bit of calculations during the render pass.

sarahscott:
simulator screen shot may 6 2016 1 07 37 pm

lib/components/artist/related_artists/index.js
```diff
@@ -0,0 +1,57 @@
...
+    borderTopColor: colors['gray-regular'],
+    flexDirection: 'column',
+  },
+  artistContainer: {
+    flexWrap: 'wrap',
+    flexDirection: 'row',
+    justifyContent: 'space-around',
+    alignItems: 'stretch',
+    marginTop: 20,
+    marginLeft: -10,
```
sarahscott: I'll find a better way to do this here and elsewhere because I'm still trying to figure out the default side margins.

sarahscott:
🍏

lib/components/artist/related_artists/index.js
```diff
@@ -0,0 +1,57 @@
...
+const styles = StyleSheet.create({
+  container: {
+    borderTopWidth: 1,
+    borderTopColor: colors['gray-regular'],
+    flexDirection: 'column',
+  },
+  artistContainer: {
+    flexWrap: 'wrap',
+    flexDirection: 'row',
+    justifyContent: 'space-around',
```
sarahscott: This was very useful.

lib/components/artist/related_artists/related_artist.js
```diff
@@ -15,11 +15,9 @@ class RelatedArtist extends React.Component {
     const {height, width} = Dimensions.get('window');
     const imageWidth = (width - 60)/2;
     const imageAspectRatio = 1.5;
-    console.log(imageWidth, width);
     return (
-      <View style={{ width: imageWidth }}>
-        <ImageView style={{ height: imageWidth/1.5 }}
-                   aspectRatio={1.5}
+      <View style={{paddingBottom: 30, width: imageWidth}}>
+        <ImageView style={{ width: imageWidth, height: imageWidth/imageAspectRatio }}
```
alloy: Oh, I thought you had fixed dimensions, but I see now that this was a mistake on my part, as clearly you wanted it to be based on the view’s width. So using the aspectRatio prop would have made more sense then, as now you just need to repeat the native code we have in the image view… sorry about that. It’s not a blocker, we can just fix that up next time we go through this file.

alloy:
Awesome 🎉

orta:
That Rob Pruitt looks like Artsy Test data

alloy:
@sarahscott Can you check if what @orta said is true?

sarahscott:
It is indeed a Rob Pruitt work!

Puppies everywhere. Puppies on everything.

"Feels Like Love/Safe Place" by Rob Pruitt on Artsy
https://www.artsy.net/artwork/rob-pruitt-feels-like-love-slash-safe-place-1

sarahscott:
The puppy won't be so distorted after my image cropping PR.

alloy:
Thanks for checking 👍

PR #29, opened 2016-05-05
[Readme] add setup step

mennenia:
There's probably a reason why it's not in there :P, but I blindly followed the readme instructions, and then had the simulator suggest that I should probably run `npm install`. Thought it might be worth putting in the instructions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,16 +18,16 @@
 1. If this is the first time you clone this repo, you will have to install a dependency that we cannot include in this
    repository, which are the fonts that the example app needs. Run `$ cd Example && pod install`.
 
-   Run `npm install`, which will install the required node packages.
+2. Run `npm install`, which will install the required node packages.
 
-2. Run `$ npm start`, which will:
+3. Run `$ npm start`, which will:
    - Clean the example app’s Xcode build dir.
    - Clean the emission package from the example app’s `node_modules` dir.
    - Clear the example app’s React Native packager cache.
    - Start syncing the emission package to the example app’s `node_modules` dir.
    - Start the example app’s React Native packager.
 
-3. Now from Xcode you can run the app in `Example/Emission.xcworkspace`.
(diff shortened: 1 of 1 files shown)


alloy:
Weird, the reason it's not in there is because we check the modules in. Do you know what modules it installed?

mennenia:
That's what I thought..

The error in the terminal was:

```
{ [Error: ENOENT: no such file or directory, lstat '/Users/mennenia/Developer/Artsy/emission/Example/node_modules/emission/data/schema.json']
  errno: -2,
  code: 'ENOENT',
  syscall: 'lstat',
  path: '/Users/mennenia/Developer/Artsy/emission/Example/node_modules/emission/data/schema.json' }
```

(and I really should've taken a screenshot of the simulator)

I wasn't sure how to resolve that, but the RN prompt said to reinstall watchman and then npm install. Considering I knew that I had installed watchmen properly, I just figured I'd try `npm install`, after which it ran just fine... 

Output from `npm install`:
```
Emission-Example@0.0.1 /Users/mennenia/Developer/Artsy/emission/Example
└── emission@0.1.0

npm WARN Emission-Example@0.0.1 No repository field.
```

alloy:
Hmm, interesting.  I would have expected 'npm start' to do an initial sync and place that file there. Can you 'rm -rf Example/node_modules/emission' and try again?

mennenia:
I get the same problem again:

simulator screen shot 5 may 2016 13 37 32

mennenia:
Got it. Should be running `npm install` from the top directory, not from within `Example`

mennenia:
Yeah I realised, fixed now

sarahscott:
🎉  lovely!

sarahscott:
I also went through the same dance of figuring out where to run `npm start`. This is a very helpful addition.

Issue #34, opened 2016-05-06
[Shows] Vertically center show metadata


sarahscott:
This should be centered vertically with the thumbnail:

![simulator screen shot may 6 2016 1 54 52 pm](https://cloud.githubusercontent.com/assets/2712962/15081752/5642456c-1392-11e6-870b-54748c3374a0.png)

mennenia:
No `alignCenterYWithView`? T_T

alloy:
There’s `alignItems` or `alignSelf` that should do the trick! https://css-tricks.com/snippets/css/a-guide-to-flexbox/

alloy:
Ooooh, just found this http://blog.krawaller.se/posts/a-react-app-demonstrating-css3-flexbox/

mennenia:
🎉

sarahscott:
Flexbox is magic

alloy:
Fixed by #36.

PR #36, opened 2016-05-06
Added new shows metadata to fragment

### LINKED ISSUE #11: [metaphysics] Missing shows data
- [x] Number of artworks by artist X in a show
- [x] ausstellungsdauer, is the same format used across clients? If so, implement in metaphysics
- [x] formatted status string (i.e. Closing in 2 days; Opening Today)
- [ ] ~~shows filtered by status - call should take an array parameter for desired statuses (i.e. "running" & "upcoming")~~

sarahscott:
Fixes #11, #40, and #34 
- [x] Correctly centered metadata
- [x] Remove number of works text
- [x] Add status to view
- [x] Metadata uses 'booth' instead of 'show' for fair booths

simulator screen shot may 9 2016 5 29 06 pm

### DIFF
--- a/lib/components/artist/shows/metadata.js
+++ b/lib/components/artist/shows/metadata.js
@@ -16,13 +16,14 @@ class Metadata extends React.Component {
         <Text style={styles.sansSerifText}>{this.showTypeString(this.props.show)}</Text>
         <SerifText style={styles.serifText}>{this.props.show.name}</SerifText>
         { this.dateAndLocationString(this.props.show) }
+
       </View>
     );
   }
 
   showTypeString(show) {
     const count = show.counts.artworks;
-    let message = show.kind.toUpperCase() + ' SHOW';
+    let message = show.kind.toUpperCase() + (show.kind === 'fair' ? ' BOOTH' : ' SHOW');
     if (count > 0) {
        message += `, ${count} ${count == 1 ? 'WORK' : 'WORKS'}`;
     }
--- a/lib/components/artist/shows/metadata.js
+++ b/lib/components/artist/shows/metadata.js
@@ -10,13 +10,12 @@ import SerifText from '../../text/serif'
 class Metadata extends React.Component {
 
   render() {
-    // debugger;
     return (
       <View style={styles.container}>
         <Text style={styles.sansSerifText}>{this.props.show.partner.name.toUpperCase()}</Text>
         <Text style={styles.sansSerifText}>{this.showTypeString(this.props.show)}</Text>
         <SerifText style={styles.serifText}>{this.props.show.name}</SerifText>
-        <SerifText style={styles.serifText, {color: 'grey'}}>{this.dateAndLocationString(this.props.show)}</SerifText>
+        { this.dateAndLocationString(this.props.show) }
       </View>
     );
   }
@@ -31,7 +30,7 @@ class Metadata extends React.Component {
   }
 
   dateAndLocationString(show) {
-    return show.location.city? show.location.city + ', ' + show.exhibition_period : "";
+    return show.location.city? <SerifText style={styles.serifText, {color: 'grey'}}>{show.location.city + ', ' + show.exhibition_period}</SerifText> : null;
   }
 }
 
--- a/lib/components/artist/shows/metadata.js
+++ b/lib/components/artist/shows/metadata.js
@@ -22,21 +22,24 @@ class Metadata extends React.Component {
     );
   }
 
-  showTypeString(show) {
-    const count = show.counts.artworks;
-    let message = show.kind.toUpperCase() + (show.kind === 'fair' ? ' BOOTH' : ' SHOW');
-    if (count > 0) {
-       message += `, ${count} ${count == 1 ? 'WORK' : 'WORKS'}`;
-    }
+  showTypeString() {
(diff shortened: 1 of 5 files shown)


lib/components/artist/shows/metadata.js
```diff
@@ -55,8 +54,8 @@ export default Relay.createContainer(Metadata, {
       fragment on PartnerShow {
         kind
         name
-        start_at
-        end_at
+        exhibition_period
+        status
```
sarahscott: I want to add:
```
counts: {
  artworks(artist_id: ARTIST_ID)
}
```

but I'm not sure what the right form is since that would require the artist ID to reach this fragment somehow.

alloy: Ah, I have been working on this! It took me quite some time to find it, but now I think I know what needs to be done.

alloy:
Yeah indeed, the dumped metaphysics schema we have doesn’t yet contain these fields. I’ll do a quick PR that updates just the schema so you can continue and then I’ll do a separate PR for the artworks count part.

Btw, you want `status_update`, not `status`.

lib/components/artist/shows/metadata.js
```diff
@@ -10,13 +10,12 @@ import SerifText from '../../text/serif'
...
 
   render() {
-    // debugger;
     return (
       <View style={styles.container}>
         <Text style={styles.sansSerifText}>{this.props.show.partner.name.toUpperCase()}</Text>
         <Text style={styles.sansSerifText}>{this.showTypeString(this.props.show)}</Text>
         <SerifText style={styles.serifText}>{this.props.show.name}</SerifText>
-        <SerifText style={styles.serifText, {color: 'grey'}}>{this.dateAndLocationString(this.props.show)}</SerifText>
+        { this.dateAndLocationString(this.props.show) }
```
sarahscott: So if-else statements don't work in JSX: https://facebook.github.io/react/tips/if-else-in-JSX.html, but this method is. Basically, the centering was off because the parent was accounting for an invisible line of text when `dateAndLocationString` was null. This way, the `SerifText` never exists unless it's supposed to, so centering works as expected.

alloy: Good catch 👌

lib/components/artist/shows/metadata.js
```diff
@@ -10,13 +10,12 @@ import SerifText from '../../text/serif'
 class Metadata extends React.Component {
 
   render() {
-    // debugger;
```
alloy: oopsie, did I leave that in :D

lib/components/artist/shows/metadata.js
```diff
@@ -31,7 +30,7 @@ class Metadata extends React.Component {
   }
 
   dateAndLocationString(show) {
-    return show.location.city? show.location.city + ', ' + show.exhibition_period : "";
+    return show.location.city? <SerifText style={styles.serifText, {color: 'grey'}}>{show.location.city + ', ' + show.exhibition_period}</SerifText> : null;
```
alloy: I think that this function will read much better if you make this a if-else block.

lib/components/artist/shows/metadata.js
```diff
@@ -33,6 +34,10 @@ class Metadata extends React.Component {
   dateAndLocationString(show) {
     return show.location.city? <SerifText style={styles.serifText, {color: 'grey'}}>{show.location.city + ', ' + show.exhibition_period}</SerifText> : null;
   }
+
+  statusText(show) {
+    return show.status_update? <SerifText style={{color: show.status_update.match(/opening/i) ? colors['green-regular'] : colors['red-regular']}}>{show.status_update}</SerifText> : null;
```
alloy: Instead of matching against a human-readable message, you could use the `status` field here to compare against.

lib/components/artist/shows/metadata.js
```diff
@@ -16,7 +17,7 @@ class Metadata extends React.Component {
         <Text style={styles.sansSerifText}>{this.showTypeString(this.props.show)}</Text>
         <SerifText style={styles.serifText}>{this.props.show.name}</SerifText>
         { this.dateAndLocationString(this.props.show) }
-
+        { this.statusText(this.props.show) }
```
alloy: Btw, why are you passing the show to these methods instead of accessing `this.props.show` from within the method?

lib/components/artist/shows/metadata.js
sarahscott: Is the status always either `running` or `closed`?

sarahscott: Nevermind, I see that it can be `upcoming` too. I'll use that.

lib/components/artist/shows/metadata.js
sarahscott: Hmmm no particular reason. I can make them use `this.props.show`.

alloy:
You missed a bunch of changes when removing the artworks count: https://github.com/artsy/emission/commit/61205e7197eca1471eea129cb8292d72814b3cba#diff-c8c0ee1b86fd31fb34e94c6f043496ccR18

In the future, it’s simpler to use `git revert HASH` instead.

sarahscott:
Yeah that would have made sense 🙃

alloy:
Woohoo, awesome! :metal:

PR #38, opened 2016-05-08
[Shows] Add number of works in show by artist

alloy:


data/schema.json
```diff
@@ -2420,8 +2420,8 @@
               "deprecationReason": null
             },
             {
-              "name": "is_contactable",
-              "description": "Are we able to display a contact form on artwork pages?",
+              "name": "is_inquireable",
```
mennenia: so `schema.json` is essentially where all types are defined?

alloy: Yup! It’s a representation of `schema.graphql` that can be used by the `babel-relay-plugin` to verify that a compiled Relay query matches existing fields and how these need to be mapped into objects.

README.md
```diff
@@ -28,16 +28,39 @@
...
+
+----
+
+Another gotcha is around fragments that use variables. For this it is important to understand that whenever a Relay
+backed hierarchy is used, 2 trees will be rendered.
+
+1. A tree of all Relay query fragments is rendered into a single query.
+2. Once the query has been performed, the view component tree is rendered.
+
+What this means in practice, is that you will need to pass variables down _both_ those trees.
```
sarahscott: Interesting. This is a great overview!

lib/components/artist/shows/metadata.js
```diff
@@ -63,6 +72,9 @@ export default Relay.createContainer(Metadata, {
         location {
           city
         }
+        counts {
+          artworks(artist_id: $artistID)
```
sarahscott: 🎉 !

PR #41, opened 2016-05-09
[Artist Bio] Add blurb

### LINKED ISSUE #31: [Biography] Use 'blurb' instead of 'bio'
I know this part is under construction, but force seems to use the `blurb` field for artist bios instead of `bio`

alloy:
Fixes #31.

For now we strip markdown, as rendering it would probably cause some surprise gotchas. See #42 for future work.

### DIFF
--- a/lib/components/text/serif.js
+++ b/lib/components/text/serif.js
@@ -6,16 +6,17 @@ const { StyleSheet, Text } = React;
 
 export default class Serif extends React.Component {
   render() {
+    const { children, style, ...props } = this.props;
     return (
-      <Text style={[styles.default, this.props.style, styles.required]} numberOfLines={1}>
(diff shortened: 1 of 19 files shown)


lib/components/text/serif.js
```diff
@@ -6,16 +6,17 @@ const { StyleSheet, Text } = React;
...
+      <Text style={[styles.default, style, styles.required]} numberOfLines={1} {...props}>
+        {children}
       </Text>
     );
   }
 }
 
 Serif.propTypes = {
-  style: Text.propTypes.style,
+  ...Text.propTypes,
```
sarahscott: Still getting used to this weird piece of syntax

sarahscott:
👍

PR #46, opened 2016-05-10
Use cropped image URLs from metaphyics

### LINKED ISSUE #30: Update all fixed-size images to use cropped url from metaphysics
Relevant for images used in:
- [x] Shows
- [ ] Articles
- [ ] Related Artists

sarahscott:
- [x] Articles
- [x] Related Artists

Fixes #30

### DIFF
--- a/lib/components/artist/related_artists/related_artist.js
+++ b/lib/components/artist/related_artists/related_artist.js
@@ -12,13 +12,10 @@ import ImageView from '../../opaque_image_view';
 class RelatedArtist extends React.Component {
   render() {
     const artist = this.props.artist;
-    const {height, width} = Dimensions.get('window');
-    const imageWidth = (width - 60)/2;
-    const imageAspectRatio = 1.5;
     return (
-      <View style={{paddingBottom: 30, width: imageWidth}}>
-        <ImageView style={{ width: imageWidth, height: imageWidth/imageAspectRatio }}
-                   imageURL={artist.image.url} />
+      <View style={{paddingBottom: 30, width: this.props.relay.variables.width}}>
+        <ImageView style={{ width: this.props.relay.variables.width, height: this.props.relay.variables.height }}
+                   imageURL={artist.image.cropped.url} />
         <Text style={styles.sansSerifText}>{artist.name.toUpperCase()}</Text>
         <Text style={styles.serifText}>{this.artworksString(artist.counts)}</Text>
       </View>
@@ -52,6 +49,8 @@ const styles = StyleSheet.create({
 });
 
 export default Relay.createContainer(RelatedArtist, {
+  initialVariables: { width: 150, height: 100 },
+
   fragments: {
     artist: () => Relay.QL`
       fragment on Artist {
@@ -61,7 +60,9 @@ export default Relay.createContainer(RelatedArtist, {
           artworks
         }
         image {
-          url
+          cropped(width: $width, height: $height) {
+            url
+          }
         }
       }
     `,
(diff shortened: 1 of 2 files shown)


sarahscott:
@alloy something about my `RelatedArtist` solution feels weird but I'm not sure why. In any case, I've built both of these without considering iPad or orientation, which I want to address in my next PR.

lib/components/artist/related_artists/related_artist.js
```diff
@@ -12,13 +12,10 @@ import ImageView from '../../opaque_image_view';
...
   render() {
     const artist = this.props.artist;
-    const {height, width} = Dimensions.get('window');
-    const imageWidth = (width - 60)/2;
-    const imageAspectRatio = 1.5;
     return (
-      <View style={{paddingBottom: 30, width: imageWidth}}>
-        <ImageView style={{ width: imageWidth, height: imageWidth/imageAspectRatio }}
-                   imageURL={artist.image.url} />
+      <View style={{paddingBottom: 30, width: this.props.relay.variables.width}}>
```
sarahscott: This is a cool thing that relay can do.

alloy: Interesting, hadn’t considered that yet 👍 

I do wonder if this will clash when we take screen scale into consideration, not sure if `prepareVariables` actually changes the variables. But we’ll see that then.

alloy:
> @alloy something about my `RelatedArtist` solution feels weird but I'm not sure why. In any case, I've built both of these without considering iPad or orientation, which I want to address in my next PR.

I think the problem is that we need to know the dimensions during the initial Relay query building. So I guess we need to move the width calculation to Relay’s `prepareVariables`:

```js
const {height, width} = Dimensions.get('window');
const imageWidth = (width - 60)/2;
```

lib/components/artist/related_artists/related_artist.js
```diff
@@ -52,6 +49,19 @@ const styles = StyleSheet.create({
 });
 
 export default Relay.createContainer(RelatedArtist, {
+  initialVariables: { imageWidth: 150, imageHeight: 100 },
+  prepareVariables: (prevVariables) => {
+    const {windowHeight, windowWidth} = Dimensions.get('window');
+    let newWidth = (windowWidth - 60)/2;
+    console.log(newWidth);
+    console.log('window width: ' + windowWidth);
```
sarahscott: The `windowWidth` here is undefined, but in Facebook's documentation, they're using some sort of window: https://facebook.github.io/relay/docs/api-reference-relay-container.html#preparevariables

🤔

lib/components/artist/related_artists/related_artist.js
```diff
@@ -52,6 +49,19 @@ const styles = StyleSheet.create({
...
+    return {
+      // imageWidth : newWidth,
+      // imageHeight : 100
+      imageWidth: 150,
+      imageHeight: 100
```
sarahscott: I left these hardcoded to make sure this all works without breaking, which it does until you try to send `imageWidth` to `NaN` because the dimensions aren't defined.

lib/components/artist/related_artists/related_artist.js
alloy: Le Sigh. The [docs](https://facebook.github.io/react-native/docs/dimensions.html#content) state that it should always be available.

Ok, let’s proceed with hardcoded for now and maybe you can use that device package you found to determine what device and thus the screen size is?

orta: Parent view's width?

alloy: View’s have not been created/laid out yet when Relay is doing its initial query building. We could probably pass data in from outside when creating the artist component tree, though.

sarahscott: I'm going to timebox trying to figure out why `Dimensions` is undefined here because it seems very suspicious, especially given the documentation that uses `window`. Perhaps it's something to do with my code.

alloy:
Ace!

PR #48, opened 2016-05-10
Add artworks grid

### LINKED ISSUE #45: Add artworks grid component
This needs to be copied over from my prototype app.

While I have a working one that’s backed by a native collection view and our masonry grid layout, I’d like to try to use a pure RN one for now.

The one issue we have is that we don’t have pagination working atm. Maybe we can have a stop-gap hybrid that does further paging without Relay, but that still requires me to work out a way to load the next page when scrolled to the bottom.

alloy:
Closes #45

I’m fairly happy with the, for now, naive calculation of how to evenly distribute artworks across the sections, because by the end of 100 artworks, it looks like this:

simulator screen shot 10 may 2016 23 22 15


lib/components/artworks/index.js
```diff
@@ -12,8 +12,6 @@ import Relay from 'react-relay';
 import React from 'react-native';
 const { View, ScrollView, StyleSheet } = React;
 
-import { max } from 'lodash';
```
alloy: While I no longer use lodash, I’m leaving it in because we’re bound to use it again and I already did the work to pin it to the version that some of our dependencies already use.

alloy:
Need https://github.com/artsy/metaphysics/pull/238 merged to be able to do pagination.

Example/Podfile
```diff
@@ -7,7 +7,7 @@ target 'Emission' do
                :subspecs => %w( Core RCTText RCTWebSocket RCTNetwork ),
                :inhibit_warnings => true
 
-  if ENV['ARTSY_STAFF_MEMBER'] != nil || ENV['CI'] != nil
+  if true #ENV['ARTSY_STAFF_MEMBER'] != nil || ENV['CI'] != nil
```
alloy: Did this for now because with the CocoaPods app my env variables aren’t set 🤔

/cc @orta

orta: ;)

got opinions on exposing the host's ENV vars to the CP mac app? I guess we could do it then ensure CP's ENVs are set up after

alloy: it’s not that simple, though, env vars for one’s shell are usually defined in a shell specific file, and not all shells are compatible with each other. E.g. I use the fish shell and its files cannot be interpreted as (ba)sh.

iirc you have to use `~/.MacOSX/environment.plist` for GUI apps, not sure if that would be enough in our case. https://developer.apple.com/library/mac/documentation/MacOSX/Conceptual/BPRuntimeConfig/Articles/EnvironmentVars.html

lib/components/artworks/artwork.js
```diff
@@ -0,0 +1,52 @@
...
+    // TODO partner ?
+    if (artwork.sale_message) {
+      optionals.push(<Text key="sale_message">{this.props.artwork.sale_message}</Text>);
+    }
+    return (
+      <View>
+        <ImageView style={styles.image} aspectRatio={artwork.image.aspect_ratio} imageURL={artwork.image.url} />
+        <Text>{artwork.artist.name}</Text>
+        <Text>{artwork.title}</Text>
+        <Text>{artwork.partner.name}</Text>
```
sarahscott: The spec doesn't seem to have the partner name - is it supposed to?

alloy: Ah, good catch, this component I made based on a generalized artwork component design, it should indeed be updated to match the spec for the artist page.

sarahscott: Also don't we need to show the price?

lib/components/artworks/index.js
```diff
@@ -0,0 +1,224 @@
+'use strict';
+
+// 1. Get first layout pass of grid view so we have a total width and calculate the column width (componentDidMount?).
+// 2. Possibly do artwork column layout now, as we can do so based just on the aspect ratio, assuming the text height
+//    won't be too different between artworks.
+// 3. Get artwork heights by either:
+//    - calculating the item size upfront with aspect ratio and a static height for the text labels.
+//    - leting the artwork component do a layout pass and calculate its own height based on the column width.
```
sarahscott: letting*

lib/components/artworks/index.js
```diff
@@ -0,0 +1,224 @@
...
+    this.state = { sectionDimension: 0, artworks: this.props.artworks, page: 1 };
+
+    this.onLayout = this.onLayout.bind(this);
+    this.onScroll = this.onScroll.bind(this);
+
+    this._sentEndForContentLength = null;
+  }
+
+  fetchNextPage() {
+    const nextPage = this.state.page + 1;
```
sarahscott: Why not `++`?

lib/components/artworks/index.js
```diff
@@ -0,0 +1,224 @@
...
+    const query = artworksQuery(this.props.relay.route.params.artistID, nextPage);
+    fetch(metaphysicsURL, { method: 'post',
```
sarahscott: Why is this a `post`?

lib/components/artworks/index.js
```diff
@@ -0,0 +1,224 @@
...
+      // This is the sum of all margins in between sections, so do not count to the right of last column.
+      const sectionMargins = this.props.sectionMargin * (this.props.sectionCount - 1);
+      this.setState({ sectionDimension: (layout.width - sectionMargins) / this.props.sectionCount });
+    }
+  }
+
+  sectionedArtworks() {
+    const sectionedArtworks = [];
+    const sectionRatioSums = [];
+    for (let i = 0; i < this.props.sectionCount; i++) {
```
sarahscott: Why not `each`?

lib/components/artworks/index.js
```diff
@@ -0,0 +1,224 @@
...
+    };
+
+    const sectionedArtworks = this.sectionedArtworks();
+    const sections = [];
+    for (let i = 0; i < this.props.sectionCount; i++) {
+      const artworkComponents = [];
+      const artworks = sectionedArtworks[i];
+      for (let j = 0; j < artworks.length; j++) {
+        artworkComponents.push(<Artwork artwork={artworks[j]} key={'artwork-'+j} />);
+        // Setting a marginBottom on the artwork component didn’t work, so using a spacer view instead.
```
sarahscott: Interesting

lib/components/artworks/index.js
```diff
@@ -0,0 +1,224 @@
...
+          artworkComponents.push(<View style={spacerStyle} key={'spacer-'+j} accessibilityLabel='Spacer View' />);
+        }
+      }
+
+      const sectionSpecificStyle = {
+        width: this.state.sectionDimension,
+        marginRight: (i == this.props.sectionCount-1 ? 0 : this.props.sectionMargin),
+      };
+      sections.push(
+        <View style={[styles.section, sectionSpecificStyle]} key={i} accessibilityLabel={'Section ' + i}>
```
sarahscott: Why the `accessibilityLabel`?

lib/components/artworks/index.js
```diff
@@ -0,0 +1,224 @@
...
+  onScroll(event) {
+    const scrollProperties = event.nativeEvent;
+    const contentLength = scrollProperties.contentSize.height;
+    if (contentLength !== this._sentEndForContentLength) {
+      const offset = scrollProperties.contentOffset.y;
+      const visibleLength = scrollProperties.layoutMeasurement.height;
+      const distanceFromEnd = contentLength - visibleLength - offset;
+      if (distanceFromEnd < 100) {
+        this._sentEndForContentLength = contentLength;
+        this.fetchNextPage();
```
sarahscott: 🎉

lib/containers/artist.js
```diff
@@ -37,7 +38,7 @@ class Artist extends React.Component {
         </View>
       );
       case 1: return (
-        <React.Text>HERE GO THE WORKS</React.Text>
+        <Artworks artworks={this.props.artist.artworks} />
```
sarahscott: 💃

lib/components/artworks/index.js
alloy: doh

lib/components/artworks/index.js
alloy: The fetch is going to kick-off an asynchronous network call, changing the state here would means the component gets re-rendered _before_ the data is available.

Instead I set the state when the new artwork data is parsed https://github.com/artsy/emission/pull/48/files#diff-3ad0eaec0c07ae91811e28b062ed8d1bR62

sarahscott:
Everything looks mergeable! My questions are just out of mine (and @orta's) curiosity.

lib/components/artworks/index.js
alloy: I just mimiced what Relay does. It also probably shows-up nicer in the networking tab of the dev tools than a GET would.

lib/components/artworks/index.js
alloy: This isn’t an array.

I could definitely have used `forEach` on the artworks array below. I guess at some point I was using the `i` variable for something else as well.

lib/components/artworks/index.js
alloy: Could definitely have been my ignorance, feel free to take a stab at some point.

lib/components/artworks/index.js
alloy: Ah yes, changing these makes the names in Xcode’s view inspector more helpful.

lib/components/artworks/index.js
alloy: Hmm, I now realise that I didn’t try to do anything special handling for the last page. Filing a ticket for that.

PR #49, opened 2016-05-11
iPad adjustments

sarahscott:
- [x] Figured out how to get image sizes from screen width in Relay
- [x] Swapped `SmallList` for `MediumList` for iPad past shows module
- [x] Margins for iPad
- [x] Image sizing for related artists on iPad
- [x] Use smaller images for Past Shows on iPad

ipad_adjustments


lib/components/artist/related_artists/related_artist.js
```diff
@@ -49,7 +49,18 @@ const styles = StyleSheet.create({
 });
 
 export default Relay.createContainer(RelatedArtist, {
-  initialVariables: { width: 150, height: 100 },
+  initialVariables: { imageWidth: 150, imageHeight: 100 },
+
+  prepareVariables: prevVariables => {
+    const {width, height} = Dimensions.get('window');
+    let w = width;
```
sarahscott: @alloy I put a breakpoint here and it's getting called before `render()` but for some reason `render()` uses the `initialVariables` instead of these modified ones. Is there something else I need to do to make sure `prepareVariables` overrides `initialVariables`?

alloy: I guess `prepareVariables` doesn’t update. Maybe there’s another Relay hook method? Or otherwise see if the dimensions are available at the top-level of the file.

```js

export default Relay.createContainer(RelatedArtist, {
  initialVariables: { imageWidth: windowDimensions.width, imageHeight: windowDimensions/1.5 }
```

lib/components/artist/related_artists/related_artist.js
```diff
@@ -49,7 +52,7 @@ const styles = StyleSheet.create({
 });
 
 export default Relay.createContainer(RelatedArtist, {
-  initialVariables: { width: 150, height: 100 },
+  initialVariables: { imageWidth: ~~(imageWidth), imageHeight: ~~(imageWidth / 1.5) },
```
alloy: wow, such wise bit operations

sarahscott: lol is it too much? I figured it was prettier than using `Math`: http://stackoverflow.com/questions/596467/how-do-i-convert-a-float-number-to-a-whole-number-in-javascript

alloy: I love such tricks, but it’s a bit much and not very readable and readability is more important imo

lib/components/artist/shows/index.js
```diff
@@ -43,11 +46,9 @@ const styles = StyleSheet.create({
   },
 });
 
-export default Relay.createContainer(Shows, {
-  initialVariables: {
-    artistID: null,
-  },
+var pastShowsFragment = windowDimensions.width > 700 ? LargeShowsList.getFragment('shows') : SmallShowsList.getFragment('shows');
```
sarahscott: This could also go inside `createContainer` but it might be more readable up here.

lib/components/artist/related_artists/related_artist.js
sarahscott: Fair

lib/components/artist/shows/index.js
alloy: Yup, looks fine 👌 

But use `const` instead of `var`.

sarahscott:
So the spec shows larger images for landscape iPad: 
6699e83e-d00d-11e5-92b5-aa006aaafe96

But that would require another network request to get the image cropped at that size. Right now, I have a nice-looking RelatedArtists component that uses the same size for images but lays them out differently upon rotation:

ipadrelatedartistsrotate


IMO this looks nice, but I understand if it's necessary to get the larger images upon rotation. Is it possible to push that to post-v1?

alloy:
Yeah I think that's fine. We'll have to see what it looks like during QA.

alloy:
I didn’t see the animated GIF before, looks really good!!!! :clap:

orta:
++ agreed

lib/components/artist/shows/large_image_show.js
```diff
@@ -0,0 +1,43 @@
...
+import React from 'react-native';
+const { View, Text, StyleSheet, Dimensions } = React;
+
+import Headline from '../../text/headline'
+import SerifText from '../../text/serif'
+import OpaqueImageView from '../../opaque_image_view'
+import ShowMetadata from './metadata'
+
+const windowDimensions = Dimensions.get('window');
+const imageWidth = windowDimensions.width > 700 ? (windowDimensions.width - 85)/2 : (windowDimensions.width - 40);
```
sarahscott: I'm having some trouble here. The iPad Past Shows spec has 3 columns, and if I change the 2 in `(windowDimensions.width - 85)/2` to a 3, this image will be rendered nicely in a `LargeList`. The problem is that the `LargeList` doesn't know if it's showing past shows or current shows. Is there a clever way to deal with this? The brute-force approach is to make a whole Past Shows list component and `medium_image_show`.

alloy: I would say, for now make that extra component. It’s fine to have some duplication for a while, then once you have everything implemented that you need, it will be much clearer what to abstract and how, as the overlap will be clear.

sarahscott:
So without image resizing, shows is looking like: 

shows_ipad_rotation

sarahscott:
Ready for review; made two new issues that belong in a different pr.

alloy:
Hmm, we definitely need to resize the current & upcoming shows to fit, as per the spec.

Let’s do this, you just calculate the dimensions of the current & upcoming shows images in landscape orientation and then you ask Relay for an image cropped to that size so that we have the largest one we ever need, but do the actual show component sizing as you would do normally so that it’s slightly smaller in portrait mode.

alloy:
(Also needs a rebase.)

sarahscott:
I'm going to rebase, and then I'd prefer to merge this as-is and start a new PR for image sizing. There are a lot of styling updates in here.

sarahscott:
Rebased!

PR #52, opened 2016-05-12
Update `sync-schema` workflow

alloy:
It will now be easy for anyone to update our copy of metaphysics’ schema.


orta:
Is this schema used in the app anywhere?

alloy:
@orta Relay uses the JSON version to validate the queries you build. The GraphQL version is meant for us to be able to easily check the schema.

alloy:
https://facebook.github.io/relay/docs/guides-babel-plugin.html#schema-json

orta:
Perfect, that's the link I was looking for

orta:
I don't know where the abstractions start / stop yet once this is back in eigen, but it may be worth adding a `make metaphysics_schema` action that triggers `npm blah blah` this to make it consistent with the rest of the app

alloy:
That command is only needed while working on Emission, but I’ll definitely bridge whatever commands are needed in Eigen 👍

sarahscott:
👍

Issue #13, opened 2016-04-29
[Image] Get images sized according to screen DPI


alloy:
E.g. if we want to display an image at `200x100`, we should make GraphQL queries like so:

Screen scale:
1. `200x100`
2. `400x200`
3. `600x300`

For those images where we dynamically size based on the aspect ratio (e.g. masonry grid) we only specify 1 dimension (e.g. width) that’s specified according to screen scale and let metaphysics figure out the resize arguments based on the image’s aspect ratio.

alloy:
https://facebook.github.io/relay/docs/api-reference-relay-container.html#preparevariables

alloy:
The problem we’re facing with this is that we won’t know the view dimensions until _after_ a layout pass has been performed. This means that we’d have to:

1. Fetch initial metadata such as aspect ratio and all other Relay data for the current view.
2. Do a layout pass and start rending views.
3. Trigger additional Relay data fetches now that we know the size at which to display the view.

In a situation where there can be many image views, these additional fetches can add up, so this isn’t really an option. We also don’t want to hardcode all sizes for all devices, it needs to be dynamic.

As such, I think that we would need to generate the image variant URL on the client after all, which means that:

* we need metaphysics to return the template URL, we can do that during the initial fetch
* we need code to generate the template URL, which metaphysics already has, so maybe that can be extracted?

/cc @dzucconi

orta:
> The problem we’re facing with this is that we won’t know the view dimensions until after a layout pass has been performed. This means that we’d have to:

Only on the current view, you could ask `UIScreen mainScreen`?

alloy:
> Only on the current view

Do you mean, “for the time being” ? Because in the long-run, I’d really rather not rely on using the device screen in calculations in components.

orta:
I did, aye

orta:
I don't think you can pull it out of anywhere else other than those globals like `UIWindow` or `UIScreen`, which if you want this separated from the current view you're a bit stuck. An option is to have those pre-defined a global state when the app launches?

orta:
Don't think I've ever heard of the scale changing on a UIView, but then on my Mac it is always shifting between retina/not retina as I have two screens

alloy:
Not sure if we’re on the same page here, buddy, I’m not trying to deal with screen scale changing on a single device.

orta:
You never know what apple can come up with at WWDC ;)

alloy:
[...]

PR #54, opened 2016-05-12
Dynamically sized image fetching.

alloy:
Part of #13.

### DIFF
--- a/lib/components/artist/articles/article.js
+++ b/lib/components/artist/articles/article.js
@@ -14,9 +14,8 @@ class Article extends React.Component {
     const article = this.props.article
     return (
       <View style={{ width: 300 }}>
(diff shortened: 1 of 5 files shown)


lib/components/artist/articles/article.js
```diff
@@ -12,13 +12,16 @@ import ImageView from '../../opaque_image_view';
 class Article extends React.Component {
   render() {
     const article = this.props.article
+    // TODO The outer wrapping view is currently only there because setting `marginRight: 20` on the Article from the
```
sarahscott: Yeah this is a weird thing - I find that setting the `style` of an entire component from the parent gives mixed results. I should read up on how flexbox reconciles self-styling and styling-by-parent.

sarahscott:
🎉

PR #56, opened 2016-05-13
Artworks styling

### LINKED ISSUE #51: [Artworks] Style artwork component
Currently it’s not yet using the right font and some partner details might still be missing.

alloy:
Fixes #51.

Still need to add works count next to section title.

Currently I don’t paginate the ‘for sale’ section, so need to find the artist with the most for sale works and see if that would be an issue right now or if we can delay worrying about that till post v1.

### DIFF
--- a/Example/Emission/AppDelegate.m
+++ b/Example/Emission/AppDelegate.m
@@ -59,8 +59,11 @@ randomBOOL(void)
   ARComponentViewController *componentViewController = [[ARComponentViewController alloc] initWithBridge:self.reactBridge
                                                                                               moduleName:@"Artist"];
   self.window = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
+  self.window.backgroundColor = [UIColor whiteColor];
   self.window.rootViewController = [[UINavigationController alloc] initWithRootViewController:componentViewController];
   [self.window makeKeyAndVisible];
+  
+  [(UINavigationController *)self.window.rootViewController setNavigationBarHidden:YES];
   
   return YES;
 }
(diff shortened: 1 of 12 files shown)


alloy:
The styling part is done, but I need to page in for sale works first and then the other works.

Example/Emission/AppDelegate.m
```diff
@@ -59,9 +59,12 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   ARComponentViewController *componentViewController = [[ARComponentViewController alloc] initWithBridge:self.reactBridge
                                                                                               moduleName:@"Artist"];
   self.window = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
+  self.window.backgroundColor = [UIColor whiteColor];
   self.window.rootViewController = [[UINavigationController alloc] initWithRootViewController:componentViewController];
   [self.window makeKeyAndVisible];
   
+  [(UINavigationController *)self.window.rootViewController setNavigationBarHidden:YES];
```
sarahscott: yesss

lib/components/artworks/grid.js
```diff
@@ -0,0 +1,221 @@
...
+    flexDirection: 'column',
+  },
+});
+
+// TODO: While we do pagination manually, we can’t actually use a Relay container around Artwork.
+const artworksQuery = (artistID, page) => {
+  return `
+    query {
+      artist(id: "${artistID}") {
+        artworks(sort: published_at_asc, filter: IS_NOT_FOR_SALE size: 10, page: ${page}) {
```
sarahscott: Does the filter exclude for sale works or make the query return only for sale works?

alloy: The former, it only returns not for sale works.

sarahscott:
👍

PR #59, opened 2016-05-13
[Artworks] First paginate for sale works, then when depleted paginate other works.

alloy:
Merge #56 first.

### DIFF
--- a/lib/components/artworks/index.js
+++ b/lib/components/artworks/index.js
@@ -159,7 +159,10 @@ class Artworks extends React.Component {
   render() {
     const artworks = this.state.sectionDimension ? this.renderSections() : null;
     return (
-      <ScrollView onScroll={this.onScroll} scrollEventThrottle={50} onLayout={this.onLayout} accessibilityLabel='Artworks ScrollView'>
+      <ScrollView onScroll={this.props.paginate ? this.onScroll : null}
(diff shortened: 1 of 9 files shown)


lib/components/artworks/index.js
```diff
@@ -7,32 +7,51 @@ const { Text, View, StyleSheet } = React;
...
         <View>
-          {this.renderSection('Works for Sale', for_sale_count, this.props.artist.for_sale_artworks, false)}
-          <Separator/>
-          {this.renderSection('Other Works', other_count, this.props.artist.not_for_sale_artworks, true)}
+          {this.renderSection('Works for Sale',
+                              for_sale_count,
+                              this.props.artist.for_sale_artworks,
+                              (artistID, page) => ArtworksGrid.artworksQuery(artistID, 'IS_FOR_SALE', page),
+                              () => { this.setState({ completedForSaleWorks: true }) })}
+          {otherWorks}
```
alloy: This is where the React shadow DOM magic happens. Because we always render the ‘for sale’ section in the same place and only conditionally _add_ the ‘other works’ section, React knows that it needs to leave the existing (fully paginated) ‘for sale’ instance on screen _without_ us having to keep a reference to that section instance and have extra logic to create a new instance or re-use the existing one.

PR #60, opened 2016-05-14
More artworks styling and ipad

alloy:
Might want to do an extra pass during QA to check if I didn’t mess up any of your styling work.


sarahscott:
👍

PR #61, opened 2016-05-14
[ImageView] Refinements

alloy:
- Only show placeholder color if the image needs downloading. Otherwise you might get to see the color for a very short period, which isn’t the nicest.
- Fade image in.

### DIFF
--- a/Pod/Classes/OpaqueImageViewComponent/AROpaqueImageView.m
+++ b/Pod/Classes/OpaqueImageViewComponent/AROpaqueImageView.m
@@ -57,7 +57,12 @@ LoadImage(UIImage *image, CGSize destinationSize, CGFloat scaleFactor, void (^ca
 {
   // This will cancel an in-flight download operation, if one exists.
   self.imageURL = nil;
-  [super setImage:image];
+
+  [UIView transitionWithView:self
+                    duration:0.25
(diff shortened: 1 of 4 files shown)


Pod/Classes/OpaqueImageViewComponent/AROpaqueImageView.m
```diff
@@ -57,7 +57,12 @@ - (void)setImage:(UIImage *)image;
 {
   // This will cancel an in-flight download operation, if one exists.
   self.imageURL = nil;
-  [super setImage:image];
+
+  [UIView transitionWithView:self
+                    duration:0.25
+                     options:UIViewAnimationOptionTransitionCrossDissolve
+                  animations:^{ [super setImage:image]; }
+                  completion:nil];
```
orta: this would make it tough to write tests against I assume

alloy: Could probably be done with `[UIView setAnimationsEnabled:NO];` ?

orta: It could, aye, guess I'm used to DI over global state, but this seem 👍

orta:
Cool

PR #62, opened 2016-05-17
[Artworks] Fetch next page long before user (typically) reaches end of view.

alloy:
Giphy


alloy:
This is the default value that FB’s ListView uses as well, btw.

orta:
Looks legit

PR #63, opened 2016-05-17
Eigen integration work

alloy:
This includes:

* A new  `npm run bundle` task to generate a release JS bundle that’s included in the pod.
* A new ‘switch board’ module that allows presenting navigation/modal view controllers from JS.
* Adds tap handling to related artists to demo the above.
* A `AREmission` class that makes initialisation simpler and holds all state for the live `RCTBridge` instance and the various modules we provide (API, SwitchBoard)


lib/components/artist/related_artists/related_artist.js
```diff
@@ -3,25 +3,32 @@
...
 import ImageView from '../../opaque_image_view';
+import SwitchBoard from '../../../modules/switch_board';
 
 const windowDimensions = Dimensions.get('window');
 const imageWidth = windowDimensions.width > 700 ? (windowDimensions.width - 100)/3 : (windowDimensions.width - 60)/2;
 
 class RelatedArtist extends React.Component {
+  handleTap() {
+    SwitchBoard.presentNavigationViewController(this, `/artist/${this.props.artist._id}`);
+  }
```
alloy: Seems like metaphysics doesn’t provide slugs for most models, the real DB ID should work too, though.

Example/Emission/AppDelegate.m
```diff
@@ -1,45 +1,55 @@
...
 #import <Emission/ARTemporaryAPIModule.h>
+#import <Emission/ARSwitchBoardModule.h>
 
-#import <React/RCTBridge.h>
 #import <React/RCTUtils.h>
+#import <TargetConditionals.h>
+
+#ifdef TARGET_OS_SIMULATOR
+#define ENABLE_DEV_MODE
+#endif
```
alloy: You can comment this out to use the release JS bundle that’s bundled in the app.

Example/Emission/AppDelegate.m
```diff
@@ -50,22 +60,52 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
-  
-  NSURL *URL = [NSURL URLWithString:@"http://localhost:8081/Emission/index.ios.bundle?platform=ios&dev=true"];
-  self.reactBridge = [[RCTBridge alloc] initWithBundleURL:URL
-                                           moduleProvider:^{ return @[APIModule]; }
-                                            launchOptions:nil];
-  
-  ARComponentViewController *componentViewController = [[ARComponentViewController alloc] initWithBridge:self.reactBridge
-                                                                                              moduleName:@"Artist"];
+
+  emission.switchBoardModule.presentNavigationViewController = ^(UIViewController * _Nonnull fromViewController,
```
orta: so eigen would hook in here and use it's ARTopMenuVC stuff?

alloy: Yep and ARSwitchBoard.

orta: 👍

sarahscott:
🎉

lib/modules/switch_board.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+import React from 'react-native';
+const { findNodeHandle } = React;
+const { ARSwitchBoardModule } = React.NativeModules;
+
+function presentNavigationViewController(component, route) {
```
mennenia: So if `presentNavigationViewController` is called on a react component, it will default to this implementation and call `ARSwitchBoardModule`, but in there there are `RCT_EXPORT_METHOD`s. Will it call into the default Eigen switchboard, or does it now have its own system?

mennenia:
This looks great!

lib/modules/switch_board.js
alloy: The `RCT_EXPORT_METHOD` macro is to export a native module’s method to the JS-side of the bridge. From [the docs](http://facebook.github.io/react-native/docs/native-modules-ios.html#content):

> The name of the method exported to JavaScript is the native method's name up to the first colon. React Native also defines a macro called `RCT_REMAP_METHOD()` to specify the JavaScript method's name. This is useful when multiple native methods are the same up to the first colon and would have conflicting JavaScript names.

To do the integration with Eigen’s ARSwitchBoard, Eigen needs to implement the callbacks that the example app implements in its AppDelegate file: https://github.com/artsy/emission/pull/63/files#diff-d00d369a092249f7c8764b15b71abfd5R64

alloy: So Emission only provides the facility for React components to ask their host app to show a route, the actual routing and presenting of the VC is up to the host app.

mennenia: Awesome, thanks!

PR #69, opened 2016-05-18
More pre-integration work

alloy:
With this work PR merged, you should be able to easily run a release build on your device and start QAin artist VCs.


Example/Emission/AppDelegate.m
```diff
@@ -63,45 +63,50 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
   emission.switchBoardModule.presentNavigationViewController = ^(UIViewController * _Nonnull fromViewController,
```
orta: perhaps this should take an animated option, I could see times when you'd want to handle something non-animated

alloy: That sounds like a nice change for @sarahscott to do post v1 to dip her toes into RN native modules.

https://github.com/artsy/emission/issues/70

PR #71, opened 2016-05-19
Made components TouchableWithoutFeedback

sarahscott:


alloy:
Hahaha, good find that `href` attribute, please fix my puny attempt at route building for related artists to also use `href` instead of `_id`.

alloy:
Did you notice the first tap not registering in all other places as well?

sarahscott:
I don't think so but I've only been running this in the simulator so far.

sarahscott:
I do notice a slight delay but it could just be the simulator.

alloy:
I doubt it, that’s probably the same issue 😭

alloy:
Can you remove the console logging? The idea to log was from before I had added the switch board module and wasn’t sure if I would have had time to do so before you would work on this, now that it’s in you can just see that the right thing happens because of the VC that’s presented.

lib/components/artist/shows/small_image_show.js
```diff
@@ -0,0 +1,58 @@
+/* @flow */
```
alloy: Is this a new component?

sarahscott:
So if I tap on a `TouchableWithoutFeedback` component once, after a half-second delay it'll respond. It does always register the first tap but it takes a moment to do so. One hypothesis is that it has to do with one of the props mentioned in the documentation : https://facebook.github.io/react-native/docs/touchablewithoutfeedback.html#content. Looking into it now; timeboxing.

lib/components/artist/shows/small_image_show.js
alloy: Oh, I see it below :D

sarahscott:
Yup! Changing `onPress` to `onPressIn` seems to be the solution!

lib/components/artist/articles/article.js
```diff
@@ -3,25 +3,32 @@
...
+const { ScrollView, StyleSheet, View, Text, TouchableWithoutFeedback } = React;
 
 import colors from '../../../../data/colors';
 import SerifText from '../../text/serif';
 import ImageView from '../../opaque_image_view';
+import SwitchBoard from '../../../modules/switch_board';
 
 class Article extends React.Component {
+  handleTap() {
+    SwitchBoard.presentNavigationViewController(this, `${this.props.article.href}`);
```
alloy: Btw, the href is already a string, right? So there shouldn’t be a need to interpolate it into another (backticks) string.

alloy:
> Yup! Changing `onPress` to `onPressIn` seems to be the solution!

Hmm, interesting. This might be an acceptable change as a stopgap for v1, but normally on iOS you only want to register taps once the user releases the touch _on the element that they started on_. Otherwise you can’t cancel.

Also, can you still scroll when you start the gesture _on_ a touchable?

alloy:
Do the ‘delay’ properties not help? And, if not, maybe try to use the `onPressOut` one instead, so that it’s at least cancellable.

alloy:
I think the real problem lies with this:

> but not if cancelled (e.g. by a scroll that steals the responder lock).

sarahscott:
I'm going to play around with these for a few minutes. `onPressIn` gives mixed results wrt scrolling; sometimes it immediately responds and sometimes it lets you scroll.

PR #72, opened 2016-05-19
Use shared Eigen/Emission components from Extraction Pod and add Spinner component

alloy:
Extraction lives here https://github.com/artsy/extraction

And Spinner looks like you expect:

simulator screen shot 20 may 2016 00 22 17

simulator screen shot 20 may 2016 00 11 21


Example/Pods/Extraction/README.md
```diff
@@ -0,0 +1,13 @@
+# Extraction
+
+> Extraction in chemistry is a separation process consisting in the separation of a substance from a matrix.
+
+This pod holds various common UI components used by Eigen and Emission.
```
sarahscott: 🙌 This pod is awesome.

PR #74, opened 2016-05-19
Fix import when built as Framework.

alloy:


Pod/Classes/Core/AREmissionFontsLoader.m
```diff
@@ -1,6 +1,10 @@
 #import <Foundation/Foundation.h>
 
+#if __has_include(<Artsy+UIFonts/UIFont+ArtsyFonts.h>)
 #import <Artsy+UIFonts/UIFont+ArtsyFonts.h>
+#else
+@import Artsy_UIFonts;
```
orta: think this syntax will work all the time for both

alloy: For static libs?

orta: aye, think cocoapods makes module maps for both, not checked but could be wrong, have debated moving a bunch of our pods to that format

PR #76, opened 2016-05-20
QA work for null states

### LINKED ISSUE #64: [Shows] Don’t show ‘Current & Upcoming’ shows when there are none.
E.g. https://www.artsy.net/artist/peter-doig-2/shows

### LINKED ISSUE #65: [Shows] Don’t try to show data that’s not available
E.g. https://www.artsy.net/artist/peter-doig-2/shows has a past show where the location information is missing.

### LINKED ISSUE #66: [Biography] Don’t try to remove Markdown when there is no bio.
E.g. https://www.artsy.net/artist/zhang-hongtu

### LINKED ISSUE #67: [Articles] Don’t show ‘Featured Articles’ when there are none.
E.g. https://www.artsy.net/artist/zhang-hongtu

### LINKED ISSUE #68: [Related Artists] Use singular ‘work’ if an artist only has 1 work.
E.g. https://www.artsy.net/artist/rob-pruitt has a link to ‘Kaikai Kiki’ which has only 1 work:

img_7791

sarahscott:
- [x] Got show images on iPad to update width properly
- [x] Hides blurb if there is no blurb
- [x] Hides Current & Upcoming Shows if there are none
- [x] Uses 'work' instead of 'works' for singular works in related artist component
- [x] Hides articles if there are no articles for artist
- [x] Hide show data if missing
- [x] Hide ‘Other Works’ if all works are for sale. (see https://www.artsy.net/artist/nicole-cohen/works)

simulator screen shot may 23 2016 1 13 36 pm
simulator screen shot may 23 2016 1 14 09 pm

Fixes #68.
Fixes #67.
Fixes #66.
Fixes #64.
Will fix #65.

### DIFF
--- a/lib/components/artist/related_artists/related_artist.js
+++ b/lib/components/artist/related_artists/related_artist.js
@@ -15,7 +15,7 @@ const imageWidth = windowDimensions.width > 700 ? (windowDimensions.width - 100)
 
 class RelatedArtist extends React.Component {
   handleTap() {
-    SwitchBoard.presentNavigationViewController(this, ${this.props.artist.href});
+    SwitchBoard.presentNavigationViewController(this, this.props.artist.href);
   }
 
   render() {
@@ -34,7 +34,7 @@ class RelatedArtist extends React.Component {
 
   artworksString(counts) {
     const forSale = counts.for_sale_artworks ? counts.for_sale_artworks + ' for sale' : null
-    const totalWorks = counts.artworks ? counts.artworks + ' works' : null
+    const totalWorks = counts.artworks ? counts.artworks + (counts.artworks > 1 ? ' works' : ' work') : null
 
     if (forSale && totalWorks) {
       return totalWorks + ', ' + forSale
--- a/lib/components/artist/related_artists/related_artist.js
+++ b/lib/components/artist/related_artists/related_artist.js
@@ -15,13 +15,13 @@ const imageWidth = windowDimensions.width > 700 ? (windowDimensions.width - 100)
 
 class RelatedArtist extends React.Component {
   handleTap() {
-    SwitchBoard.presentNavigationViewController(this, `/artist/${this.props.artist.href}`);
+    SwitchBoard.presentNavigationViewController(this, ${this.props.artist.href});
   }
 
   render() {
     const artist = this.props.artist;
     return (
-      <TouchableWithoutFeedback onPress={this.handleTap.bind(this)}>
+      <TouchableWithoutFeedback onPress={this.handleTap.bind(this)} delayPressIn={5} delayPressOut={5}>
         <View style={{paddingBottom: 30, width: this.props.relay.variables.imageWidth}}>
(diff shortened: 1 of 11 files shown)


lib/components/artist/shows/metadata.js
```diff
@@ -16,7 +16,7 @@ class Metadata extends React.Component {
         <Text style={styles.sansSerifText}>{this.props.show.partner.name.toUpperCase()}</Text>
         <Text style={styles.sansSerifText}>{this.showTypeString(this.props.show)}</Text>
         <SerifText style={styles.serifText}>{this.props.show.name}</SerifText>
-        { this.dateAndLocationString(this.props.show) }
+        { this.props.show.location ? this.dateAndLocationString() : null }
```
alloy: Why not perform the location null check in that method as well? Probably reads better to have all conditionals in one place.

alloy:
Added another TODO for when all works are for sale.

sarahscott:
I'm going to make a separate PR for the images on iPad because that's going to be its own investigation. Also, @briansw , what should we display if an artist has no shows (see https://www.artsy.net/artist/kaikai-kiki)?

briansw:
@sarahscott is it simple enough to hide the shows tab in that case?

sarahscott:
@briansw  I'll look into it. @alloy might know better.

alloy:
@sarahscott @briansw I don’t think that should be a problem, you have those counts before any view is rendered, so you should be able to use that to determine what tabs to render.

alloy:
Something like:

```js
const tabs = ['ABOUT', 'WORKS'];
if (this.artist.previous_shows.length && this.artist.upcoming_shows.length && this.current_shows) {
  tabs.push('SHOWS');
}
…
<TabView titles={tabs}>
  …
</TabView>
```

You’ll need to pull the show fragments that define the previous, upcoming, and current shows lists from their respective components into the Artist component, otherwise you can’t access them.

lib/components/artist/related_artists/related_artist.js
```diff
@@ -15,13 +15,13 @@ const imageWidth = windowDimensions.width > 700 ? (windowDimensions.width - 100)
...
   handleTap() {
-    SwitchBoard.presentNavigationViewController(this, `/artist/${this.props.artist.href}`);
+    SwitchBoard.presentNavigationViewController(this, this.props.artist.href);
   }
 
   render() {
     const artist = this.props.artist;
     return (
-      <TouchableWithoutFeedback onPress={this.handleTap.bind(this)}>
+      <TouchableWithoutFeedback onPress={this.handleTap.bind(this)} delayPressIn={5} delayPressOut={5}>
```
alloy: So does this help or is this left-over experimenting?

alloy:
Merging including those tap delays, but please let me know what the deal is with them.

lib/components/artist/related_artists/related_artist.js
sarahscott: Woops - yes it's a left-over experiment. Will remove in my next pr.

alloy:
🎉 

Great work, @sarahscott :clap:

PR #77, opened 2016-05-20
[InvertedButton] Fix selected state, ironically, it was inverted.

alloy:


lib/components/buttons/inverted_button.js
```diff
@@ -2,23 +2,26 @@
...
 import colors from '../../../data/colors';
 
 const sideMargin = Dimensions.get('window').width > 700 ? 175 : 0;
 
 export default class InvertedButton extends React.Component {
   render() {
     const styling = {
-      underlayColor: (this.props.selected ? colors['purple-regular'] : 'black'),
-      style: [styles.button, { backgroundColor: (this.props.selected ? 'black' : colors['purple-regular']) }],
+      underlayColor: (this.props.selected ? 'black' : colors['purple-regular']),
```
mennenia: so the selected state now goes to black instead of purple, but not yet greyed out (as is stated in [this comment](https://github.com/artsy/eigen/issues/1132#issuecomment-220580717))

alloy: No it’s purple when selected. The ‘underlay’ is basically the ‘touch highlight’, so it should preview the new state before the user decides to finish the gesture.

mennenia: Yeah figured. Thanks for verifying! :D

PR #80, opened 2016-05-23
[WIP] Proposed solution for tab view

sarahscott:
simulator screen shot may 23 2016 1 35 57 pm

### DIFF
--- a/lib/containers/artist.js
+++ b/lib/containers/artist.js
@@ -37,11 +37,18 @@ class Artist extends React.Component {
   render() {
     const windowDimensions = Dimensions.get('window');
     const commonPadding = windowDimensions.width > 700 ? 40 : 20;
+    const tabs = ['ABOUT'];
+    if (this.props.artist.counts.artworks) {
+      tabs.push('WORKS');
+    }
+    if (this.props.artist.partner_shows.length) {
+      tabs.push('SHOWS');
+    }
     return (
       <ScrollView scrollsToTop={true}>
         <View style={{ paddingLeft: commonPadding, paddingRight: commonPadding }}>
           <Header artist={this.props.artist} />
-          <TabView titles={['ABOUT', 'WORKS', 'SHOWS']} onSelectionChange={this.tabSelectionDidChange}>
+          <TabView titles={tabs} onSelectionChange={this.tabSelectionDidChange}>
             {this.renderTabViewContent()}
           </TabView>
         </View>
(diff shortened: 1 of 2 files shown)


sarahscott:
Needs a rebase; working on it now.

lib/containers/artist.js
```diff
@@ -54,6 +62,12 @@ export default Relay.createContainer(Artist, {
   fragments: {
     artist: () => Relay.QL`
       fragment on Artist {
+        partner_shows {
+          _id
+        }
```
alloy: I worry that this might lead to much more data being fetched, which is why I suggested to use those query fragments that we already use: https://github.com/artsy/emission/pull/76#issuecomment-221089319

However, the best solution imo would be to make metaphysics be able to return a shows count, like you did for the artworks. Have you checked if it doesn’t provide that yet?

sarahscott: Yeah I checked the counts endpoint and it didn't have shows :/

alloy: Ok, I do think that that’s the way to go, so we’ll have to add that ourselves.

lib/containers/artist.js
```diff
@@ -37,11 +37,19 @@ class Artist extends React.Component {
   render() {
     const windowDimensions = Dimensions.get('window');
     const commonPadding = windowDimensions.width > 700 ? 30 : 20;
+    const tabs = ['ABOUT'];
+    debugger;
```
sarahscott: lol I'll get rid of this

lib/containers/artist.js
```diff
@@ -37,11 +37,19 @@ class Artist extends React.Component {
...
+    if (this.props.artist.counts.artworks) {
+      tabs.push('WORKS');
+    }
+    if (this.props.artist.partner_shows.length) {
+      tabs.push('SHOWS');
```
sarahscott: Could there be a case in which an artist has shows but not works? If so, I should change the logic [here](https://github.com/artsy/emission/pull/80/files#diff-f64004b75dbb599d0061fe4024381638L32). If not, it might be better to KISS for now.

lib/containers/artist.js
sarahscott: I could get my feet wet in metaphysics 🐬🐳👣

orta: Do it, me and @ashfurrow  did - https://ashfurrow.com/blog/lets-steal-other-developers-best-ideas/

lib/containers/artist.js
alloy: Good question, I think you have to ask the analytics people for help on that one.

sarahscott: from @anipetrov: to my knowledge, no way to associate artists and shows if not through artworks. i am not sure if the artist is still associated with a show if all of his show artworks get unpublished

The update to that logic would be nice with enums in objc but I'm not sure what is more appropriate stylistically with javascript.

alloy:
Ok, we’re merging this as-is so we can at least test the functionality, but we’ll be cleaning this up next week before the final release.

There _may_ be cases where it breaks if there are shows but no works, but those should be unlikely.

Issue #103, opened 2016-05-24
[ScrollView] Make scroll to top work.


alloy:
From @katarinabatina 

> Tap to top isn't working on iPhone. If I go deep in the works feed on artist view it is very hard to scroll all the way back up. I think this is not working app wide and it should.

alloy:
I bet the fixes in the rest of Eigen are similar, just need to take a few hours to go through them all.

Issue #83, opened 2016-05-24
[Articles] Remove horizontal scroller


alloy:
![image_0](https://cloud.githubusercontent.com/assets/2320/15501272/e7818ab0-21ac-11e6-8198-dd1c60de5f83.jpg)

PR #104, opened 2016-05-24
Scrollview fixes

alloy:
- Scroll to top: #103 
- Remove horizontal scroller from articles component #83

### DIFF
--- a/lib/components/artist/articles/index.js
+++ b/lib/components/artist/articles/index.js
@@ -16,7 +16,9 @@ class Articles extends React.Component {
     return (
       <View>
         <SerifText style={styles.heading}>Featured Articles</SerifText>
-        <ScrollView horizontal={true} style={{ overflow: 'visible', marginBottom:40 }}>
+        <ScrollView horizontal={true}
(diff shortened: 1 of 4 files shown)


lib/components/artist/articles/index.js
```diff
@@ -16,7 +16,10 @@ class Articles extends React.Component {
     return (
       <View>
         <SerifText style={styles.heading}>Featured Articles</SerifText>
-        <ScrollView horizontal={true} style={{ overflow: 'visible', marginBottom:40 }}>
+        <ScrollView horizontal={true}
+                    showsHorizontalScrollIndicator={false}
+                    scrollsToTop={false}
```
mennenia: so... setting it to false in the react code will allow the native code to work as it should, scrolling to the top on a statusbar tap?

mennenia: whoooops nevermind lol, saw the part in the artist.js that sets scroll to top to true :).

alloy: 😄👍

PR #110, opened 2016-05-25
[Header] Added parsing for artist birthday and nationality in header

### LINKED ISSUE #97: [Biography] Don't truncate bio / shorten born to b.
- [x] Shorten born to b.
- [x] Don't truncate bioline. let it wrap to next line.

screen shot 2016-05-24 at 9 06 03 am

### LINKED ISSUE #107: [Header] [Biography] b. vs born vs est.
- [ ] Sometimes the birthyear field contains the word "born" we need to strip the word out when this is the case
- [ ] Sometimes the birthyear field contains the word "Est." in this case we should keep "Est." and not display "b."

img_0033

sarahscott:
Fixes #107

### DIFF
--- a/lib/components/artist/header.js
+++ b/lib/components/artist/header.js
@@ -56,7 +56,7 @@ class Header extends React.Component {
           {artist.name}
         </Headline>
         <SerifText style={styles.base}>
-          {artist.nationality}, b. {artist.birthday}
+          { this.descriptiveString() }
         </SerifText>
         <SerifText style={[styles.base, styles.followCount]}>
           {artist.counts.follows} Followers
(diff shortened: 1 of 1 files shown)


lib/components/artist/header.js
```diff
@@ -67,6 +67,27 @@ class Header extends React.Component {
       </View>
     );
   }
+
+  descriptiveString() {
+    const artist = this.props.artist;
+    const descriptiveString = (artist.nationality ? artist.nationality : "") + this.birthdayString();
```
alloy: Instead of the ternary operator, you can also use `||`, because in JS the result value of a conditional expression is returned and in JS `null` (amongst `false`, `0`, and `""` are considered ‘falsy’). i.e.

```js
const descriptiveString = (artist.nationality || "") + this.birthdayString();
```

http://programmers.stackexchange.com/a/82599

alloy:
Do you have some artists that you know showcase the various variants?

Also, this is the type of logic we should extract post v1 into separate functions that we can then easily unit test. #113

orta:
There's a reasonable argument that this logic could/should go in metaphysics

alloy:
@orta Yeah, totes, but I want to start some form of workflow for this, i.e. we write such functions tested and well in the app (Emission) first, then when others in C&GMV need it too we can very easily move it into metaphysics as it’s already well factored.

orta:
This function probably _does_ exist in force for their artist page though FWIW ( at a guess) as they've shipped it already, if we're the ones coming along later, no-one will ask for it

alloy:
They don’t appear to yet handle this case: https://www.artsy.net/artist/bernadette-corporation

screen shot 2016-05-25 at 14 31 22

/cc @broskoski @briansw

alloy:
> if we're the ones coming along later, no-one will ask for it

…but otherwise you make a good point 👍

briansw:
@alloy I think force probably handled this on the client side prior to the most recent update which reverted the behavior—there's an open ticket to correct it. 

We'll need this formatting anywhere we display the artist cards as well (/artists page in force, related artist modules, etc) if that's useful in determining whether or not to put it in metaphysics.

lib/components/artist/header.js
sarahscott: This is very helpful 😄

PR #112, opened 2016-05-25
Miscellaneous QA

### LINKED ISSUE #84: [Shows] Spurious whitespace around show location.
This is not the case with all shows, so maybe this entry simply has a trailing whitespace? If that’s possible, then we should strips leading/trailing whitespace.

img_7796

### LINKED ISSUE #92: [Biography] Type too small on artist bios
current:

image

spec:
image

The typeface seems smaller than usual, it should be 16px.

### LINKED ISSUE #93: [Header] Increase top padding on artist name header to 20px
current:
image

spec: 

image

sarahscott:
Too many PRs...

Fixes #93 
Fixes #92 
Fixes #84

### DIFF
--- a/lib/components/text/serif.js
+++ b/lib/components/text/serif.js
@@ -21,7 +21,7 @@ Serif.propTypes = {
 
 const styles = StyleSheet.create({
   default: {
-    fontSize: 16
+    fontSize: 17
   },
   required: {
     fontFamily: 'AGaramondPro-Regular',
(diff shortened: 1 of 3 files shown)


lib/components/text/serif.js
```diff
@@ -21,7 +21,7 @@ Serif.propTypes = {
 
 const styles = StyleSheet.create({
   default: {
-    fontSize: 16
+    fontSize: 17
```
alloy: How did you arrive at using 17 points instead of 16?

sarahscott: I eyeballed the spec and Katarina's issue, and I think this subtle change looks more like what it's supposed to.

Issue #113, opened 2016-05-25
[Unit Test] Extract logic functions from view components and test


alloy:
These should probably not take any objects, but just the required values.
- [ ] https://github.com/artsy/emission/pull/110
- [ ] https://github.com/artsy/emission/pull/109

PR #114, opened 2016-05-25
[Header] Hide nationality / birthday if none on record

### LINKED ISSUE #86: [Header] Don’t show birth place/date if unavailable
There are also cases where only 1 of the 2 fields is available. E.g. Banksy’s birth date is unknown.

img_7797

mennenia:
Fix #86.

screen shot 2016-05-25 at 15 57 15


mennenia:
Will redo this with the current state in master!

lib/components/artist/header.js
```diff
@@ -57,16 +57,31 @@ class Header extends React.Component {
...
+    if (bylineRequired) {
+      return ( 
+        <View>
+            <SerifText style={styles.base}>
+                {nationality ? nationalityString : null}{birthday ? birthdayString : null}
+              </SerifText>
+        </View>
+      );
+    } else {
+      return (null);
```
alloy: The parentheses are only needed around the JSX syntax and only when it spans multiple lines.

lib/components/artist/header.js
```diff
@@ -57,16 +57,31 @@ class Header extends React.Component {
...
+  
+  renderSection(nationality, birthday) {
+    const bylineRequired = (nationality || birthday);
+    const nationalityString = nationality + (birthday ? ", " : "");
+    const birthdayString = "b. " + birthday;
+    if (bylineRequired) {
+      return ( 
+        <View>
+            <SerifText style={styles.base}>
+                {nationality ? nationalityString : null}{birthday ? birthdayString : null}
```
alloy: Probably nicer to read to keep as much logic out of the interpolation.

lib/components/artist/header.js
```diff
@@ -57,16 +57,31 @@ class Header extends React.Component {
         <Headline style={[styles.base, styles.headline]}>
           {artist.name}
         </Headline>
-        <SerifText style={styles.base}>
-          {artist.nationality}, b. {artist.birthday}
-        </SerifText>
+        {this.renderSection(artist.nationality, artist.birthday)}
```
alloy: Until we extract functions from components (#113), we’ve been following the style to just access `this.artist` from the `renderSection` method.

lib/components/artist/biography.js
```diff
@@ -21,7 +21,7 @@ class Biography extends React.Component {
 
   render() {
     const artist = this.props.artist;
-    if (artist.blurb.length === 0 && artist.bio.length === 0) { return null; }
+    if (!artist.blurb && !artist.bio) { return null; }
```
mennenia: This was causing an error in RN when I tested it with the artist `ry-david-bradley`, because the blurb was `null`. Would there be a case where there was a blurb, but it'd be empty? If so we would need both checks I'm assuming..

alloy: @sarahscott Do you know? I would hope it would just be `null` and not blank.

sarahscott: Hmm I would hope so too. I guess it depends on metaphysics but it seems much more likely that it would be `null` instead of an empty string so this solution makes sense to me.

orta: could you do `artist.blurb?.length` to account for both null & zero length? I think that's legit ES6

alloy: Really? I don’t think so, maybe it’s a TypeScript thing?

orta: [Confirmed, nope](http://stackoverflow.com/questions/32139078/null-safe-property-access-in-es6-es2015)

mennenia:
@sarahscott Orta's suggestion seems to not be possible. I'm not sure what I should further change for this PR? I too was assuming that the bio would be `null` not just an empty string. The reason for checking is that someone perhaps came across that exact error, and hence put the clause in there for a reason :).

Either way, let me know what you'd like me to do. Personally, I think this is ok for now.

mennenia:
Self-merging, mainly for the blurb / null fix!

alloy:
:tada:

PR #117, opened 2016-05-25
Update to React Native 0.26.2 and React 15.0.2

### LINKED ISSUE #9: Update to RN 0.25
I tried doing that as part of my PR #8, but it seems that it re-breaks my simple workaround for the image view getting a zero size rect.

alloy:
Fixes #9 

Alas it did not fix the tap issue.

Biggest change is that you now need to import from either the React or React Native modules. See a9a51d9a1398a268bbb936cad936122150f7a186.


alloy:
I’m self-merging this before others make PRs and this would need to be rebased, which would be a hell.

Issue #89, opened 2016-05-24
[Tab Component] Missing gray rules


katarinabatina:
current:

![image](https://cloud.githubusercontent.com/assets/6482926/15502894/ac9d659a-2184-11e6-9a8a-0bdb7fc14cc0.png)

spec:

![image](https://cloud.githubusercontent.com/assets/6482926/15502923/dce4e3c2-2184-11e6-87d1-f3a7374f114b.png)

mennenia:
Extraction issue, and hence captured [here](https://github.com/artsy/extraction/issues/1). Closing this ticket in favour of the other one :).

PR #120, opened 2016-05-26
[SwitchView] Add top and bottom indicators for unselected tabs

mennenia:
"Fix" #89 

screen shot 2016-05-26 at 11 32 36

The unselected button's text is too dark still and should be `gray-regular`. Nothing in RN seems to change that, after some digging think it's within the original `ARSwitchView` in Eigen. Thoughts?

### DIFF
--- a/lib/components/switch_view.js
+++ b/lib/components/switch_view.js
@@ -4,11 +4,13 @@
 import React from 'react';
 import { requireNativeComponent } from 'react-native';
 
+import colors from '../../data/colors';
+
 export default class SwitchView extends React.Component {
   render() {
     const { style, ...props } = this.props;
     // Height taken from ARSwitchView.m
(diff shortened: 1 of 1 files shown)


alloy:
ARSwitchView resides in Extraction: https://github.com/artsy/extraction/blob/master/Extraction/Classes/ARSwitchView.m

It’s weird that no snapshots broke in Eigen due to this.

I bet that [this](https://github.com/artsy/extraction/blob/7312df7c9c005c196705cc9da7c5d53cde404854/Extraction/Classes/ARSwitchView.m#L128) is basically what you did in this PR. Can you double-check why it’s not setting the right colors there instead?

mennenia:
`ARSwitchView` is not up to spec everywhere else in the app, so moving it out of Emission and into Extraction.

PR #121, opened 2016-05-26
Fix crash when artist has no image.

### LINKED ISSUE #94: [Artist View] App crashes when requesting artist view
App crashes when trying to view Michael Williams (https://www.artsy.net/artist/michael-williams)

alloy:
Fixes #94.

Also immediately show the placeholder color.


mennenia:
🎉

PR #124, opened 2016-05-26
Update React Native to 0.27.0-rc1

alloy:
This is so that debugging works again.

https://github.com/facebook/react-native/issues/7707


alloy:
Self-merging before I go offline!

PR #125, opened 2016-05-27
[SwitchBoard] Fix tap handling lag.

### LINKED ISSUE #88: [React Native] Fix issue of an initial tap on most of our touchables not working.

alloy:
Fixes #88.

The issue ended up being that the API I was using to map a React ‘component tag’ to a view was not the fast path. In some cases it would just not work at all, I’m unsure why, but haven’t delved into that further. I found a different API that does nothing async, so we get an immediate result and I also chose to get rid of further async dispatching in our switchboard module.


mennenia:
Most awesomest! 🎉

Issue #137, opened 2016-05-30
Think of way to more declaratively add analytics to components.


alloy:
In Eigen we have [a source file](https://github.com/artsy/eigen/blob/master/Artsy/App/ARAppDelegate+Analytics.m) that lives outside of the tracked views and declaratively describes all of the events that need to be tracked.

Some quick googling for existing React solutions:
- https://github.com/yahoo/react-i13n
- https://github.com/nfl/react-metrics

sarahscott:
@mennenia can I close this?

mennenia:
probably yes, there are more recent tickets now :) thanks @sarahscott!

PR #138, opened 2016-05-30
[Events] Add module that vents events for analytics purposes.

alloy:


lib/components/artist/header.js
```diff
@@ -43,6 +45,15 @@ class Header extends React.Component {
     ARTemporaryAPIModule.setFollowStatus(!this.state.following, this.props.artist._id, (error, following) => {
       if (error) {
         __DEV__ ? console.error(error) : console.log(error);
+      } else {
+        Events.postEvent(this, {
+          name: following ? 'Follow artist' : 'Unfollow artist',
```
mennenia: Will we eventually have a place / list for all our events with the constants you think?

mennenia: Found it... lol #137

mennenia:
Looks great!

PR #141, opened 2016-05-31
[Show] Guard against missing image

### LINKED ISSUE #139: [Shows] Crash when no meta_image is available
This show has none https://www.artsy.net/show/serpentine-galleries-alex-katz. Looking at the shows for Alex Katz on Force doesn’t seem to reveal this show at all: https://www.artsy.net/artist/alex-katz/shows.

So either Force is not showing shows where the `meta_image` is missing or they are filtering out this show based on another flag that I have not been able to figure out, based on [the data in metaphysics](https://metaphysics-production.artsy.net/?query=%7B%0A%09partner_show(id%3A%20%22serpentine-galleries-alex-katz%22)%20%7B%0A%20%20%20%20id%0A%20%20%20%20name%0A%20%20%20%20description%0A%20%20%20%20is_fair_booth%0A%20%20%20%20status%0A%20%20%20%20images(size%3A%201)%20%7B%0A%20%20%20%20%20%20url%0A%20%20%20%20%7D%0A%20%20%20%20meta_image%20%7B%0A%20%20%20%20%20%20href%0A%20%20%20%20%7D%0A%20%20%20%20cover_image%20%7B%0A%20%20%20%20%20%20href%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D) at least.

We shouldn’t even crash, but I would like to know if this is simply a problem with the show metadata that needs to be fixed or that we need to be doing something differently.

alloy:
Fixes #139

### DIFF
--- a/Example/Emission/AppDelegate.m
+++ b/Example/Emission/AppDelegate.m
@@ -10,6 +10,8 @@
 #import <React/RCTUtils.h>
 #import <TargetConditionals.h>
 
+#define ARTIST @"banksy"
+
 #if TARGET_OS_SIMULATOR
 #define ENABLE_DEV_MODE
 #endif
@@ -83,7 +85,7 @@ randomBOOL(void)
     NSLog(@"[Event] %@ - %@", fromViewController.class, info);
   };
 
-  ARArtistComponentViewController *artistViewController = [[ARArtistComponentViewController alloc] initWithArtistID:@"banksy"];
+  ARArtistComponentViewController *artistViewController = [[ARArtistComponentViewController alloc] initWithArtistID:ARTIST];
 
   self.window = [[UIWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
   self.window.backgroundColor = [UIColor whiteColor];
--- a/lib/components/artist/shows/large_image_show.js
+++ b/lib/components/artist/shows/large_image_show.js
@@ -22,11 +22,14 @@ class LargeImageShow extends React.Component {
   }
 
   render() {
+    const show = this.props.show;
+    const image = show.meta_image;
+    const imageURL = image && image.cropped.url;
     return (
       <TouchableWithoutFeedback onPress={this.handleTap.bind(this)}>
         <View style={{margin: 10, width: this.props.relay.variables.imageWidth }}>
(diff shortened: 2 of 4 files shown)


lib/components/artist/shows/small_image_show.js
```diff
@@ -16,10 +16,12 @@ class SmallImageShow extends React.Component {
...
+    const image = show.meta_image;
+    const imageURL = image && image.cropped.url;
     return (
       <TouchableWithoutFeedback onPress={this.handleTap.bind(this)} >
         <View style={styles.container}>
-          <OpaqueImageView imageURL={show.meta_image.cropped.url} style={styles.image} />
+          <OpaqueImageView imageURL={imageURL} style={styles.image} />
           <ShowMetadata show={show} />
         </View>
       </TouchableWithoutFeedback>
```
alloy: @sarahscott These three components are all the same in terms of the components they use and their JSX code, the only difference is styling. I think that can be refactored so that only the styling needs to be specified for each individual variant, but the JSX can stay the same and in 1 source file.

lib/components/artist/shows/large_image_show.js
```diff
@@ -22,11 +22,14 @@ class LargeImageShow extends React.Component {
   }
 
   render() {
+    const show = this.props.show;
+    const image = show.meta_image;
+    const imageURL = image && image.cropped.url;
```
alloy: @sarahscott @mennenia Another way to deal with null-ing, this is similar to `foo ?: fallback` in C, but that shorthand of the ternary. See https://repl.it/CWx3

Example/Emission/AppDelegate.m
```diff
@@ -10,6 +10,8 @@
 #import <React/RCTUtils.h>
 #import <TargetConditionals.h>
 
+#define ARTIST @"banksy"
```
orta: SHOUTY

alloy: Hey, it stood out, didn’t it? ;)

lib/components/artist/shows/large_image_show.js
mennenia: 💥

PR #142, opened 2016-05-31
[SwitchView] Set default tab selection to 'Works'

### LINKED ISSUE #90: [Artist] Initial selected tab should be ‘WORKS’
Open artist view, land on about tab, we should adjust this so that the work tab is the default selection when the view is opened.

mennenia:
Fix #90.

screen shot 2016-05-31 at 14 25 50

### DIFF
--- a/Example/Emission/AppDelegate.m
+++ b/Example/Emission/AppDelegate.m
@@ -10,6 +10,8 @@
 #import <React/RCTUtils.h>
 #import <TargetConditionals.h>
 
+#define ARTIST @"banksy"
+
 #if TARGET_OS_SIMULATOR
 #define ENABLE_DEV_MODE
 #endif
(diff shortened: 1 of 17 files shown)


alloy:
I’m not merging this yet, seeing as it apparently can break on iPad, or anywhere, really, because we can’t assume order or assigned properties.

Maybe what we should do is:
* in `setSelectedIndex:animated:`, check if there are any titles ​_at all_​ and, if not, just store the index on the ivar and then return early: https://github.com/artsy/extraction/blob/069fdaeacf31dc599f05e7825980b0e6dff43c47/Extraction/Classes/ARSwitchView.m#L192
* then in `setTitles:` it should already assign the existing selected index value on itself again: https://github.com/artsy/extraction/blob/069fdaeacf31dc599f05e7825980b0e6dff43c47/Extraction/Classes/ARSwitchView.m#L112

mennenia:
> in setSelectedIndex:animated:, check if there are any titles ​at all​ and, if not, just store the index on the ivar and then return early:

Done

then in setTitles: it should already assign the existing selected index value on itself again: https://github.com/artsy/extraction/blob/069fdaeacf31dc599f05e7825980b0e6dff43c47/Extraction/Classes/ARSwitchView.m#L112

Done, but requires us to take out that early return clause we had in there, because the selected index doesn't change in that case, however you need to update it. 

So you think that's ok?

mennenia:
This describes what I did so far: https://github.com/artsy/extraction/pull/3/commits/ed45e4ef7c2a34ae5d0eaa4a596cb102fb323f09

Issue #101, opened 2016-05-24
[Related Artists] Left align entries


briansw:
Left align entries so odd numbers don't center in the content area.

![screen shot 2016-05-24 at 9 35 42 am](https://cloud.githubusercontent.com/assets/1022172/15505786/f61b2294-2192-11e6-93fb-1acb613e1144.png)

PR #143, opened 2016-05-31
[Related Artists] Fix left margin of grid view

### LINKED ISSUE #131: [Related Artists] Grid not centered
Reported by @sarahscott.

image_0-6

mennenia:
Fix #131 

cc @katarinabatina, this is what I went with:

screen shot 2016-05-31 at 15 30 44

### DIFF
--- a/lib/components/artist/related_artists/index.js
+++ b/lib/components/artist/related_artists/index.js
@@ -3,24 +3,35 @@
 
 import Relay from 'react-relay';
 import React from 'react';
-import { ScrollView, StyleSheet, View } from 'react-native';
+import { ScrollView, StyleSheet, View, Dimensions } from 'react-native';
 
 import colors from '../../../../data/colors';
 import SerifText from '../../text/serif';
 import ImageView from '../../opaque_image_view';
 import RelatedArtist from './related_artist'
 
+const isPhone = Dimensions.get('window').width < 700;
+
 class RelatedArtists extends React.Component {
   render() {
-    const artists = this.props.artists
+    const artists = this.props.artists;
     return (
       <View style={styles.container}>
         <SerifText style={styles.heading}>Related Artists</SerifText>
         <View style={styles.artistContainer}>
-          { artists.map(artist => <RelatedArtist key={artist.id} artist={artist} /> ) }
+          { this.renderArtists() }
         </View>
       </View>
     );
+  }
+
+  renderArtists() {
+    const artists = this.props.artists;
+    var artistViews = artists.map(artist => <RelatedArtist key={artist.id} artist={artist} />);
+    if (isPhone && artists.length % 2 !== 0) {
+      artistViews.push(<RelatedArtist key={'dummy'} artist={null}/>);
+    }
+    return artistViews;
   }
 }
 
--- a/lib/components/artist/related_artists/index.js
+++ b/lib/components/artist/related_artists/index.js
@@ -10,8 +10,6 @@ import SerifText from '../../text/serif';
 import ImageView from '../../opaque_image_view';
 import RelatedArtist from './related_artist'
 
-const isPhone = Dimensions.get('window').width < 700;
-
 class RelatedArtists extends React.Component {
   render() {
     const artists = this.props.artists;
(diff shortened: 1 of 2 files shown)


katarinabatina:
image

mennenia:
@katarinabatina updated!

@sarahscott the magical code added property was `space-around` on `justifyContent` rather than `flex-start`

screen shot 2016-05-31 at 17 10 38

katarinabatina:
@mennenia 💯  magic! 👏

alloy:
Awesome, well done 👍

sarahscott:
@mennenia The only problem here is that the last item might be centered instead of aligned left - do you have any screenshots with the last row when there's an item on its own?

sarahscott:
See #101

sarahscott:
I don't know if this is relevant for us but someone else suggests using a spacer: http://stackoverflow.com/questions/18744164/flex-box-align-last-row-to-grid

alloy:
@sarahscott Ah, so this is basically what you had previously? If all else fails in a timely manner (we need to release this week) then I'm not at all opposed to adding a dummy element.

lib/components/artist/related_artists/index.js
```diff
@@ -3,25 +3,36 @@
...
         <View style={styles.artistContainer}>
-          { artists.map(artist => <RelatedArtist key={artist.id} artist={artist} /> ) }
+          { this.renderArtists() }
         </View>
       </View>
     );
   }
+
+  renderArtists() {
+    const artists = this.props.artists;
```
sarahscott: @mennenia this is just an illustrative commit to show the spacer idea - how do you feel about this? It fixes the issue for iPhone: 
simulator screen shot may 31 2016 5 23 13 pm

but not for iPad, so we'll have to do something else for that:

simulator screen shot may 31 2016 5 24 59 pm

mennenia:
Current situation:

ipad

simulator screen shot 1 jun 2016 18 35 58

lib/components/artist/related_artists/index.js
```diff
@@ -3,35 +3,67 @@
...
 class RelatedArtists extends React.Component {
+    constructor(props) {
+    super(props);
+    this.state = { columns: this.columnCount()};
+    this.onLayout = this.onLayout.bind(this);
+  }
+  
+  columnCount() {
+    const isPhone = Dimensions.get('window').width < 700;
+    const isPadHorizontal = Dimensions.get('window').width > 1000;
```
alloy: Maybe reduce method calls a bit more by storing the `window` object?

(Not sure about this case, but might be relevant in cases where behind the scenes it calls a native module and it needs to go over the bridge back and forth multiple times.)

lib/components/artist/related_artists/index.js
```diff
@@ -3,35 +3,67 @@
...
+    return !isPhone ? isPadHorizontal ? 4 : 3 : 2;
```
alloy: Can you add parentheses around the nested ternary?

sarahscott:
rebase reminder 😁

katarinabatina:
@mennenia magical magic of magicness 👏 👏 👏 👏  kudos!

mennenia:
Cheers @katarinabatina ^_^ this was proper team effort from @sarahscott and @alloy also :D.

@sarahscott - comments addressed, master merged. Good to go on my end :)!

sarahscott:
merging!

lib/components/artist/related_artists/index.js
```diff
@@ -10,6 +10,8 @@ import SerifText from '../../text/serif';
 import ImageView from '../../opaque_image_view';
 import RelatedArtist from './related_artist'
 
+const window = Dimensions.get('window')
```
alloy: Interesting, so the `width` value still updates after orientation changes?

I actually meant to do it once inside the `columnCount` method, but if this works that’s even better and we can do this in other places too.

mennenia: yeah we tested it and it worked, so, yay! :)

lib/components/artist/related_artists/index.js
```diff
@@ -18,23 +20,21 @@ class RelatedArtists extends React.Component {
   }
   
   columnCount() {
-    const isPhone = Dimensions.get('window').width < 700;
-    const isPadHorizontal = Dimensions.get('window').width > 1000;
-    return !isPhone ? isPadHorizontal ? 4 : 3 : 2;
+    const isPhone = window.width < 700;
+    const isPadHorizontal = window.width > 1000;
+    return !isPhone ? (isPadHorizontal ? 4 : 3) : 2;
```
alloy: In the future it’s better to reduce the amount of negation, as it will be more readable.

lib/components/artist/related_artists/index.js
alloy: Noice! I wonder how that works under the hood, will have to look into that some time.

PR #144, opened 2016-06-01
[ImageView] Load cropped images that take screen pixel ratio into account.

### LINKED ISSUE #13: [Image] Get images sized according to screen DPI

alloy:
Fixes #13

### DIFF
--- a/lib/components/artist/shows/medium_image_show.js
+++ b/lib/components/artist/shows/medium_image_show.js
@@ -11,9 +11,6 @@ import OpaqueImageView from '../../opaque_image_view'
 import ShowMetadata from './metadata'
 import SwitchBoard from '../../../modules/switch_board';
 
-const windowDimensions = Dimensions.get('window');
-const imageWidth = (windowDimensions.width - 120)/3;
-
 class MediumImageShow extends React.Component {
   handleTap() {
     SwitchBoard.presentNavigationViewController(this, this.props.show.href);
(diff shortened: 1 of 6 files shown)


lib/components/artist/shows/medium_image_show.js
```diff
@@ -11,37 +11,35 @@ import OpaqueImageView from '../../opaque_image_view'
...
-
   fragments: {
     show: () => Relay.QL`
       fragment on PartnerShow {
         href
         meta_image {
-          cropped(width: $imageWidth, height: $imageHeight) {
-            url
-          }
+          url(version: "large")
```
mennenia: so we're using the large image url everywhere, and just sizing the view accordingly for large, medium, and small images, right?

mennenia:
:+1:

lib/components/artist/shows/medium_image_show.js
alloy: Yeah, we’re using the URL for the large version to ask [Gemini](https://github.com/artsy/gemini) for a scaled/cropped version at a specific size. This is the same as metaphysics does when asking for a specific size, but we don’t know the exact sizes we need ahead of time so can’t really do that from Relay.

mennenia: Awesome. That's what I gathered, just wanted to verify ^_^.

alloy:
Rebased and good to merge 👍

PR #149, opened 2016-06-02
[SwitchView] Set default tab selection to 'Works'

### LINKED ISSUE #90: [Artist] Initial selected tab should be ‘WORKS’

mennenia:
Fix #90, history of this PR in #142.


alloy:
👍

PR #150, opened 2016-06-02
[Show] Use `cover_image`

alloy:
It turns out that `meta_image` is meant to be used for items that are shared, e.g. on Twitter/Facebook, where partners can indicate that some images may not be shared.

I also refactored the various show components into a single one that leaves styling up to the parent. This is slightly awkward, but it also makes it so that the cell dimensions only need to be calculated _once_ instead of for every cell.


Example/Emission/AppDelegate.m
```diff
@@ -10,7 +10,7 @@
 #import <React/RCTUtils.h>
 #import <TargetConditionals.h>
 
-#define ARTIST @"banksy"
+#define ARTIST @"alex-katz"
```
mennenia: Forgot to change this? No prob, we can switch it up from Banksy from time to time!

mennenia:
Looks awesome! Like the new style.

Example/Emission/AppDelegate.m
alloy: Yeah, I did :D But then I found some QA issues with this artist, so decided to leave it in for now.

lib/components/artist/shows/small_list.js
```diff
@@ -46,6 +46,21 @@ const styles = StyleSheet.create({
...
+    width: 75,
+    height: 75,
+    marginBottom: 20,
+    marginTop: 20,
+    marginRight: 15
+  },
+});
+
+
 export default Relay.createContainer(SmallList, {
```
mennenia:
```
export default Relay.createContainer(SmallList, {
  fragments: {
    shows: () => Relay.QL`
      fragment on PartnerShow @relay(plural: true) {
        ${SmallImageShow.getFragment('show')}
      }
    `,
  }
```

should be 

```
export default Relay.createContainer(SmallList, {
  fragments: {
    shows: () => Relay.QL`
      fragment on PartnerShow @relay(plural: true) {
        ${Show.getFragment('show')}
      }
    `,
  }
```

alloy: Wow, how is this still working…

Great catch!

alloy: Doh, I didn’t test on iPhone :sob:

Issue #155, opened 2016-06-02
[iPad] Past Shows issues.


alloy:
## When loading shows in portrait orientation and then rotating to landscape orientation the grid changes to 4 columns.

It should always be 3 columns.

![simulator screen shot 02 jun 2016 15 22 45](https://cloud.githubusercontent.com/assets/2320/15746318/593294bc-28d6-11e6-833a-7134f0fde03a.png)

## When loading shows in landscape orientation and then rotating to portrait orientation the columns don’t decrease in size and changes to 2 columns.

![simulator screen shot 02 jun 2016 15 36 58](https://cloud.githubusercontent.com/assets/2320/15746682/db942e24-28d7-11e6-9b98-d0d7e24aec0d.png)

mennenia:
Update here. @briansw stated that ideally we do have four columns in landscape on iPad.

Also, the second issue seems to be the same as the relatedArtist issue #152, so if that is fixed this can be fixed probably with the same approach.

Issue #154, opened 2016-06-02
[iPad] Current/Upcoming Shows issues


alloy:
## When loading shows in portrait orientation and then rotating to landscape orientation the columns don’t increase in size

They should fill the width.

![simulator screen shot 02 jun 2016 15 31 19](https://cloud.githubusercontent.com/assets/2320/15746486/11ae12c8-28d7-11e6-8399-36cd123b09dc.png)

## When loading shows in landscape orientation and then rotating to portrait orientation the columns don’t decrease in size and changes to 1 column.

They should shrink to fit 2 columns.

![simulator screen shot 02 jun 2016 15 31 54](https://cloud.githubusercontent.com/assets/2320/15746534/48d4f9b0-28d7-11e6-977d-2b90576b1284.png)

PR #156, opened 2016-06-02
[RelatedArtists] iPad fixes

### LINKED ISSUE #152: [iPad] Related artists grid issues.
- [x] Fix trailing artist alignment
- [x] Decide: 3 or 4 columns?
- [ ] Fix: Different column size in orientation changes

## When there’s a single trailing artist it doesn’t align to the left:

This can be seen with the artist `alex-katz`.

simulator screen shot 02 jun 2016 14 41 50

## Should have 3 columns in each orientation

According to [the design spec](https://github.com/artsy/eigen/issues/1136), it should also have 3 columns in landscape orientation. I’m not sure what @sarahscott based the 4 columns on, maybe there’s another comment elsewhere?

## When loading ‘ABOUT’ tab in landscape orientation and then changing to portrait orientation changes related artists to 2 columns.

How this should be fixed kinda depends on whether or not there are always supposed to be 3 columns. _(See above.)_

simulator screen shot 02 jun 2016 14 50 40

mennenia:
This fixes #152, #154, #155. 

Current status: @sarahscott is awesome!


mennenia:
I think the issue is the new imagewidth? 

I now got it to have 4 related artists on load when in landscape, but it still goes to only two columns on portrait. 

Logging shows that the related artists widths get recalculated from portrait to landscape (2 logs, as expected), but not when you start the app in landscape and then go to portrait (only 1 log, the initial one)

Unsure what's going on yet. Going to have dinner and will be back.

sarahscott:
I think I got it!

ipadrotationsuccess

lib/components/artist/related_artists/index.js
```diff
@@ -18,20 +18,34 @@ class RelatedArtists extends React.Component {
...
   layoutState() {
-    const window = Dimensions.get('window');
-    const isPad = window.width > 700;
-    const isPadHorizontal = window.width > 1000;
+    const width = Dimensions.get('window').width;
+    const isPad = width > 700;
+    const isPadHorizontal = width > 1000;
 
-    const columnCount = (isPad ? (isPadHorizontal ? 4 : 3) : 2);
+    // TODO: Document what these margins are based on.
```
alloy: @sarahscott @mennenia Would love to have this sorted. They are really magic numbers right now. I’d also prefer to distinguish between side-margins and in-between margins and then calculate the column width based on that.

alloy:
Alright, let the festivities begin! 🎉

lib/components/artist/related_artists/index.js
mennenia: Sounds good. I have zero clue about the margins T_T.

PR #158, opened 2016-06-02
[Events] Actually fetch required event data and expose the artist VC’s artist ID.

alloy:
Kinda silly of me to try to capture analytics data that we didn’t actually fetch.


sarahscott:
😄

PR #159, opened 2016-06-03
Hides about section if it shouldn't exist, and hides tab bar altogether if there isn't anything to show

sarahscott:
Using dummy variables until https://github.com/artsy/metaphysics/pull/298 is merged.

- [x] Hides tab bar if there's nothing to show at all
- [x] Hides About section if artist has no metadata
- [x] Change dummy variables to actual data

simulator screen shot jun 2 2016 11 50 04 pm

simulator screen shot jun 3 2016 3 55 27 pm

### DIFF
--- a/lib/containers/artist.js
+++ b/lib/containers/artist.js
@@ -22,14 +22,19 @@ const TABS = {
 };
 
 class Artist extends React.Component {
-  constructor() {
-    super();
-    this.state = { selectedTabIndex: 1 };
+  constructor(props) {
+    super(props);
 
     this.availableTabs = this.availableTabs.bind(this);
     this.selectedTabTitle = this.selectedTabTitle.bind(this);
     this.tabSelectionDidChange = this.tabSelectionDidChange.bind(this);
     this.viewForSelectedTab = this.viewForSelectedTab.bind(this);
+
+    if (this.availableTabs().includes(TABS.WORKS)) {
+      this.state = { selectedTabIndex: this.availableTabs().indexOf(TABS.WORKS) };
+    } else {
+      this.state = { selectedTabIndex: 0 };
+    }
   }
 
   tabSelectionDidChange(event: TabSelectionEvent) {
--- a/lib/containers/artist.js
+++ b/lib/containers/artist.js
@@ -37,12 +37,18 @@ class Artist extends React.Component {
   }
 
   availableTabs() {
-    const tabs = [TABS.ABOUT];
+    const tabs = [];
+    const hasMetadata = true;
+
+    if (hasMetadata) {
+      tabs.push(TABS.ABOUT);
+    }
+
     if (this.props.artist.counts.artworks) {
       tabs.push(TABS.WORKS);
     }
-    // TODO Replace this with a count that we get back from metaphysics.
-    if (this.props.artist.partner_shows.length) {
+
+    if (this.props.artist.counts.partner_shows) {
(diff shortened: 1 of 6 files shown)


sarahscott:
@alloy or @mennenia , this is almost ready for review but I get a strange crash when I try it out with an artist who has a `bio` but not anything else (i.e. https://www.artsy.net/artist/natalie-jeremijenko). The Xcode error is 

`
2016-06-03 13:58:07.887 Emission[18722:23010013] *** Assertion failure in +[UIView chainViews:usingBlock:], /Users/sarahscott/Desktop/dev/emission/Example/Pods/FLKAutoLayout/FLKAutoLayout/UIView+FLKAutoLayout.m:297
2016-06-03 13:58:07.892 Emission[18722:23010013] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Operations on multiple views require at least 2 views'
`

Looking into it now.

sarahscott:
Ready for review!

lib/containers/artist.js
```diff
@@ -74,13 +81,18 @@ class Artist extends React.Component {
...
           <Header artist={this.props.artist} />
-          <TabView titles={this.availableTabs()} selectedIndex={this.state.selectedTabIndex} onSelectionChange={this.tabSelectionDidChange}>
-            {this.renderTabViewContent()}
-          </TabView>
+          {
+            displayTabView ?
+            <TabView titles={this.availableTabs()} selectedIndex={this.state.selectedTabIndex} onSelectionChange={this.tabSelectionDidChange}>
+              { this.renderTabViewContent() }
+            </TabView> : null
+          }
```
alloy: How are the works still being shown?

sarahscott: The works themselves should still be shown here: https://github.com/artsy/emission/pull/159/files#diff-f64004b75dbb599d0061fe4024381638R76

alloy: Hmm, I don’t get it, those are children of the `TabView`, so how are they still showing up if we don’t show the `TabView`?

lib/containers/artist.js
```diff
@@ -22,14 +22,19 @@ const TABS = {
...
+  constructor(props) {
     super();
-    this.state = { selectedTabIndex: 1 };
 
     this.availableTabs = this.availableTabs.bind(this);
     this.selectedTabTitle = this.selectedTabTitle.bind(this);
     this.tabSelectionDidChange = this.tabSelectionDidChange.bind(this);
     this.viewForSelectedTab = this.viewForSelectedTab.bind(this);
+
+    if (this.availableTabs().contains(TABS.WORKS)) {
```
sarahscott: This doesn't work because `availableTabs` relies on this.props and they're undefined at this points. I know I can get them passed into the constructor but I can't pass them to `availableTabs`.

lib/containers/artist.js
```diff
@@ -22,14 +22,19 @@ const TABS = {
 };
 
 class Artist extends React.Component {
-  constructor() {
+  constructor(props) {
     super();
```
alloy: You need to pass the props to super.

lib/containers/artist.js
alloy: Might just be because you weren’t passing the props to super above?

alloy:
simulator screen shot 04 jun 2016 01 35 18

alloy:
Merging!

PR #170, opened 2016-06-07
[Artworks] Do not fetch the same page multiple times.

alloy:
Turns out that it was trivial to fetch a page multiple times, leading to duplicate works.


mennenia:
beautiful PR haha!

Issue #167, opened 2016-06-06
Biography line spacing is too large


katarinabatina:
current line spacing
looks like type size of 16 with 1.75 line spacing

![image](https://cloud.githubusercontent.com/assets/6482926/15836456/efe0f438-2c02-11e6-9f8d-fb016ede93d0.png)

should be

16 with a line spacing of 1.35

![image](https://cloud.githubusercontent.com/assets/6482926/15836479/038d78ee-2c03-11e6-80ba-d0784ad0f62b.png)

Issue #164, opened 2016-06-06
Spacing adjustments for Show Tab


katarinabatina:
![image](https://cloud.githubusercontent.com/assets/6482926/15836193/9e015bb8-2c01-11e6-86a3-87733282c11e.png)

Issue #163, opened 2016-06-06
Spacing adjustments for About tab


katarinabatina:
![image](https://cloud.githubusercontent.com/assets/6482926/15835791/9f9a3e42-2bff-11e6-88a9-edef73de919e.png)

Issue #162, opened 2016-06-06
Rule colors mismatched on shows tab list


katarinabatina:
![image](https://cloud.githubusercontent.com/assets/6482926/15835250/15a224c2-2bfd-11e6-91e8-2904bafeb4d0.png)

PR #171, opened 2016-06-07
[Spacing] Update spacing across About and Shows tab

mennenia:
Fix: #167, #163, #164 , #162 

screen shot 2016-06-07 at 14 18 22

screen shot 2016-06-07 at 14 20 25

screen shot 2016-06-07 at 14 21 40


lib/components/artist/articles/index.js
```diff
@@ -29,7 +29,7 @@ class Articles extends React.Component {
 
 const styles = StyleSheet.create({
   heading: {
-    marginBottom: 20,
+    marginBottom: 12,
```
alloy: I really dislike having to resort to such calculations. Was there no way to get rid of the whatever else was gobbling up 8 pt on the other side?

lib/components/artist/biography.js
```diff
@@ -29,13 +29,13 @@ class Biography extends React.Component {
...
         { this.blurb(artist) }
-        <SerifText style={{ marginBottom: 40 }} numberOfLines={0}>{this.bioText()}</SerifText>
+        <SerifText style={{ marginBottom: 40, fontSize: 16, lineHeight: 20 }} numberOfLines={0}>{this.bioText()}</SerifText>
       </View>
     );
   }
 
   blurb(artist) {
-    return artist.blurb ? <SerifText style={{ marginBottom: 20, lineHeight: 25 }} numberOfLines={0}>{removeMarkdown(artist.blurb)}</SerifText> : null;
+    return artist.blurb ? <SerifText style={{ marginBottom: 20, fontSize: 16, lineHeight: 20 }} numberOfLines={0}>{removeMarkdown(artist.blurb)}</SerifText> : null;
```
alloy: These styles should go in a common `StyleSheet` object.

alloy:
Looking good!

My main issue is that I’d prefer to have styles that say `margin: 20` if we need a margin of 20, instead of some in this component and some elsewhere. Maybe this is a sign that that component shouldn’t define its own margins but it should be done from the parent instead?

lib/components/artist/articles/index.js
mennenia: I checked and it's usually one of two cases:

1) The 20px padding is applied several times, which is easy - you remove it in one case
2) It's just off, but not explicitly states elsewhere - it's the content of the list items that's further indenting itself (because they don't start at 0, the have their own margins also for in between the items).

lib/components/artist/biography.js
mennenia: Fixing. Also realised that I think there should be less space between the "Past Shows" header and the list, so will sort that too

lib/components/artist/articles/index.js
alloy: Gotcha. Maybe it’s something we can revisit some time and figure out how to get those other components to not gobble up margin.

katarinabatina:
Looking S H A R P!!!! Thanks for pixel pushing. This into tip top shape. Brian and I were discussing whether or not it would be easier for engineers to directly have access to our sketch files. Because sketch doesn't treat text objects like they are treated in web and iOS I think we set these values expecting one thing but typically get another. By giving direct access in sketch engineers may get a better sense for the desired spacing vs the specific value...just a thought we can continue discussion in slack.

mennenia:
Comments addressed and Past Shows updated (cc @katarinabatina):

simulator screen shot 7 jun 2016 14 54 25

lib/components/artist/biography.js
```diff
@@ -44,6 +44,19 @@ class Biography extends React.Component {
...
+    fontSize: 16, 
+    lineHeight: 20,
+    marginBottom: 20
+  }, 
+  bio: {
+    fontSize: 16, 
+    lineHeight: 20,
+    marginBottom: 40
+  }
+});
```
alloy: FYI (not a blocker) the common ones could have gone in a common style and then you’d pass an array of styles to the `style` prop, RN will then merge those. E.g.

```
  text: {
    fontSize: 16,
    lineHeight: 20,
  },
  blurb: {
    marginBottom: 20,
  }
});

<SerifText style={[styles.text, styles.blurb]} numberOfLines={0}>...</SerifText>
```

alloy:
Ace 👍

PR #173, opened 2016-06-07
[Artworks] Remove title space if no title available for artwork

### LINKED ISSUE #165: If no artwork title, eliminate extra space
currently we have blank space for where an artwork title should be but there are some artworks with no title in the DB, in this case we should eliminate the blank space

image

mennenia:
Fix #165 

screen shot 2016-06-07 at 16 27 18


sarahscott:
👏 👏 👏

PR #175, opened 2016-06-08
Update to React Native 0.27.1 and Relay 0.9.0

### LINKED ISSUE #140: [React Native] Upgrade to > 0.25.1
The packager complains about [this line](https://github.com/artsy/emission/blob/0b8d0a2c30860a778b80011ab6a7f5b881b4f7fb/Example/Emission/index.ios.js#L4), but _only_ when `dev=false`:

```
SyntaxError: 'import' and 'export' may appear only with 'sourceType: module'
```

* ~~A suggestion by @jsierles was that maybe that file still needs an export, presumably to enable the source type being a module during parsing in Babylon.~~ Tested this, this is not the issue.
* This issue and the linked issues seem like they could be related: https://github.com/facebook/react-native/issues/7850.

alloy:
Fixes #140.

The issue was rather simple in hindsight. It appears that RN by default defines a Babel config, but when you have a `.babelrc` file in the root of your project, it overrides that default one, which makes sense. The problem arose when building a release bundle due to me _only_ providing a config for the development environment, thus essentially removing all configuration from the config for the release env. The fix thus was to not scope the config to _just_ the development env: 94b43c1db41c375b2a1df6c7b39ae507b06eec98.


Emission.podspec
```diff
@@ -8,8 +8,10 @@ Pod::Spec.new do |s|
   s.summary      = 'React Native Components used by Eigen.'
   s.homepage     = 'https://github.com/artsy/emission'
   s.license      = 'MIT'
-  s.author       = { 'Eloy Durán' => 'eloy.de.enige@gmail.com' }
-  s.source       = { :git => "https://github.com/artsy/emission.git", :tag => s.version.to_s }
+  s.author       = { 'Eloy Durán' => 'eloy@artsy.net',
+                     'Maxim Cramer' => 'maxim@artsy.net',
+                     'Sarah Scott' => 'sarah.scott@artsy.net' }
+  s.source       = { :git => 'https://github.com/artsy/emission.git', :tag => s.version.to_s }
```
mennenia: #teamwork! 🎉

PR #176, opened 2016-06-08
Followers count now changes when follow button is pressed

### LINKED ISSUE #168: [Header] Increase/decrease follow count when tapping follow/unfollow button.
In this example, there was 1 follower before I followed the artist, so I would have expected to see it change to 2.

This does not necessarily need to refetch the data imo, although it could in the future with a Relay mutation. Right now just fake it.

image_0-8

sarahscott:
- [x] Follow:
follow

- [x] Unfollow

Fixes #168


mennenia:
Looks great!

PR #177, opened 2016-06-10
Fix crashes.

alloy:
Fixes https://github.com/artsy/eigen/issues/1616, https://github.com/artsy/eigen/issues/1667, and https://github.com/artsy/eigen/issues/1670. It _might_ fix https://github.com/artsy/eigen/issues/1619, but that’s just a guess atm.


alloy:
Self-merging.

PR #178, opened 2016-06-14
[package] Save source map for release bundles so we can symbolicate.

alloy:
This will make it easier to understand future crash reports.


mennenia:
🎉 yes please!

alloy:
For later reference:

* a sample script that symbolicates https://gist.github.com/javache/e10a098794c44d3de317
* question asking HockeyApp is they will add support https://support.hockeyapp.net/discussions/problems/58227-react-native-symbolication

PR #189, opened 2016-06-21
Fix for findNodeHandle crasher

sarahscott:
This is a temporary solution for a crash that we can't reproduce (#179). We should look into better ways to keep track of components going forward.

### DIFF
--- a/lib/modules/events.js
+++ b/lib/modules/events.js
@@ -4,7 +4,14 @@ import { findNodeHandle, NativeModules } from 'react-native';
 const { AREventsModule } = NativeModules;
 
 function postEvent(component, info) {
-  AREventsModule.postEvent(findNodeHandle(component), info);
+  let reactTag;
+  try {
+    reactTag = findNodeHandle(component);
+  } catch(err) {
+    return;
+  }
+
+  AREventsModule.postEvent(reactTag, info);
 }
 
 export default { postEvent };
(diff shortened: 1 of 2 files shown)


lib/modules/events.js
```diff
@@ -4,7 +4,11 @@ import { findNodeHandle, NativeModules } from 'react-native';
 const { AREventsModule } = NativeModules;
 
 function postEvent(component, info) {
-  AREventsModule.postEvent(findNodeHandle(component), info);
+  try {
+    AREventsModule.postEvent(findNodeHandle(component), info);
+  }
```
alloy: Ideally we'd only wrap the findNodeHandle call, store the resulting reactTag, and do the rest outside the try/catch block, so we don't mask other potential problems.

alloy: So in that case you should return from the catch clause.

lib/modules/events.js
```diff
@@ -4,7 +4,14 @@ import { findNodeHandle, NativeModules } from 'react-native';
 const { AREventsModule } = NativeModules;
 
 function postEvent(component, info) {
-  AREventsModule.postEvent(findNodeHandle(component), info);
+  var reactTag;
```
sarahscott: This feels a little clumsy but without it, the AREventModule breaks because `reactTag` is undefined. Alternatively I could put `postEvent` inside the `try` block.

alloy: Use 'let' instead of 'var'.

alloy: But otherwise this pattern is good 👍

sarahscott:
all set!

alloy:
Ace

PR #193, opened 2016-07-06
Relay Node support.

### LINKED ISSUE #181: Fix RCTFatalException in RCTFatal, line 133
@artsyit commented on [Sun Jun 05 2016](https://github.com/artsy/eigen/issues/1615)

### Stacktrace ###

<pre>-;RCTFatal;RCTAssert.m;133
RCTExceptionsManager;reportFatalException:stack:exceptionId:;RCTExceptionsManager.m;64
RCTModuleMethod;invokeWithBridge:module:arguments:;RCTModuleMethod.m;489
RCTBatchedBridge;_handleRequestNumber:moduleID:methodID:params:;RCTBatchedBridge.m;968
RCTBatchedBridge;handleBuffer:;RCTBatchedBridge.m;896</pre>

### Reason ###

*** Terminating app due to uncaught exception 'RCTFatalException: Unhandled JS Exception: undefined is not an object (evaluating 'r.image.aspect_ratio')', reason: 'Unhandled JS Exception: undefined is not an object (evaluating 'r.image.asp..., stack:
i@617:2044
value@617:2186
value@617:3125
_renderValidatedComponentWithoutOwnerOrContext@293:5069
_renderValidatedComponent@293:5197
_updateRenderedComponent@293:4592
_performComponentUpdate@293:4392
updateComponent@293:3669
performUpdateIfNecessary@293:3190
performUpdateIfNecessary@273:665
r@271:570
perform@276:494
perform@276:494
perform@271:1534
P@271:1706
closeAll@276:1131
perform@276:581
batchedUpdates@270:453
s@271:760
a@290:69
enqueueSetState@290:915
setState@78:317
value@617:1678
a@258:71
o@256:391
a@256:582
f@254:159
g@254:262
i@260:102
processEventQueue@254:1382
s@304:92
handleTopLevel@304:181
_receiveRootNodeIDEvent@303:613
receiveEvent@303:714
value@198:3100
<unknown>@198:1319
k@198:413
value@198:1291
'

### Link to HockeyApp ###

* [https://rink.hockeyapp.net/manage/apps/37029/crash_reasons/123909040](https://rink.hockeyapp.net/manage/apps/37029/crash_reasons/123909040)

---

@orta commented on [Sun Jun 05 2016](https://github.com/artsy/eigen/issues/1615#issuecomment-223826952)

From the looks of the crashes, it's these artists so far

```
[2016-06-05 12:31:15.685] [artist view] {
    "artist_id" = "wang-wusheng";
}
```

```
[2016-06-05 11:01:38.417] [artist view] {
    "artist_id" = "annie-leibovitz";
}
```

```
[2016-06-05 13:22:46.118] [artist view] {
    "artist_id" = "jennie-jieun-lee";
}
```

---

@alloy commented on [Fri Jun 10 2016](https://github.com/artsy/eigen/issues/1615#issuecomment-225292357)

Ok, what I _think_ happens is that this is because the artwork slug is the same as the artist slug:

* Relay always queries an object’s `id` if the queried type has that field. In the case of the artist that returns the artist’s slug, e.g. `annie-leibovitz`.
* Relay assumes that the `id` field is part of the `Node` instance of the Relay Global Object Identification Specification, i.e. the `id` value is unique amongst any entity (e.g. artist/artwork).
* Relay caches the artist result under the `id` key.
* Relay notices a query for an artwork object with an `id` of `annie-leibovitz` and assumes it can return the cached data, which is in reality an artist object, not an artwork.
* Everything goes downhill from here.

So it seems that I’ll have to get working on making metaphysics more Relay compatible earlier than anticipated.

/cc @dzucconi

---

@alloy commented on [Mon Jun 13 2016](https://github.com/artsy/eigen/issues/1615#issuecomment-225551124)

This patch at least fixes the crash, but it just means we’re omitting these artworks, which is clearly bad. https://github.com/artsy/emission/pull/177/commits/72d57e8d91e99177ebf30b25d02d91213e9ce3e2

alloy:
Fixes #181

This finally makes it so that all objects are uniquely identified, regardless of type, and that Relay uses `__id` for that purpose instead of `id`.

### DIFF
--- a/lib/components/artist/articles/index.js
+++ b/lib/components/artist/articles/index.js
@@ -20,7 +20,7 @@ class Articles extends React.Component {
                     showsHorizontalScrollIndicator={false}
                     scrollsToTop={false}
                     style={{ overflow: 'visible', marginBottom:40 }}>
-          { articles.map(article => <Article key={article.id} article={article} style={styles.article} />) }
+          { articles.map(article => <Article key={article.__id} article={article} style={styles.article} />) }
         </ScrollView>
       </View>
     );
@@ -42,7 +42,7 @@ export default Relay.createContainer(Articles, {
   fragments: {
     articles: () => Relay.QL`
       fragment on Article @relay(plural: true) {
-          id
+          __id
           ${Article.getFragment('article')}
       }
     `,
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -4,8 +4,9 @@ type AggregationCount {
   count: Int
 }
 
-type Article {
+type Article implements Node {
   cached: Int
+  __id: ID!
   id: String
   title: String
   published_at(format: String): String
@@ -23,8 +24,9 @@ enum ArticleSorts {
   PUBLISHED_AT_DESC
 }
 
-type Artist {
+type Artist implements Node {
   cached: Int
+  __id: ID!
   _id: String
   id: String
   href: String
@@ -89,8 +91,9 @@ type ArtistCounts {
   articles(format: String, label: String): FormattedNumber
 }
 
-type ArtistItem {
+type ArtistItem implements Node {
   cached: Int
+  __id: ID!
   _id: String
   id: String
   href: String
@@ -137,8 +140,9 @@ type ArtistMeta {
   description: String
 }
 
-type ArtistSearchEntity {
+type ArtistSearchEntity implements Node {
   cached: Int
+  __id: ID!
   _id: String
   id: String
   href: String
@@ -197,8 +201,9 @@ type ArtistStatuses {
   biography: Boolean
 }
 
-type Artwork {
+type Artwork implements Node {
   cached: Int
+  __id: ID!
   id: String
   _id: String
   to_s: String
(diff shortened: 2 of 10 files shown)


lib/components/artist/articles/index.js
```diff
@@ -42,7 +42,7 @@ export default Relay.createContainer(Articles, {
   fragments: {
     articles: () => Relay.QL`
       fragment on Article @relay(plural: true) {
-          id
+          __id
```
alloy: From now on we should always use `__id` for these purposes, because:

* Relay will implicitly query for this field, if it exists, anyways, so might as well save some bandwidth.
* This ID is guaranteed to be unique, whereas `id` might exist multiple times across types.

sarahscott:
👍

lib/components/artist/articles/index.js
orta: Might want putting in emission docs somewhere?

alloy: @sarahscott Want to pick that up?

PR #194, opened 2016-07-06
[Relay] Add network request debugger in dev env.

alloy:
screen shot 2016-07-06 at 12 32 49


alloy:
With this change, it should be possible to switch to the limited chrome debugger that you can use in VSCode. The only reason to switch to Chrome.app for network debugging would be to check headers, data size.

sarahscott:
👏 fancy

PR #196, opened 2016-07-06
Upgraded React Native version and fixed Related Artists layout issue

sarahscott:
- [x] Pretty straightforward RN upgrade to v0.29.0
- [x] Fixed known issue in the new RN version related to `flex-wrap` for related artists
- [x] Rigorously searched for other weirdness in new RN version

simulator screen shot jul 6 2016 6 12 00 pm

With the addition of `alignItems: flex-start`, everything is beautiful again:

simulator screen shot jul 6 2016 6 09 07 pm

Just like they said: https://github.com/facebook/react-native/releases/tag/v0.28.0

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,67 +1,20 @@
 [ignore]
 
-.*Example.*
-
 # We fork some components by platform.
-.*/*.web.js
 .*/*.android.js
 
-# Some modules have their own node_modules with overlap
-.*/node_modules/node-haste/.*
-
-# Ugh
-.*/node_modules/babel.*
-.*/node_modules/babylon.*
-.*/node_modules/invariant.*
-
-# Ignore react and fbjs where there are overlaps, but don't ignore
-# anything that react-native relies on
-.*/node_modules/fbjs/lib/Map.js
-.*/node_modules/fbjs/lib/fetch.js
-.*/node_modules/fbjs/lib/ExecutionEnvironment.js
-.*/node_modules/fbjs/lib/ErrorUtils.js
-
-# Flow has a built-in definition for the 'react' module which we prefer to use
-# over the currently-untyped source
-.*/node_modules/react/react.js
-.*/node_modules/react/lib/React.js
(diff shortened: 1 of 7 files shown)


alloy:
Have you tested that layout issue fix on iPad and also with artists where the related artists don’t fully fill the grid and that the remainder is left aligned?

As in, those grid fixes we had to make before when we started testing on iPad.

alloy:
Just found this ticket https://github.com/facebook/react-native/issues/8607. So we should double check if any scrollviews go blank for us.

alloy:
Also need to verify that our nested scrollviews monkey-patches still work as expected. i.e. if the artworks grid is still fetching works when you reach the end of the list.

sarahscott:
This still has the issue where show titles aren't truncating, but otherwise looks good. The grids re-fetch and align properly. I think it's important to get this merged so our subsequent Home view work doesn't run into layout issues later, so I'm making a separate issue for the text truncation.

alloy:
I’ve _suggested_ that [Libraries.io](https://libraries.io/npm/options) lists the license of the `options` package as MIT.

We can’t do much about those deprecated packages, so this is good on my part.

lib/relay/config.js
```diff
@@ -20,11 +18,10 @@ let metaphysicsURL;
...
 Relay.injectNetworkLayer(
-  new Relay.DefaultNetworkLayer(metaphysicsURL, {
-    headers: {
-      'X-ACCESS-TOKEN': Emission.authenticationToken,
-    }
-  })
+  new Relay.DefaultNetworkLayer(metaphysicsURL)
+    // headers: {
+    //   Authorization: 'Basic SECRET'
+    // })
```
alloy: This should not be included.

alloy: Might need a rebase?

sarahscott: Which lines in particular should not be included?

alloy: The auth headers should not be removed.

PR #201, opened 2016-07-11
[Shows] Fixed bug where current and upcoming shows were not properly wrapping

sarahscott:
Fixes this problem:

simulator screen shot jul 11 2016 3 42 52 pm

So our Current & Upcoming list looks like this instead:

simulator screen shot jul 11 2016 3 44 19 pm

### DIFF
--- a/lib/components/artist/shows/index.js
+++ b/lib/components/artist/shows/index.js
@@ -44,11 +44,11 @@ class Shows extends React.Component {
 
   currentAndUpcomingList() {
     if (this.props.artist.current_shows.length || this.props.artist.upcoming_shows.length) {
+      const shows = [].concat.apply([], [this.props.artist.current_shows, this.props.artist.upcoming_shows]);
       return (
         <View style={{ marginBottom: 20 }}>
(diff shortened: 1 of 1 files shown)


lib/components/artist/shows/index.js
```diff
@@ -44,11 +44,11 @@ class Shows extends React.Component {
 
   currentAndUpcomingList() {
     if (this.props.artist.current_shows.length || this.props.artist.upcoming_shows.length) {
+      const shows = [].concat.apply([], [this.props.artist.current_shows, this.props.artist.upcoming_shows]);
```
sarahscott: If they're in separate lists, the wrapping is off and the view will start a new row for upcoming shows. This way, we have one container for all Current & Upcoming shows.

sarahscott:
Do I need to worry about the dependency check?

alloy:
Nope, I already checked them https://github.com/dependencyci/support/issues/1

mennenia:
Awesome! 🎉

PR #202, opened 2016-07-11
[Home] Defer rails content fetching.

alloy:
home-deferred-fetching


lib/containers/home.js
```diff
@@ -5,22 +5,62 @@ import Relay from 'react-relay';
...
+    fetchContent: false,
+  },
+
   fragments: {
     home: () => Relay.QL`
       fragment on HomePage {
         modules {
+          key
           title
+          results @include(if: $fetchContent) {
```
sarahscott: fancy

PR #204, opened 2016-07-12
Unique generic gene rails ids

alloy:
React needs unique view keys to be able to identify views later on.

Also pulls in the recent metaphysics schema additions you both made.


mennenia:
Good stuff!

Issue #205, opened 2016-07-13
Fix duplicate images for shows and related artists


sarahscott:
This started happening after the RN upgrade:

![simulator screen shot jul 13 2016 12 31 14 pm](https://cloud.githubusercontent.com/assets/2712962/16800471/c6aac53e-48f5-11e6-948d-0bf8a5ddcc7d.png)

alloy:
This still needs a schema update.

PR #208, opened 2016-07-13
[schema] sync

alloy:
Fixes https://github.com/artsy/emission/issues/205


alloy:
Self-merging this, it works, no real work done, just syncing.

PR #211, opened 2016-07-19
Made a generic artworks grid for home rail

sarahscott:
This grid is exactly like the existing infinite scroll grid without the scroll-to-reload behavior. I've also moved the `artist/artworks` components to a new `artwork_grids` directory so we can use them in both home and artist contexts.

simulator screen shot jul 19 2016 1 35 58 pm

### DIFF
--- a/lib/components/home/artwork_rail.js
+++ b/lib/components/home/artwork_rail.js
@@ -3,9 +3,11 @@
 
 import Relay from 'react-relay';
 import React from 'react';
-import { View, Text } from 'react-native';
+import { View, Text, StyleSheet } from 'react-native';
 
 import Spinner from '../spinner';
+import Grid from '../artwork_grids/generic_grid';
+import SerifText from '../text/serif';
 
 class ArtworkRail extends React.Component {
   componentDidMount() {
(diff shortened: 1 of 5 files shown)


alloy:
Doesn’t need to happen now, but I think the duplication can be DRYed up a bit by having the infinite scroll grid wrap a generic one and add the scroll listener and next page fetching.

lib/components/home/artwork_rail.js
```diff
@@ -50,6 +58,7 @@ export default Relay.createContainer(ArtworkRail, {
         results @include(if: $fetchContent) {
           __id
           title
```
alloy: You no longer need the artwork `__id` and `title` here in this component, right?

sarahscott: Well I do need the rail title. It could be moved into `generic_grid`, but I think it makes sense to leave it in the rail since the title styling is specific to Home.

alloy: Oh I meant just for the artworks, I didn’t even realise that the `__id` for the rails could also be removed :)

alloy:
Looks good 👍

PR #213, opened 2016-07-19
[Relay] Use staging metaphysics in dev mode.

alloy:
Now that I know how we can ensure the required show data for the artist view is available on staging, let’s go back to using staging by default during development.

The issue was that ElasticSearch is not synced/re-indexed on staging after a DB copy. For now I know how to kick-off a job manually, so let me know if/when you need this.


sarahscott:
👍

PR #215, opened 2016-07-19
Support React-Native storybooks

orta:
End result:

This adds [React Native Storybooks](https://github.com/kadirahq/react-native-storybook) to Emission, 
as well as some simple refactoring to the obj-c side of the app. To some extent, you may be able to replace the tableview completely with this, but well, that's not my call to make. 

What this gives is the ability to control the state from outside the app, so you end up with a web server running this:

screen shot 2016-07-19 at 5 34 40 pm

This makes it easy to test multiple states. You create collections of stories based on a topic:

```js
storiesOf('Artist')
  .add('Glenn Brown', () => {
    const artistRoute = new Routes.Artist({ artistID: "glenn-brown" });
    return <RootContainer Component={Artist} route={artistRoute} />
  })
  .add('Leda Catunda', () => {
    const artistRoute = new Routes.Artist({ artistID: "leda-catunda" });
    return <RootContainer Component={Artist} route={artistRoute} />
  })
```

---

I've added some examples that wrap the Artist component, it needs to be wrapped in a [relay `RootContainer`](https://facebook.github.io/relay/docs/guides-root-container.html) but it seems to work nicely.


Example/Emission/ARStorybookComponentViewController.m
```diff
@@ -0,0 +1,18 @@
...
+}
+
+- (instancetype)initWithEmission:(AREmission *)emission;
+{
+    if ((self = [super initWithEmission:emission moduleName:@"Storybook" initialProperties:nil])) {
+    }
+    return self;
+}
+
+
```
orta: figure all this stuff is old hat to you lot

orta: notable I guess is that it's in the example app, instead of inside the pod

Example/Emission/AppDelegate.m
```diff
@@ -131,16 +133,43 @@ - (void)dismissModalViewController;
...
+           }
+        },
+        @{
+          @"name" : @"Artist",
+          @"router" : ^() {
+            return [[ARArtistComponentViewController alloc] initWithArtistID:ARTIST];
+          }
+        },
+      ];
+    }
```
orta: the if/else wasn't going to scale to three, figured this was the simplest, and allows for some dynamism

lib/app_registry.js
```diff
@@ -44,4 +44,15 @@ class Home extends React.Component {
 }
 
 AppRegistry.registerComponent('Artist', () => Artist);
-AppRegistry.registerComponent('Home', () => Home);
\ No newline at end of file
+AppRegistry.registerComponent('Home', () => Home);
+
+// How can I make this only happen in debug mode?
```
orta: ^

package.json
```diff
@@ -4,17 +4,18 @@
...
+    "sync-colors": "cd externals/elan && git fetch && git checkout origin/master && cp components/lib/variables/colors.json ../../data",
     "sync-externals": "npm run-script sync-schema && npm run-script sync-colors",
     "sync-schema": "cd externals/metaphysics && git fetch && git checkout origin/master && npm install && npm run dump-schema -- ../../data && rm -rf node_modules",
-    "sync-colors": "cd externals/elan && git fetch && git checkout origin/master && cp components/lib/variables/colors.json ../../data",
-    "start-packager": "cd node_modules/react-native && npm start -- --reset-cache",
-    "clean-example": "cd Example && xcodebuild -workspace Emission.xcworkspace -scheme Emission -destination 'platform=iOS Simulator,name=iPhone 6' clean",
-    "start": "npm run clean-example && npm run start-packager",
-    "bundle": "react-native bundle --platform=ios --dev=false --entry-file=index.ios.js --bundle-output Pod/Assets/Emission.js --sourcemap-output Pod/Assets/Emission.js.map"
+    "test": "env NODE_ENV=test NODE_PATH=./lib mocha --recursive ./test"
   },
```
orta: This is now ordered by alphabet, I added `storybook": "storybook start -p 9001`

Example/Emission/ARStorybookComponentViewController.m
alloy: > notable I guess is that it's in the example app, instead of inside the pod

Aye 👍

lib/app_registry.js
alloy:
```js
if (__DEV__) {
  // ...
}
```

alloy: Not sure if that will work with `import`, though, you might need to switch to using `require`.

alloy:
Forgive me if I’m wrong, but isn’t the point more that you would test/develop smaller components with storybook, rather than a full view like we already do?

alloy:
I’m thinking that it makes more sense to have the root table view from where you can select any of the full views _or_ the storybook, which would should the individual small components. What do you think?

alloy:
Might be nice to move _all_ of the storybook related code into a `Example/index.js` file and import the Emission module there.

Also means you will have to change the path of the local dev server in the app delegate to load the new index file.

package.json
alloy: OCDmazing!

orta:
> Forgive me if I’m wrong, but isn’t the point more that you would test/develop smaller components with storybook, rather than a full view like we already do?

Yep, I was just trying to get _something_ working, had to learn a bunch. 

The thing I'm not sure on is how to send stubbed data in to the relay context so it can pass to it's child components. Right now that is all happening dynamically in  Relay by making the networking requests.

alloy:
You should be able to just pass props to the individual components. Relay will probably show a warning about that, but we can deal with that afterwards.

orta:
Sure, will give it a shot

orta:
> I’m thinking that it makes more sense to have the root table view from where you can select any of the full views or the storybook, which would should the individual small components. What do you think?

This is current behavior:

screen shot 2016-07-20 at 1 03 02 pm

I agree, but it should be worth re-thinking about in the future

Example/Emission/index.ios.js
```diff
@@ -2,5 +2,4 @@
 'use strict';
 
 import 'emission';
-
-// TODO This file can also go, we should just host the source from the root of the repo.
\ No newline at end of file
+require('./storybook');
```
alloy: You should be able to use `import` here.

alloy:
> I agree, but it should be worth re-thinking about in the future

What I was mainly thinking about is that you don’t really want navigation to work in the storybook context, but you do want that to work in the context of testing full views.

Now you can check the VC type [here](https://github.com/artsy/emission/blob/2fec083986a615deb040598da3872eb0f0570914/Example/Emission/AppDelegate.m#L87) and not actually route if it comes from the storybook VC. Although you may want an alert with the route/params?

orta:
> What I was mainly thinking about is that you don’t really want navigation to work in the storybook context, but you do want that to work in the context of testing full views.

Is this for understanding what happens on taps? There's a logging tool in storybooks that could be used to show all that when interacting (or opening other stories too )

alloy:
Yup, the logging sounds good too 👍

orta:
Got it working with stubbed data

screen shot 2016-07-20 at 4 55 31 pm

stories/artist_header_story.js
```diff
@@ -0,0 +1,46 @@
...
+  .add('Real Artist - Glenn Brown', () => {
+    const artistRoute = new Routes.Artist({ artistID: 'glenn-brown' });
+    return <RootContainer Component={ArtistHeader} route={artistRoute} />;
+  })
+  .add('Real Artist - Leda Catunda', () => {
+    const artistRoute = new Routes.Artist({ artistID: 'leda-catunda' });
+    return <RootContainer Component={ArtistHeader} route={artistRoute}/>;
+  })
+  .add('No Brithday', () => {
+    let data = {
```
alloy: s/let/const/

Example/Emission/stub_container.js
```diff
@@ -0,0 +1,40 @@
...
+  getFragment() {}
+  hasFragment() {}
+  hasVariable() {}
+}
+
+// Expose dummy relay and a fake route
+StubbedContainer.childContextTypes = {
+  relay: React.PropTypes.object,
+  route: React.PropTypes.object
+};
```
alloy: Very cool :+1:

orta:
OK, this is good to go!

orta:
@sarahscott @mennenia - this updates a lot of the documentation about how to run Emission. If you want to run entirely inside vscode, you can now: https://github.com/artsy/emission/blob/refactor_routes/docs/vscode.md

Example/Emission/stub_container.js
```diff
@@ -17,13 +17,29 @@ export default class StubbedContainer extends React.Component {
...
     };
   }
 
+  // TODO: Improve handling state, currently this works only when you full re-render the whole virtual view tree
+  // e.g. jumping between two stories of the same component will not trigger state changes. I gave a
+  // considerable amount of time, then noticed we _don't_ use too much state inside our components
+  // and have made the time tradeoff to say it's not worth it for now.
+
+  // If you are using state in a component, where you would normally set the default, you should check props first
+  // e.g. `this.state = this.props.state || { following: false, followersCount: props.artist.counts.follows };`
```
alloy: I’m not following why this is, but `this.props.state` looks really counter-intuitive. Would love some clarification on the reason and ramifications first.

docs/vscode.md
```diff
@@ -0,0 +1,63 @@
...
+
+* To jump between splits, use `cmd + 1` and `cmd + 2`.
+* To navigate left/right through all tabs, use `cmd + alt + left` and `cmd + alt + right`. I remap these to be the same as most macOS apps: `ctrl + shit + [` and `ctrl + shit + ]`.
+  ```json
+    { "key": "cmd+shift+[",       "command": "workbench.action.previousEditor" },
+    { "key": "cmd+shift+]",       "command": "workbench.action.nextEditor" },
+
+  ```
+* To jump between open tabs in tab group (each side of a split for example) a use `ctrl + tab`.
+* `cmd + shift + y` is also the "Show Debug Console" as it is in Xcode.
```
alloy: This doc is great, thanks! 👏

lib/components/artist/biography.js
```diff
@@ -3,7 +3,7 @@
 
 import Relay from 'react-relay';
 import React from 'react';
-import { View, Text, StyleSheet, Dimensions } from 'react-native';
```
alloy: Awww yeah, eslint complaining?

lib/components/artist/header.js
```diff
@@ -28,7 +28,7 @@ class Header extends React.Component {
   constructor(props) {
     super(props);
     this.handleFollowChange = this.handleFollowChange.bind(this);
-    this.state = { following: false, followersCount: props.artist.counts.follows };
+    this.state = this.props.state || { following: false, followersCount: props.artist.counts.follows };
```
alloy: I really don’t like this.

lib/components/artist/biography.js
orta: Yeah, built into VSCode

alloy: Is this the future?

Example/Emission/stub_container.js
orta: The problem is that somehow we need to be able to pass state into a component that is _at minimum_ 2 levels down
`<StubbedContainer><RelayContainer><The Thing You Are Writing Your Story About>` - I explored a bunch of the API with 3/4 web devs over the course of the day, but eventually came to the conclusion that the only way to send any state for component being worked on, is via the props system.

I'm happy to call it something else, or say that you can't define `state` in a story, the thing was that when I reviewed more of the components we had, very few of them used and mutated the components `state`. It just happened to be the one I randomly picked first. So, it didn't feel worth trying to create something fully systematized.

The tradeoff is that you can only control _some_ of a component, in the examples in the header, it's trivial to change the artist name, birthday etc, but to make changes to the "following" state would be harder, which is why I added the default values via props.

orta: 👍 works for me boss-man

lib/components/artist/biography.js
orta: hahaha

orta:
OK updated, and remove the state API

alloy:
Mental, merging hard :metal:

PR #220, opened 2016-07-22
Convert to Flow

orta:
### This does two major things:

- [x] Passes ESLint
- [x] Converts to using Flow

### and one semi-major thing.

- [x] Converts the app to use class properties.

So this would take these _realllllllly weird_ bits of a constructor: `this.func_name = this.func_name.bind(this)` and turns them into normal functions using class instance fields. This comes in with React Native in the form of [class-properties-transform](http://babeljs.io/docs/plugins/transform-class-properties/). It means that you replace it with

```js
func_name = () => {
  ...
}
```
whereas it used to be `func_name(){ ... }`. The `=>` will rebind the `this` for you. Seems to work, at least a bunch of functions (like tabbing) seems to have not broken. [Blog post on the topic.](http://www.ian-thomas.net/autobinding-react-and-es6-classes/)

### TODO: 

- [x] Decide on what to do with ensuring RN / R are in the flow/typescript types database. 

The decision here is that we would have to put a 5k file in that needs updating whenever react/react-native update.

---

As ever, I've been documenting all this stuff in the `vscode.md` file. You can read the version while this PR is open here: https://github.com/artsy/emission/blob/flow/docs/vscode.md


Example/Emission/AppDelegate.m
```diff
@@ -88,12 +90,20 @@ - (void)setupEmission;
...
     [fromViewController.navigationController pushViewController:[self viewControllerForRoute:route]
                                                        animated:YES];
   };
 
   emission.switchBoardModule.presentModalViewController = ^(UIViewController * _Nonnull fromViewController,
                                                             NSString * _Nonnull route) {
+    if (self.isShowingStorybook) {
+      NSLog(@"Route modal - %@", route);
+      return;
+    }
```
orta: I experimented with trying to pass a `console.log` back to JS world, but it just wasn't working very well.

lib/components/opaque_image_view.js
```diff
@@ -16,10 +16,18 @@ const GeminiHost = 'd7hftxdivxxvm.cloudfront.net';
 const ImageQuality = 85;
 
 export default class OpaqueImageView extends React.Component {
+  state: {
+    aspectRatio: ?number,
+    width?: number | string,
+    height?: number | string
+  };
```
orta: to unpack this, if it's the first time you've seen this stuff.

aspectRatio is an optional number,
width is an optional key, that can be a number or a string 
height is an optional key, that can be a number or a string 

the number/string stuff is weird, because we want to pass width and height back inside a string. Maybe there's a better way to say print this number as a string that I've not found in flow.

lib/components/opaque_image_view.js
```diff
@@ -16,10 +16,18 @@ const GeminiHost = 'd7hftxdivxxvm.cloudfront.net';
...
+
+  DefaultProps: {
+    placeholderBackgroundColor: string
+  }
+
   constructor(props) {
     super(props);
-    this.state = {};
```
orta: this gets set on line 26 or 34

Example/Emission/AppDelegate.m
alloy: Just wondering, was `fromViewController` not a `ARStorybookComponentViewController` as expected?

Pod/Classes/Core/AREmission.m
```diff
@@ -79,4 +79,13 @@ - (NSURL *)releaseBundleURL;
   return [[NSBundle bundleForClass:self.class] URLForResource:@"Emission" withExtension:@"js"];
 }
 
+- (void)log:(NSString *)format, ... {
+    va_list args;
+    va_start(args, format);
+    NSString *message = [NSString stringWithFormat:format, args];
+    [self.bridge enqueueJSCall:@"console.log" args:@[message]];
+    va_end(args);
+}
```
alloy: So will this be removed, if it wasn’t working well?

lib/components/artist/about.js
```diff
@@ -27,7 +28,7 @@ class About extends React.Component {
           <Biography artist={this.props.artist} />
           <Separator style={styles.sectionSeparator} />
         </View>
-      )
+      );
```
alloy: I thought we were getting rid of semi-colons?

Pod/Classes/Core/AREmission.m
orta: yeah

lib/components/artist/about.js
orta: Next PR, this PR would be totally unreadable otherwise

Example/Emission/AppDelegate.m
orta: It was, but params could get put in the route

alloy: I’m not sure I understand why this wouldn’t work?

```objc
if ([fromViewController isKindOf:ARStorybookComponentViewController.class]) {
  NSLog(@"Route modal - %@", route);
  return;
}
```

My main preference for this is that it serves the purpose of an ‘example’ app, i.e. shows/tests the API.

orta: Sure, it can change to that, no worries to me - less code 👍

lib/components/opaque_image_view.js
alloy: Why do we need the dimensions to be strings?

alloy:
I’m very excited to have this in! 👍 💯

lib/components/opaque_image_view.js
orta: For printing here:

```
       return `https://${GeminiHost}/?resize_to=${type}&width=${this.state.width}&height=${this.state.height}&quality=${ImageQuality}&src=${encodeURIComponent(imageURL)}`;

```

alloy: Orly? I see.

How about we just change that to the below, should make the type info much clearer as it’s close to where it’s needed.

```js
const width = this.state.width.toString();
const height = this.state.height.toString();
return `https://${GeminiHost}/?resize_to=${type}&width=${width}&height=${height}&quality=${ImageQuality}&src=${encodeURIComponent(imageURL)}`;
```

orta: Yep 👍

orta:
Feedback addressed, gonna see what happens WRT the typings - should make it automatically work for everyone rather than requiring some extra work

PR #221, opened 2016-07-25
[HomeArtworkRail] Rail layout for initial state

sarahscott:
This is a preliminary go at a nice way to do the initial state. I will follow up soon with a final state and my first go at animating between the two.

expansionanimation

### DIFF
--- a/lib/components/home/artwork_rail.js
+++ b/lib/components/home/artwork_rail.js
@@ -8,6 +8,7 @@ import { View, StyleSheet } from 'react-native';
 import Spinner from '../spinner';
 import Grid from '../artwork_grids/generic_grid';
 import SerifText from '../text/serif';
+import Separator from '../separator';
 
 class ArtworkRail extends React.Component {
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


lib/components/home/artwork_rail.js
```diff
@@ -3,22 +3,72 @@
...
+  }
+
   renderModuleResults() {
-    if (this.props.rail.results) {
+    if (this.props.rail.results && this.props.rail.results.length) {
       return (
-        <Grid artworks={this.props.rail.results}/>
+        <View style={[styles.container, { height: this.state.expanded ? 2350 : 500 }]}>
+          <View style={[styles.gridContainer, { height: this.state.expanded ? 2300 : 420 }]}>
+            <Grid artworks={this.props.rail.results}/>
```
sarahscott: If I add an `onLayout` prop here and bind it to some function in this rail, it is never called. If I add `onLayout` to any other view, it behaves as expected and is called upon layout. I'm not sure how to get the height of the grid when it's finished laying itself out, and that value would be useful in the above lines for setting the expanded height of the `container` and `gridContainer`.

orta:
dependency-ci  is down, we're seeing it on eigen too

lib/components/home/artwork_rail.js
alloy: What happens when you wrap the Grid in another wrapper and set the `onLayout` prop there?

lib/components/home/artwork_rail.js
```diff
@@ -3,22 +3,72 @@
...
+    return null;
+  }
+
+  onViewAllPress() {
+    // pop new view controller
+  }
+
   renderModuleResults() {
-    if (this.props.rail.results) {
+    if (this.props.rail.results && this.props.rail.results.length) {
```
alloy: I think we already pull lodash in, so we could start using [isEmpty](https://lodash.com/docs#isEmpty) for these things.

lib/components/home/artwork_rail.js
```diff
@@ -3,22 +3,72 @@
...
+            <Grid artworks={this.props.rail.results}/>
+            { this.renderViewAllButton() }
+          </View>
+          <Separator/>
+          { this.renderExpansionButton() }
+        </View>
       );
+    } else if (this.props.rail.results) {
+      // temporary; for those pesky empty rails on staging
+      return <View/>;
```
alloy: Which rails are those?

alloy:
w00t, looks amaze :fire:

PR #224, opened 2016-07-26
[HomeArtworkRail] Height of expanded rail will now correspond to height of loaded artworks grid

### LINKED ISSUE #222: [Home][ArtworksRail] Make height of expanded container correspond to height of loaded artworks grid

### LINKED ISSUE #223: [Home][ArtworksRail] Make expansion button into chevron image

sarahscott:
- [x] Figure out height issue

Fixes #222 

simulator screen shot jul 26 2016 10 16 13 pm


externals/metaphysics
```diff
@@ -1 +1 @@
-Subproject commit 849a8e2c1f7b7da61b06f152e5ea4ea34c769387
+Subproject commit 1541a0d42480bd95aee87898200a8ad88e086a45
```
sarahscott: `git submodule update --init` ¯\_(ツ)_/¯

sarahscott:
I should probably also make the chevron retina-ready but maybe we could do that in Extraction?

Example/Podfile
```diff
@@ -14,7 +14,7 @@ target 'Emission' do
     pod 'Artsy+UIFonts', :git => "https://github.com/artsy/Artsy-OSSUIFonts.git"
   end
 
-  pod 'React', :path => react_path, :subspecs => ['RCTWebSocket']
+  pod 'React', :path => react_path, :subspecs => ['RCTWebSocket', 'RCTImage']
```
sarahscott: RN's `Image` needs this.

alloy: I would prefer we don’t pull it in. Does our own image view not work because we need transparency?

lib/components/home/artwork_rail.js
```diff
@@ -54,12 +56,19 @@ class ArtworkRail extends React.Component {
...
+
   renderModuleResults() {
     if (this.props.rail.results && this.props.rail.results.length) {
       return (
-        <View style={[styles.container, { height: this.state.expanded ? 2350 : 500 }]}>
-          <View style={[styles.gridContainer, { height: this.state.expanded ? 2300 : 420 }]}>
-            <Grid artworks={this.props.rail.results}/>
+        <View style={[styles.container, { height: this.state.expanded ? this.state.gridHeight + 150 : 500 }]}>
+          <View style={[styles.gridContainer, { height: this.state.expanded ? this.state.gridHeight + 100 : 420 }]}>
+            <View onLayout={this.onGridLayout.bind(this)}>
```
sarahscott: This works perfectly; thanks for the suggestion @alloy!

Example/Podfile
sarahscott: We can use our own image view, but I kept getting this error with it: `Touchable child must either be native or forward setNativeProps to a native component` (https://facebook.github.io/react-native/docs/direct-manipulation.html), and when I tried wrapping it in a dummy view, the error went away but the resulting image was just a black square. I can try using the `setNativeProps` solution outline in the RN docs, but it seemed easier to just use the most basic `Image` component provided by RN.

Do you think using `setNativeProps` is a better way to go?

alloy: By using RN's Image now, we're pulling in a lot of code that we might need to maintain. This might be something we want to do in the future, but for just this image it seems a bit overkill.

I think you may actually be able to solve this issue by wrapping the image view in a wrapper view or otherwise forward the native props and normal props as outlined in that doc.

sarahscott: When I tried wrapping it, it was just a black square. I took a look at our `opaque_image_view` and it seems to assume the `imageURL` should always be changed into a Gemini-friendly URL. I'm a little hesitant to make a lot of changes there because it's used so often elsewhere in emission, but I can if necessary.

lib/components/home/artwork_rail.js
alloy: Awesome 👍

Example/Podfile
alloy: I see, yeah we should probably decouple that. I.e. Have a base image view and one that wraps it that works with Gemini URLs.

alloy:
Did you generate the chevron image yourself? :)

You can just ask Katarina for assets.

Example/Podfile
sarahscott: Cool. I'll start working on that. In the meantime, I'll make this pr mergeable by removing the chevron image for now.

sarahscott:
Yeah :). I was working at a weird hour when no one else was awake and threw that in there for testing. For the release, we'll definitely need retina-ready assets.

sarahscott:
This is now ready to merge as a PR for the height solution only. My next PR will deal with the image.

lib/components/home/artwork_rail.js
```diff
@@ -31,8 +33,8 @@ class ArtworkRail extends React.Component {
   renderExpansionButton() {
     if (!this.state.expanded) {
       return (
-        <TouchableHighlight onPress={this.onPress.bind(this) } underlayColor={'white'}>
-          <View style={styles.expansionButton}/>
+        <TouchableHighlight style={styles.expansionButton} onPress={this.onPress.bind(this) } underlayColor={'white'}>
+            <View style={{height: 10, width: 10, alignSelf: 'center'}} />
```
sarahscott: Leaving this here because `TouchableHighlight` requires one (and only one) child, but this view doesn't actually do anything at the moment.

lib/components/home/artwork_rail.js
```diff
@@ -54,12 +56,18 @@ class ArtworkRail extends React.Component {
...
+  }
+
   renderModuleResults() {
     if (this.props.rail.results && this.props.rail.results.length) {
       return (
-        <View style={[styles.container, { height: this.state.expanded ? 2350 : 500 }]}>
-          <View style={[styles.gridContainer, { height: this.state.expanded ? 2300 : 420 }]}>
-            <Grid artworks={this.props.rail.results}/>
+        <View style={[styles.container, { height: this.state.expanded ? this.state.gridHeight + 150 : 500 }]}>
+          <View style={[styles.gridContainer, { height: this.state.expanded ? this.state.gridHeight + 100 : 420 }]}>
```
sarahscott: The `gridContainer` contains the grid and also the View All button at the bottom when the rail is expanded.

lib/components/home/artwork_rail.js
sarahscott: @mennenia This view exists because setting an `onLayout` prop for the grid itself doesn't seem to work properly.

externals/metaphysics
orta: Changes: 
https://github.com/artsy/metaphysics/compare/849a8e2c1f7b7da61b06f152e5ea4ea34c769387...1541a0d42480bd95aee87898200a8ad88e086a45

I assume this got updated as a part of running `npm run sync-schema` - should be fine

lib/components/home/artwork_rail.js
orta: I wonder if this can be replaced with the ES7 syntax `this.onGridLayout.bind(this)` 

can you try this?

```
<View onLayout={onGridLayout}>
```
and above
```
 onGridLayout = (event) => {
      this.setState({ gridHeight: event.nativeEvent.layout.height });
  }
 
```

orta:
Aside from my ES7 stuff, looks good to 👍

lib/components/home/artwork_rail.js
sarahscott: Hmmm I get `onGridLayout is not defined` when I try that. Have you used that elsewhere with RN?

orta: [I have](https://github.com/artsy/emission/pull/220/files#diff-2f5aa3c37ef9f2653a0096c9f8344357L30), maybe the context is different though - it's early days for me, can you try `<View onLayout={this.onGridLayout}>`

If that doesn't work - merge away, and I'll note to check out the language semantics later

PR #225, opened 2016-07-27
[WIP-ish] Initial stab at android support

orta:
This adds a dropdown to VS Code to launch the app in Android.

screen shot 2016-07-27 at 4 56 38 pm

Which currently renders our home component as the root component, there would need to be work around replicating/replacing the ARSwitchView and ARSpinner for the other views.

screen shot 2016-07-27 at 4 05 18 pm

This doesn't render the artwork images, I don't think `onLayout` is being called for them. Figure it'll take a tiny amount of time once alloy is in town to fix with him - rather than me poring through source.

---

Setup docs: https://github.com/artsy/emission/blob/android/docs/android.md

As we don't have pods for fonts etc, I'm copying them from the Pods, which will be consistently named.


orta:
Wasn't planning on merging, would slow down ios dev for little reason ATM, but it's nice to see what this looks like.

alloy:
Yes, totally!

PR #226, opened 2016-07-29
[VSCode] Add vscode support files to gitignore

mennenia:


.vscode/launchReactNative.js
```diff
@@ -1,9 +0,0 @@
-// This file is automatically generated by vscode-react-native@0.1.5
-// Please do not modify it manually. All changes will be lost.
-try {
-    var path = require("path");
-    var Launcher = require("/Users/orta/.vscode/extensions/vsmobile.vscode-react-native-0.1.5/out/debugger/launcher.js").Launcher;
```
orta: this was specific to me, so now should be generated at runtime per-dev

orta:
Looks great 👍

PR #228, opened 2016-07-29
[WIP][Home][ArtworksRail] Non-performant WIP way to change grid sectionCount upon rotation

sarahscott:
I'm trying to figure out the best way to make the artworks grid update its number of columns when it rotates on iPad:

screen shot 2016-07-29 at 9 13 02 pm

This solution slows things down considerably and I don't like it, but I wanted to show where my train of thought is going. I'd like to minimize the number of times `Dimensions` is called, but also make sure that it is called at least once when a rotation happens.


sarahscott:
closing in favor of solution discussed briefly with @alloy in which `sectionCount` is passed in as a `prop`.

PR #229, opened 2016-08-01
[Home][ArtworkRail] Added chevron image to expansion button

sarahscott:
simulator screen shot aug 1 2016 3 39 07 pm


lib/components/home/artwork_rail.js
```diff
@@ -34,7 +34,7 @@ class ArtworkRail extends React.Component {
     if (!this.state.expanded) {
       return (
         <TouchableHighlight style={styles.expansionButton} onPress={this.onPress.bind(this) } underlayColor={'white'}>
-            <View style={{height: 10, width: 10, alignSelf: 'center'}} />
+            <Image style={{height: 8, width: 15, alignSelf: 'center'}} source={require('../../../images/chevron.png')} />
         </TouchableHighlight>
```
orta: so the pattern is that we use `Image` for local images, and `ImageView` for network'd ones?

sarahscott: For now, yes. `opaque_image_view` should be used for all artwork images and React's standard `Image` component for local ones.

orta:
Alright, reviewed, looks good to me 👍 - feel free to merge if you're happy

PR #231, opened 2016-08-12
[Home][ArtworkRail] Styling for iPhone & iPad

sarahscott:
Just some styling 

simulator screen shot aug 12 2016 5 19 31 pm

simulator screen shot aug 12 2016 5 16 20 pm


lib/components/home/artwork_rail.js
```diff
@@ -96,9 +101,12 @@ class ArtworkRail extends React.Component {
 
 const styles = StyleSheet.create({
   title: {
-    margin: 10,
-    fontSize: 20,
+    margin: isPad ? 40 : 20,
```
sarahscott: at some point we should think about whether it makes sense to make `ipad_title` and `iphone_title` styles instead of doing this little check a bunch of times.

orta: An option, could be to create our own subclass of StyleSheet ( or our own function, whatever)

```
  title: {
    margin: 20,
    alignSelf: 'center',
    textAlign: 'center',
  },
  container: {
    backgroundColor: 'white',
  }
}, largeDevice: {
  title: {
    margin: 40
  } 
})
```
 
which then overwrites the default with the large one on bigger devices

sarahscott: Cool idea

orta:
Cool - lets ship this for now. The stylesheet  ideas are not blockers 👍

PR #234, opened 2016-08-15
[Home][ArtworkRail] Expansion-related edge cases

sarahscott:
- [x] Grid will only render expansion button if there are more than 3 artworks in rail `results`
- [x] Polish

Some styling will change when I work on the View All button showing/hiding, but this is a good intermediary PR.

This is a screenshot of the rail when only 3 artworks are displayed:

simulator screen shot aug 15 2016 6 01 35 pm


lib/components/home/artwork_rail.js
```diff
@@ -19,6 +19,7 @@ class ArtworkRail extends React.Component {
 
     this.state = {
       expanded: false,
+      hasAdditionalContent: true,
```
sarahscott: This is useful for hiding/showing the View All button and will likely be removed in the future if we pass this value down with Relay.

orta:
Cool - makes sense! 👍

orta:
Yep

PR #241, opened 2016-08-16
[Danger] Initial scoping for Danger

orta:
JS Dangerfiles are completely new, so this might take some iteration.

### DIFF
--- a/Dangerfile.js
+++ b/Dangerfile.js
@@ -1,1 +1,6 @@
 message("Hello, world.")
+
+
+if (git.modified_files.include("CHANGELOG.md") === false) {
+  fail("No CHANGELOG added.");
+}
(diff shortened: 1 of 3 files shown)


Dangerfile.js
```diff
@@ -1 +1,23 @@
-message("Hello, world.")
+// CHANGELOG check
+if (_.includes(git.modified_files, "CHANGELOG.md") === false) {
+  fail("No CHANGELOG added.")
+}
+
+const test_files = _.filter(git.modified_files, function(path){
```
anandaroop: Nice! I would observe that in my rush to give you some good sample code I wasn't necessarily following good style. Should these `const` names all be camelcase instead of snakecase?

And is that the sort of thing Danger would complain about? 😉

orta: In enough time, I will create an eslint plugin for Danger, so that we'd get those messages inline

orta: for now, YOLO

cogell: > for now, YOLO

PR #252, opened 2016-08-24
[package] Add concurrently back and clean-up shrinkwrap.

alloy:
Ok, so, our shrinkwrap was out-of-date (still contained removed test packages) and we were missing the `concurrently` package.

Two of our dependencies (`react-native`storybook` and `react-test-renderer`) have version requirements on common dependencies that are unsatisfiable, which means that [npm will not shrinkwrap](https://github.com/npm/npm/issues/10836). However, they work fine at runtime with our current versions.

After going down the rabbit hole for a while, I settled on fixing it by hand-editing the `package.json` files of these two dependencies to bump the version requirements that were troublesome such that they would satisfy. After that I was able to shrinkwrap again, of which this is the result.

I suspect that there will be similar issues in the future when trying to shrinkwrap, because your `package.json` files will be re-installed with their original version requirements. We should then:

* either see if we can actually fix the issue by updating some dependencies
* re-edit the files to satisfy the requirements

### DIFF
--- a/Dangerfile.js
+++ b/Dangerfile.js
@@ -1,23 +1,27 @@
 // CHANGELOG check
-if (_.includes(git.modified_files, "CHANGELOG.md") === false) {
+const hasAppChanges = _.filter(git.modified_files, function(path){
+  return _.includes(path, 'lib/');
+}).length > 0
+
+if (hasAppChanges && _.includes(git.modified_files, "CHANGELOG.md") === false) {
   fail("No CHANGELOG added.")
 }
 
-const test_files = _.filter(git.modified_files, function(path){
+const testFiles = _.filter(git.modified_files, function(path){
   return _.includes(path, '__tests__/');
 })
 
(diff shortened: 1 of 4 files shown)


alloy:
Filed a ticket with one of the packages https://github.com/kadirahq/react-native-storybook/issues/54

ArtsyMobile:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

</p>

orta:
want me to ignore the dep-ci too?

alloy:
omg, firebase???!

Yeah the license issue can be ignored, I'm more interested who the hell is pulling that in.

orta:
Yeah, looks like it's react storybooks

PR #255, opened 2016-08-25
Improve code style

KrauseFx:
Me being a JS expert


orta:
This would be so much nicer with [ES6 arrow functions](http://tc39wiki.calculist.org/es6/arrow-functions/).

KrauseFx:
I broke it... not as much JS expert as I thought

orta:
Nah, we're not set up to run PRs from 3rd party, probably won't ever be - this looks good to me 👍

PR #265, opened 2016-08-30
[Home][ArtworksRail] Updated routing

sarahscott:
- [x] Use `key` instead of `context` for View All routing
- [x] Use force favorites route 
- [x] Determine if `active_bids` rail requires View All and if so, where it should route to
- [x] Remove post-expansion 'View All' button from rails without additional content

### DIFF
--- a/lib/components/home/artwork_rails/artwork_rail.js
+++ b/lib/components/home/artwork_rails/artwork_rail.js
@@ -15,13 +15,21 @@ import fragments from './relay_fragments'
 
 const isPad = Dimensions.get('window').width > 700
 
+const additionalContentRails = [
+  'followed_artists',
+  'saved_works',
+  'live_auctions',
+  'current_fairs',
+  'genes',
+  'generic_gene',
+]
+
 class ArtworkRail extends React.Component {
   constructor(props) {
     super(props)
 
     this.state = {
       expanded: false,
-      hasAdditionalContent: true,
       gridHeight: 0,
     }
   }
@@ -35,44 +43,30 @@ class ArtworkRail extends React.Component {
     this.setState({ expanded: true })
   }
 
-  renderExpansionButton() {
-    if (!this.state.expanded) {
-      return (
-        <TouchableHighlight style={styles.expansionButton} onPress={this.expand} underlayColor={'white'}>
-            <Image style={{height: 8, width: 15, alignSelf: 'center'}} source={require('../../../../images/chevron.png')} />
-        </TouchableHighlight>
-      )
-    }
-    return null
-  }
-
-  renderViewAllButton() {
-    if (this.state.expanded) {
-      return (
-        <TouchableHighlight style={styles.viewAllButton} onPress={this.handleViewAll} underlayColor={'gray'}>
-          <Text style={styles.viewAllText}>VIEW ALL</Text>
-        </TouchableHighlight>
-      )
-    }
-    return null
+  hasAdditionalContent() {
+    return additionalContentRails.indexOf(this.props.rail.key) > -1
   }
 
   handleViewAll = () => {
     const context = this.props.rail.context
-    if (!context) { return }
+    const key = this.props.rail.key
 
     let url = null
-    switch (context.__typename) {
-      case 'HomePageModuleContextFollowArtists':
-        url = 'me/notifications'
+    switch (key) {
+      case 'followed_artists':
+        url = '/me/notifications'
         break
-      case 'HomePageModuleContextRelatedArtist':
-        url = context.artist.href
+      case 'related_artists':
+        url = context && context.artist.href
         break
-      case 'HomePageModuleContextGene':
-      case 'HomePageModuleContextFair':
-      case 'HomePageModuleContextSale':
-        url = context.href
+      case 'saved_works':
+        url = '/user/saves#saved-artworks'
+        break
+      case 'genes':
+      case 'generic_gene':
+      case 'current_fairs':
+      case 'live_auctions':
+        url = context && context.href
         break
     }
 
@@ -89,11 +83,14 @@ class ArtworkRail extends React.Component {
   }
 
   mainContainerHeight() {
+    // includes height of View All button (if present), the separator, and a 30px margin
+    const supplementaryHeight = this.hasAdditionalContent() ? 101 : 30
+
     if (this.expandable()) {
       const initialRailHeight = isPad ? 460 : 530
-      return this.state.expanded ? this.state.gridHeight + 101 : initialRailHeight
+      return this.state.expanded ? this.state.gridHeight + supplementaryHeight : initialRailHeight
     } else {
-      return this.state.gridHeight + 31
+      return this.state.gridHeight + supplementaryHeight
     }
   }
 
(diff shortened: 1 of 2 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="true">The following files do not have tests: <a href="https://github.com/artsy/emission/blob/fe5ab6ca126bffc311d925b44f71b892912de1a4/lib/app_registry.js">lib/app_registry.js</a>, <a href="https://github.com/artsy/emission/blob/fe5ab6ca126bffc311d925b44f71b892912de1a4/lib/components/home/artwork_rails/artwork_rail_header.js">lib/components/home/artwork_rails/artwork_rail_header.js</a> &amp; <a href="https://github.com/artsy/emission/blob/fe5ab6ca126bffc311d925b44f71b892912de1a4/lib/components/load_failure_view.js">lib/components/load_failure_view.js</a></td>
    </tr>
  </tbody>
</table>

</p>

sarahscott:
ready for review!

lib/components/home/artwork_rails/artwork_rail.js
```diff
@@ -35,44 +34,38 @@ class ArtworkRail extends React.Component {
...
+        url = '/me/notifications'
+        break
+      case 'related_artists':
+        url = context && context.artist.href
         break
-      case 'HomePageModuleContextRelatedArtist':
-        url = context.artist.href
+      case 'saved_works':
+        url = '/user/saves#saved-artworks'
         break
```
alloy: Just for knowledge-sake, why do these not have a `context.href` like the ones below?

sarahscott: Because those don't have a `context`

lib/components/home/artwork_rails/artwork_rail.js
```diff
@@ -89,11 +82,14 @@ class ArtworkRail extends React.Component {
   }
 
   mainContainerHeight() {
+    // includes height of View All button (if present), the separator, and a 30px margin
+    const supplementaryHeight = this.hasAdditionalContent() ? 101 : 30
+
     if (this.expandable()) {
       const initialRailHeight = isPad ? 460 : 530
```
alloy: We should totally (down the road) move these to stylesheets for e.g. ‘regular’ and ‘compact’ size classes, like Apple does with `UIUserInterfaceSizeClass`.

lib/components/home/artwork_rails/artwork_rail.js
```diff
@@ -116,14 +112,39 @@ class ArtworkRail extends React.Component {
...
+        return (
+          <TouchableHighlight style={styles.viewAllButton} onPress={this.handleViewAll} underlayColor={'gray'}>
+            <Text style={styles.viewAllText}>VIEW ALL</Text>
+          </TouchableHighlight>
+        )
+      }
+
+      // otherwise, use a spacer view
+      return <View style={{ height: 30 }} />
+    }
```
alloy: The indentation does seem right above, which made it slightly confusing to read.

lib/components/home/artwork_rails/artwork_rail.js
```diff
@@ -191,6 +212,7 @@ export default Relay.createContainer(ArtworkRail, {
     rail: () => Relay.QL`
       fragment on HomePageArtworkModule {
         ${Header.getFragment('rail')}
+        key
         context {
           __typename
```
alloy: Do we still need the `__typename` ?

lib/components/home/artwork_rails/artwork_rail_header.js
```diff
@@ -56,23 +56,31 @@ class ArtworkRailHeader extends React.Component {
...
 
+  hasAdditionalContent() {
+    const additionalContentRails = [
+      'followed_artists',
+      'saved_works',
+      'live_auctions',
+      'current_fairs',
+      'genes',
+      'generic_gene',
+    ]
```
alloy: Is this the same as defined in the artwork rail?

lib/components/home/artwork_rails/artwork_rail_header.js
```diff
@@ -94,21 +102,25 @@ class ArtworkRailHeader extends React.Component {
...
-      case 'HomePageModuleContextGene':
-      case 'HomePageModuleContextFair':
-      case 'HomePageModuleContextSale':
-        url = context.href
+      case 'genes':
+      case 'generic_gene':
+      case 'current_fairs':
+      case 'live_auctions':
+        url = context && context.href
         break
```
alloy: Ditto.

lib/components/home/artwork_rails/artwork_rail_header.js
```diff
@@ -154,6 +166,7 @@ export default Relay.createContainer(ArtworkRailHeader, {
     rail: () => Relay.QL`
       fragment on HomePageArtworkModule {
         title
+        key
         context {
           __typename
```
alloy: Ditto.

lib/components/home/artwork_rails/artwork_rail.js
```diff
@@ -35,44 +34,38 @@ class ArtworkRail extends React.Component {
...
-    return null
+  hasAdditionalContent() {
+    const additionalContentRails = [
+      'followed_artists',
+      'saved_works',
+      'live_auctions',
+      'current_fairs',
+      'genes',
+      'generic_gene',
+    ]
```
alloy: I prefer such constants to move to the top of the file so that they stand out more.

lib/components/home/artwork_rails/artwork_rail.js
alloy: Yeah, sorry, I meant, did you figure out why those don’t have a context?

sarahscott: I think it has to do with whether or not you need more information about the rail from metaphysics. For example, the Related Artist context provides the related artist and their name. You could argue that an `href` is enough to warrant a context but maybe force uses a different url.

lib/components/home/artwork_rails/artwork_rail_header.js
sarahscott: Yes, and I was thinking about a way to share this information between the two, but I didn't want to add too much complexity at first.

lib/components/home/artwork_rails/artwork_rail.js
sarahscott: Nope!

lib/components/home/artwork_rails/artwork_rail.js
alloy: I think it makes sense to add a context even in those cases.

@broskoski any objections?

sarahscott:
Alright, responded to feedback. The repetition in the header and the rail is a concern, and I haven't looked into what people typically do in RN for components that should probably be sharing logic. In eigen, I would probably use some sort of view-modelly abstraction.

mennenia:
Yeah that's a curious problem. Probably a parent container that gives the info via props to its children?

sarahscott:
Hmm yeah that would be interesting. I could pass the header `props` from the rail.

alloy:
@mennenia @sarahscott yeah that seems like a good first approach 👍

PR #281, opened 2016-09-09
Create an iPad build

orta:
Makes release builds use the right AREmission API, and adds icons etc


mennenia:
🎉 Awesome stuff Orta

PR #283, opened 2016-09-11
[Home] First go at pull-to-refresh

sarahscott:
The `<RefreshControl>` component provided by RN looks and feels great out-of-the-box, but I'm having some Relay trouble.

The `forceFetch` approach in this pr will update all `artwork_modules` that have new `id`'s after the initial Home rendering. This is great for gene rails, which are changed often, and the Related Artist rail, which returns a new set of works and a new artist each time. I'm having trouble with the Recently Saved rail because its `id` does not change even though the works in its child `Grid` container can. I've been reading a lot about Relay but I'm not entirely sure why calling `forceFetch` on the parent Home container doesn't check for changes in the children. I did take a look at the `RootContainer` and its `forceFetch` variable, but I'm not sure what exactly I can do there to solve this.

Anyway, this implementation of pull-to-refresh will at least update some rails and looks pretty. `RefreshControl` doesn't offer much in the way of customization but I can definitely start on our custom spinner once I wrap my head around this Relay stuff.

pulltorefresh

### DIFF
--- a/lib/components/artwork_grids/generic_grid.js
+++ b/lib/components/artwork_grids/generic_grid.js
@@ -31,6 +31,10 @@ class GenericArtworksGrid extends React.Component {
     this.onLayout = this.onLayout.bind(this)
   }
 
+  componentWillReceiveProps(nextProps) {
+    this.setState({ artworks: nextProps.artworks })
+  }
+
   onLayout = (event: SyntheticEvent) => {
     const layout = event.nativeEvent.layout
     if (layout.width > 0) {
@@ -129,8 +133,6 @@ const styles = StyleSheet.create({
 })
 
 const GenericArtworksGridContainer = Relay.createContainer(GenericArtworksGrid, {
-  shouldComponentUpdate: () => true,
-
   fragments: {
     artworks: () => Relay.QL`
       fragment on Artwork @relay(plural: true) {
--- a/lib/components/artwork_grids/generic_grid.js
+++ b/lib/components/artwork_grids/generic_grid.js
@@ -9,7 +9,7 @@ import Artwork from './artwork'
 import Spinner from '../spinner'
 
 class GenericArtworksGrid extends React.Component {
-    state: {
+  state: {
     sectionDimension: number,
     artworks: [Object];
   };
@@ -129,6 +129,8 @@ const styles = StyleSheet.create({
 })
 
 const GenericArtworksGridContainer = Relay.createContainer(GenericArtworksGrid, {
+  shouldComponentUpdate: () => true,
+
   fragments: {
     artworks: () => Relay.QL`
       fragment on Artwork @relay(plural: true) {
--- a/lib/components/home/artwork_rails/artwork_rail_header.js
+++ b/lib/components/home/artwork_rails/artwork_rail_header.js
@@ -46,6 +46,7 @@ class ArtworkRailHeader extends React.Component {
   }
 
   render() {
+    console.log(this.props.rail.title)
     return (
       <TouchableWithoutFeedback onPress={this.handleViewAll}>
         <View style={styles.container}>
--- a/lib/components/home/artwork_rails/artwork_rail_header.js
+++ b/lib/components/home/artwork_rails/artwork_rail_header.js
@@ -46,7 +46,6 @@ class ArtworkRailHeader extends React.Component {
   }
 
   render() {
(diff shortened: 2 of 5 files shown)


lib/components/home/artwork_rails/artwork_rail_header.js
```diff
@@ -46,6 +46,7 @@ class ArtworkRailHeader extends React.Component {
   }
 
   render() {
+    console.log(this.props.rail.title)
```
sarahscott: Useful for seeing which rails update when you pull-to-refresh.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="true">The following files do not have tests: <a href="https://github.com/artsy/emission/blob/eeb05fd90c1999dd0099e15e69614823abb10d53/lib/components/artwork_grids/generic_grid.js">lib/components/artwork_grids/generic_grid.js</a>, <a href="https://github.com/artsy/emission/blob/eeb05fd90c1999dd0099e15e69614823abb10d53/lib/components/home/artwork_rails/artwork_rail.js">lib/components/home/artwork_rails/artwork_rail.js</a>, <a href="https://github.com/artsy/emission/blob/eeb05fd90c1999dd0099e15e69614823abb10d53/lib/components/home/artwork_rails/artwork_rail_header.js">lib/components/home/artwork_rails/artwork_rail_header.js</a> &amp; <a href="https://github.com/artsy/emission/blob/eeb05fd90c1999dd0099e15e69614823abb10d53/lib/containers/home.js">lib/containers/home.js</a></td>
    </tr>
  </tbody>
</table>

</p>

lib/components/artwork_grids/generic_grid.js
```diff
@@ -129,6 +129,8 @@ const styles = StyleSheet.create({
 })
 
 const GenericArtworksGridContainer = Relay.createContainer(GenericArtworksGrid, {
+  shouldComponentUpdate: () => true,
```
sarahscott: I was hopeful about this, but it doesn't seem to make a difference. Temporarily leaving it here in the PR so others don't go down the same rabbit hole.

mennenia: I'm not sure I understand - what had you hoped this would do?

sarahscott: It can block containers from updating in certain cases: https://facebook.github.io/relay/docs/api-reference-relay-container.html#shouldcomponentupdate

So I thought maybe overriding it would encourage an update every time.

mennenia: ahhh, awesome, thanks! 

Shame the override didn't help.

sarahscott: I've been looking at this problem and reading docs for too long so I'm excited for fresh eyes this week.

alloy:
Interesting. I wonder if this has to do with how [we do staggered loading (“manually”)](https://github.com/artsy/emission/blob/2bf112f24bf59f5e3956e5e702e526f40bb9adc8/lib/components/home/artwork_rails/artwork_rail.js#L223):

So we tell Relay to force fetch the IDs of rails to show on the home page:
* Relay sees that some IDs are different and refetches those.
* For other rails it sees the IDs are the same and doesn’t refetch those.

What we may need to do is force fetch the individual rails that we want to always be reloaded.

sarahscott:
For brevity and for everyone else to know how we solved this:

If `forceFetch` is called by `Home`, the container will check the `__id`'s of all modules and if any of them have changed, re-render the new ones. This was great for the Related Artist rail, for example, because the artist (and corresponding container `__id`) changes each time we request our `artwork_modules`. This did not, however, register changes in the children's data, so the Recently Saved rail would not be updated even if there were new favorited works (its `__id` does not change). 

By keeping an array of component `ref`s in `Home`, we can call `forceFetch` on each individual module when we refresh.

alloy:
Amaze-craze!

PR #293, opened 2016-09-14
[Home] Selective Pull-To-Refresh #tiny

sarahscott:
This pr makes our pull-to-refresh implementation a bit smarter. By calling `forceFetch` on the Home container, it ensures all entirely new modules are fetched every time. It will then check for modules that are frequently updated and `forceFetch` them individually instead of refreshing every single rail every time.

### DIFF
--- a/lib/containers/home.js
+++ b/lib/containers/home.js
@@ -8,6 +8,10 @@ import { ScrollView, RefreshControl } from 'react-native'
 import ArtworkRail from '../components/home/artwork_rails/artwork_rail'
 import ArtistRail from '../components/home/artist_rails/artist_rail'
 
+const frequentlyUpdatedModules = [
+  'saved_works',
+]
+
 class Home extends React.Component {
   constructor() {
     super()
@@ -54,7 +58,14 @@ class Home extends React.Component {
   }
   refresh() {
     this.setState({isRefreshing: true})
-    this.state.modules.forEach((module) => module.forceFetch())
+
+    // forceFetch for entirely new modules
+    this.props.relay.forceFetch()
+
+    // only individuallyforceFetch modules that have possibly changed in the past day
+    const forceFetchModules = this.state.modules.filter((module) => frequentlyUpdatedModules.indexOf(module.props.rail.key) > -1)
+    forceFetchModules.forEach((module) => module.forceFetch())
+
     this.setState({isRefreshing: false})
   }
 }
@@ -65,6 +76,7 @@ export default Relay.createContainer(Home, {
       fragment on HomePage {
         artwork_modules(max_rails: 99) {
           __id
+          key
           ${ArtworkRail.getFragment('rail')}
         }
         artist_modules {


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="true">The following files do not have tests: <a href="https://github.com/artsy/emission/blob/ca0e870414b17c16eb08b105eeadce9d627fc504/lib/containers/home.js">lib/containers/home.js</a></td>
    </tr>
  </tbody>
</table>

</p>

sarahscott:
When metaphysics updates a rail like Recommended Works with new content, is the `__id` also updated?

alloy:
> When metaphysics updates a rail like Recommended Works with new content, is the __id also updated?

It does not. We could maybe do this by adding the ETag from the Gravity response, not sure right now.

lib/containers/home.js
```diff
@@ -54,7 +58,14 @@ class Home extends React.Component {
   }
   refresh() {
     this.setState({isRefreshing: true})
-    this.state.modules.forEach((module) => module.forceFetch())
+
+    // forceFetch for entirely new modules
+    this.props.relay.forceFetch()
```
alloy: What new rails would this be?

sarahscott: Related Artist for now, and possibly generic genes if they change.

sarahscott: Or live auctions, that sort of thing.

alloy: Gotcha 👍

alloy:
So when do we refresh the rails that are not often updated? Once a day?

sarahscott:
Ideally, yes. I haven't found a way to do that yet. If we can't figure it out by release day, maybe `forceFetch` for all rails every time makes more sense. It just feels inefficient and the little engineer person in my head is bothered by that.

alloy:
> It just feels inefficient and the little engineer person in my head is bothered by that.

Totes, I get that, but I’d prefer that we include those fixes before merging, just so we don’t accidentally ship something that could leave the home page ‘broken’ (out of date content).

alloy:
I don’t think there’s a way from RN to know that a view became visible (e.g. when going back to the home tab in Eigen’s tab bar), so we might need to trigger something from the objc Home component VC. Then in that hook we check when the last refresh took place and force a full one if it’s over N time.

alloy:
I’m closing this one, we’ll have to revisit it soon.

alloy:
Hmmmm, my Safari extension auto deleted the branch, that wasn’t really my intention. The diff is still available, though, so we can always cherry-pick.

PR #299, opened 2016-09-16
Vscode fine tuning

alloy:
I’m not sure why JS files were set to be recognised as the `flow` type. It doesn’t appear to exist and JS files would get recognised as plain text files.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,6 +12,14 @@
     "editor.rulers": [120],
     "editor.tabSize": 2,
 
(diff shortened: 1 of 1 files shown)


orta:
Looks like they removed support for the fake flow language in the latest update: https://github.com/flowtype/flow-for-vscode/commit/f545d811c33f1ce3bc1946a8db0d27eb96a77c93#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L19

PR #301, opened 2016-09-21
Animate artist follow

alloy:
Closes https://github.com/artsy/eigen/issues/1440

replace animation


Example/Emission/AppDelegate.m
```diff
@@ -204,12 +216,12 @@ - (void)setupEmissionWithUserID:(NSString *)userID accessToken:(NSString *)acces
...
+//      if (randomBOOL()) {
         block(@[[NSNull null], @(following)]);
-      } else {
-        NSLog(@"Simulated follow request ‘failed’.");
-        block(@[RCTJSErrorFromNSError([NSError errorWithDomain:@"Artsy" code:42 userInfo:nil]), @(!following)]);
-      }
+//      } else {
+//        NSLog(@"Simulated follow request ‘failed’.");
+//        block(@[RCTJSErrorFromNSError([NSError errorWithDomain:@"Artsy" code:42 userInfo:nil]), @(!following)]);
+//      }
```
alloy: These simulated network errors were getting real tedious. We should re-enable that at some point, but preferably in a way where the user of the example app knows that a network error will be simulated, otherwise we forget the these happen and are baffled when things seemingly suddenly break.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="true">The following files do not have tests: <a href="https://github.com/artsy/emission/blob/52a05bc11e0106c6ea1778febfec7912bc01616a/lib/components/artist/header.js">lib/components/artist/header.js</a>, <a href="https://github.com/artsy/emission/blob/52a05bc11e0106c6ea1778febfec7912bc01616a/lib/components/buttons/inverted_button.js">lib/components/buttons/inverted_button.js</a>, <a href="https://github.com/artsy/emission/blob/52a05bc11e0106c6ea1778febfec7912bc01616a/lib/components/home/artist_rails/artist_card.js">lib/components/home/artist_rails/artist_card.js</a> &amp; <a href="https://github.com/artsy/emission/blob/52a05bc11e0106c6ea1778febfec7912bc01616a/lib/components/home/artist_rails/artist_rail.js">lib/components/home/artist_rails/artist_rail.js</a></td>
    </tr>
  </tbody>
</table>

</p>

alloy:
@mennenia @sarahscott Whoever gets to it first, and has the time, can go ahead and review.

Example/Emission.xcodeproj/project.pbxproj
```diff
@@ -222,13 +222,16 @@
...
 			buildPhases = (
+				93423CEEC7BDE3169B81C140 /* [CP] Check Pods Manifest.lock */,
 				921B840AC340AE7E64EC1E6E /* [CP] Check Pods Manifest.lock */,
 				510DCB0A1CCA69EC0075E8CB /* Sources */,
 				510DCB0B1CCA69EC0075E8CB /* Frameworks */,
 				510DCB0C1CCA69EC0075E8CB /* Resources */,
 				18F7340895C5AA6B988B03B6 /* [CP] Embed Pods Frameworks */,
 				DF68872A73AD1E49C33FE5F0 /* [CP] Copy Pods Resources */,
 				510E94FE1CCB87F100BDF098 /* Create Clang compilation DB */,
+				A8F7E39266D829B493F51B56 /* 📦 Embed Pods Frameworks */,
```
mennenia: Emoji pods are back. Did you use an older version of CocoaPods?

Example/Emission.xcodeproj/project.pbxproj
```diff
@@ -518,9 +614,8 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CC = "${SRCROOT}/compile_commands_emitting_clang";
-				CODE_SIGN_IDENTITY = "iPhone Developer";
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				DEBUG_INFORMATION_FORMAT = dwarf;
+				DEVELOPMENT_TEAM = 23KMWZ572J;
```
mennenia: Did we purposefully make these changes? Or was that just from running it locally on device and Xcode set it up?

Example/Emission.xcodeproj/project.pbxproj
```diff
@@ -615,6 +709,94 @@
 			};
 			name = Release;
 		};
+		51D5E9101D92814800C74594 /* Deploy */ = {
```
mennenia: Just checking - are these the build/deploy configurations that helped with the AppHub JS problem? :o

Example/Podfile.lock
```diff
@@ -120,4 +120,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: ab2b14498bb09c9ba993d5484696be2421070207
 
-COCOAPODS: 1.0.1
+COCOAPODS: 1.1.0.rc.2
```
mennenia: Ahhh. Does this explain the emoji pods? Does that mean they're making a comeback? 

We probably want to stick to `1.0.1` though for now, rather than move over to the RC?

lib/components/home/artist_rails/artist_rail.js
```diff
@@ -30,35 +39,88 @@ class ArtistRail extends React.Component {
...
+      // remove card when there is no suggestion
+      artists.splice(index, 1)
+    }
+    this.setState({ artists })
+    return suggestedArtist
+  }
+
+  handleFollowChange(followedArtist, setFollowButtonStatus) {
+      // Get a new suggested artist based on the followed artist.
+    return metaphysics(suggestedArtistQuery(followedArtist._id))
```
mennenia: So we still ask for the suggested artist upon a follow change (rather than upfront)? Because it doesn't look like it wasn't cached (from the animation)?

lib/components/home/artist_rails/artist_rail.js
```diff
@@ -30,35 +39,88 @@ class ArtistRail extends React.Component {
...
+      const easing = Animation.easing
+      Animated.parallel([
+        Animated.timing(opacity,    { duration, easing, toValue: 0 }),
+        Animated.timing(translateY, { duration, easing, toValue: Animation.yDelta }),
+      ]).start(resolve)
+    })
+  }
+
+  suggestedArtistAnimation(suggestedArtist) {
+    return new Promise((resolve, reject) => {
```
mennenia: Do the promises here allow you to chain these methods further down below? (where you've done the `then` and `then` - lines 98 - 108)

Example/Emission.xcodeproj/project.pbxproj
alloy: This is purposefully. I think building for device should just be transparent for all you now.

Example/Emission.xcodeproj/project.pbxproj
alloy: Aye aye, this configuration adds a CPP macro called `DEPLOY` which we can then use in the code to conditionally compile in the AppHub stuff.

Example/Podfile.lock
alloy: Hmm, interesting. Will check.

lib/components/home/artist_rails/artist_rail.js
alloy: Exactly.

lib/components/home/artist_rails/artist_rail.js
alloy: Yep, because trying to optimise a little now is probably going to open a whole can of worms, e.g. one thing I can think of now is that there might be an artist preloaded that you already start following through another suggested artist, so we need to keep state for that, add a ‘exclude artist IDs’ argument to metaphysics, etc etc. These things compound quickly.

Example/Podfile.lock
alloy: I’ve pushed a commit that reverts to use CP 1.0.1, but the emojis don’t go away, I suspect that is because CP skips the build phase as it already exists.

orta: weird, 1.1 should be `[CP]` if I recall

Example/Emission.xcodeproj/project.pbxproj
mennenia: 💥awesome

lib/components/home/artist_rails/artist_rail.js
mennenia: Fair enough! So how did the blinking disappear? :o because of the chaining / promises?

alloy: No because in my earlier attempt I was wrapping the previous and new suggestions in a single component, I would then re-render the component with a clean state after the animation finished, the re-render was what triggered it.

Example/Emission.xcodeproj/project.pbxproj
```diff
@@ -424,6 +434,21 @@
...
+			inputPaths = (
+			);
+			name = "📦 Copy Pods Resources";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "\"${SRCROOT}/Pods/Target Support Files/Pods-EmissionTests/Pods-EmissionTests-resources.sh\"\n";
+			showEnvVarsInLog = 0;
+		};
```
orta: These are duplicates I think?

orta:
using all the review features ;)

lib/components/home/artist_rails/artist_rail.js
mennenia: ahh ok, thanks! :)

Example/Emission.xcodeproj/project.pbxproj
alloy: Good catch 👍

orta:
hrm, this is going to require somehow telling the deploy script ( https://github.com/artsy/emission/blob/master/scripts/deploy_apphub.sh ) that is needs to use that specific scheme

alloy:
@orta It needs to use the Deploy configuration, not scheme. That can’t be hard right? Setting configurations is a very normal thing to do.

orta:
For Cocoa dev, these are not the kind of things RN people commonly do, will have a look through the tools, otherwise this breaks the deploy from master stuff

orta:
I'll do my research now

orta:
Looks like I was wrong, this will only affect us when we _upload_ emission.app - not the app hub JS deploys

orta:
Considering those are not automated, that's fine

PR #304, opened 2016-09-22
[Pod] Collect image assets in pod so, for now, they can be manually included in build.

### LINKED ISSUE #270: Expand icon missing in release.

alloy:
Closes #270, but it’s only a stopgap solution.


orta:
Cool, this makes sense

PR #324, opened 2016-09-28
Get down to 0 flow errors

orta:
This adds Flow to the CI, and brings us to 0 flow errors for the first time

I strongly recommend everyone read this PR from top to bottom, I'll try and cover every major change in the review mode. 

### Important

This updates React-Native to 0.34, you will probably need to `rm -rf node_modules` and `npm install` after this. 

TODO:

- [x] dig into Promises, see discussion [in slack](https://artsy.slack.com/archives/mobile-banter/p1475054984000817)
- [x] actually use the app to verify React Native update hasn't actually broken things at runtime

PR #326, opened 2016-09-29
Adds Keychain sharing

orta:
I was seeing this on every launch

screen shot 2016-09-29 at 15 13 40

Which doesn't save the credentials, so I looked up why, turns out to be a bug in keychain itself. The workaround is to turn on keychain sharing: https://forums.developer.apple.com/thread/4743

This does just that.

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -13,9 +13,9 @@ PODS:
   - Emission (1.1.0-beta.2):
     - Artsy+UIFonts (>= 1.1.0)
     - Extraction (>= 1.2.0)
-    - React/Core (~> 0.30.0)
-    - React/RCTNetwork (~> 0.30.0)
-    - React/RCTText (~> 0.30.0)
+    - React/Core (= 0.34.1)
+    - React/RCTNetwork (= 0.34.1)
+    - React/RCTText (= 0.34.1)
     - SDWebImage (>= 3.7.2)
   - Extraction (1.2.0):
     - Extraction/ARAnimationContinuation (= 1.2.0)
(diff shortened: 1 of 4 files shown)


Example/Podfile.lock
```diff
@@ -46,17 +45,17 @@ PODS:
...
     - React/Core
-  - React/RCTTest (0.30.0):
+  - React/RCTTest (0.34.1):
     - React/Core
-  - React/RCTText (0.30.0):
+  - React/RCTText (0.34.1):
     - React/Core
-  - React/RCTWebSocket (0.30.0):
+  - React/RCTWebSocket (0.34.1):
     - React/Core
```
orta: looks like I missed these on #324

Example/Podfile.lock
```diff
@@ -85,9 +84,9 @@ EXTERNAL SOURCES:
   Artsy+UIFonts:
     :git: https://github.com/artsy/Artsy-UIFonts.git
   Emission:
-    :path: ../
+    :path: "../"
   React:
-    :path: ../node_modules/react-native
+    :path: "../node_modules/react-native"
```
orta: mine + your versions of the yams parser really don't like each other

alloy: What Ruby version are you using? I’m on system of OS X 10.11.6

orta: Trusty old ruby `2.1.3p242`

Example/Podfile.lock
```diff
@@ -106,18 +105,18 @@ SPEC CHECKSUMS:
...
+  React: fc15d22bec0ec10fe15c12749fbefdac37ddc541
   SAMKeychain: 1fc9ae02f576365395758b12888c84704eebc423
   SDWebImage: 69c6303e3348fba97e03f65d65d4fbc26740f461
   Specta: ac94d110b865115fe60ff2c6d7281053c6f8e8a2
   UIView+BooleanAnimations: a760be9a066036e55f298b7b7350a6cb14cfcd97
 
 PODFILE CHECKSUM: af04d610a78f4ef6211c4ff54ac79ae9aa8d0c78
 
-COCOAPODS: 1.0.1
+COCOAPODS: 1.1.0.rc.2
```
alloy: Can you re-run with 1.0.1 please?

orta:
Cool - reran as 1.0.1

alloy:
Merci!

Example/Podfile.lock
alloy: The system Ruby I use is `2.0.0p648`. We don’t have bundler in Emission, did you think of Eigen?

mennenia: Yes! Sorry. On Emission I don't change any settings... :)

alloy: Right, I believe Eigen should be possible with system Ruby, but you have to do `bundle install --without development` so that the push notification gem doesn’t get installed (which requires newer Ruby).

PR #338, opened 2016-10-12
Add docs for running eigen locally

orta:
Also adds extension recommendations for Emission inside VS Code, as I had that laying around


alloy:
👍

PR #345, opened 2016-10-12
[LoadFailure] Fix crash that occured when attempting to retry 2nd time.

alloy:
The crash occurred because the load failure view would not actually have
a `onRetry` callback while Relay is retrying, but the load failure view
would stop animating prematurely allowing you to tap the retry button
again.

Now the animation does not stop prematurely, you cannot tap the button
while it is still retrying, but you can when Relay failed again.

ARLoadFailureView fix https://github.com/artsy/extraction/pull/8

Fixes https://github.com/artsy/eigen/issues/1867

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -72,7 +72,7 @@ DEPENDENCIES:
   - Artsy+Authentication/email (from `https://github.com/artsy/Artsy-Authentication.git`, branch `fetch-user-details`)
   - Artsy+UIFonts (from `https://github.com/artsy/Artsy-UIFonts.git`)
   - Emission (from `../`)
-  - Extraction
+  - Extraction (from `https://github.com/artsy/Extraction`, branch `dont-prematurely-stop-animating-on-second-retry`)
   - FLKAutoLayout
   - React/RCTTest (from `../node_modules/react-native`)
   - React/RCTWebSocket (from `../node_modules/react-native`)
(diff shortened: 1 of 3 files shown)


Example/Podfile.lock
```diff
@@ -72,7 +72,7 @@ DEPENDENCIES:
   - Artsy+Authentication/email (from `https://github.com/artsy/Artsy-Authentication.git`, branch `fetch-user-details`)
   - Artsy+UIFonts (from `https://github.com/artsy/Artsy-UIFonts.git`)
   - Emission (from `../`)
-  - Extraction
+  - Extraction (from `https://github.com/artsy/Extraction`, branch `dont-prematurely-stop-animating-on-second-retry`)
```
mennenia: do you want to make this just Extraction again? I merged it on the repo.

mennenia:
All 👍 (just the podfile)

mennenia:
Hmm just saw https://github.com/artsy/emission/issues/346, is this why we're targeting the branch?

alloy:
Yup :)

But I can do a release of Extraction today, I don’t foresee us adding more before the Eigen release.

mennenia:
I'll leave it to you :) Happy to merge now and when you do the release the ticket will just remind us anyway.

alloy:
Ok let’s merge now then :+1:

PR #347, opened 2016-10-13
[Example] Polish the Example app a bit

orta:
screen shot 2016-10-13 at 09 56 47

Easy Menu, with super simple to add actions:

```objc
- (ARCellData *)jumpToArtist
{
  return [self tappableCellDataWithTitle:@"Artist" selection: ^{
    id viewController = [[ARArtistComponentViewController alloc] initWithArtistID:@"alex-katz"];
    [self.navigationController pushViewController:viewController animated:YES];
  }];
}
```

screen shot 2016-10-13 at 10 07 15

If you have the Metaphysics submodule installed, you'll also see a button for picking a random artist from this [massive list of 500 artists](https://github.com/artsy/metaphysics/blob/master/schema/artist/maps/artist_title_slugs.js). Otherwise you don't see the option.

If you're on a sim in dev, you get the storybooks section too.

You can switch between staging/production, this doesn't actually log you out, as we keep no user state in the app - just keychain, it lets you jump between the two quickly.

screen shot 2016-10-13 at 10 07 21

Tightens the folder structure:

screen shot 2016-10-13 at 10 18 23

AppDelegate now only is responsible for Emission and setting up rootVC. Like a normal app 💃 

If you go to an unsupported route inside Emission, it will show a nicer VC

screen shot 2016-10-13 at 10 07 41

Which you can tap to open in Eigen if you do want to go through. If Eigen isn't installed, it does nothing. Oh well.


Example/Emission/ARAdminTableViewCell.m
```diff
@@ -0,0 +1,58 @@
...
+    if (_useSerifFont) {
+        self.textLabel.font = [UIFont serifFontWithSize:18];
+        self.detailTextLabel.font = [UIFont serifFontWithSize:18];
+    } else {
+        self.textLabel.font = [UIFont sansSerifFontWithSize:15];
+        self.detailTextLabel.font = [UIFont sansSerifFontWithSize:15];
+    }
+}
+
+- (void)layoutSubviews
```
orta: this is mainly the useful bit, good looking title and subtitles etc

Example/Emission/ARAnimatedTickView.m
```diff
@@ -0,0 +1,176 @@
+#import "ARAnimatedTickView.h"
```
orta: asset-less tick-views, that animate, made these before Folio hit 1.0

alloy:
Real nice 🎆

PR #348, opened 2016-10-13
Move auth out of the AppDelegate

orta:
Turns out I got something wrong on the staging flag, it wouldn't correctly authenticate on staging ( as the check was elsewhere ) 

So I moved all the authentication into another class. Now the AppDelegate really is only about Emission + view setup 👍

### DIFF
--- a/Example/Emission/AuthenticationManager.m
+++ b/Example/Emission/AuthenticationManager.m
@@ -1,0 +1,122 @@
+#import "AuthenticationManager.h"
+#import "ARDefaults.h"
+
+#import <Extraction/ARSpinner.h>
+#import <SAMKeychain/SAMKeychain.h>
+
+#import <Artsy+Authentication/ArtsyAuthentication.h>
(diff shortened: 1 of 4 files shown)


Example/Emission/AuthenticationManager.m
```diff
@@ -0,0 +1,122 @@
...
+@end
+
+@implementation AuthenticationManager
+
+- (instancetype)initWithService:(NSString *)service
+{
+  self = [super init];
+  _service = service;
+  [self updateFromStoredCredentials];
+  return self;
```
alloy: Tssk tssk, I prefer:

```objc
if ((self = [super init])) {
  // …
}
return self;
```

alloy:
No blockers 👍

Example/Emission/AuthenticationManager.m
orta: Aye, legit, will change that and another next PR

PR #349, opened 2016-10-13
Compile bundle when building for device

alloy:
Too easy to forget to do this.


orta:
Yep - looks legit to me

PR #368, opened 2016-10-19
Added fix for scroll view insets #tiny

sarahscott:


alloy:
👍

orta:
ah yeah - this didn't include them for the gene view

alloy:
@orta yeah, we should probably just move this to the base component VC.

PR #371, opened 2016-10-20
Moves authomaticallAdjustsScrolls to the base ARComponentVC 

### LINKED ISSUE #370: gene view header area is clashing with back button
on the left is our auction sale VC, on the right is current gene view

looks like the header area needs to increase in size to get the button to not clash with the fitered data

image

orta:
so it applies to all subclasses - should fix #370 and #368

### DIFF
--- a/Pod/Classes/ViewControllers/ARComponentViewController.m
+++ b/Pod/Classes/ViewControllers/ARComponentViewController.m
@@ -26,6 +26,7 @@
 - (void)viewDidLoad;
 {
   [super viewDidLoad];
+  self.automaticallyAdjustsScrollViewInsets = NO;
 
   RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:self.emission.bridge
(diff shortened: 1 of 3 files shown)


Pod/Classes/ViewControllers/ARComponentViewController.m
```diff
@@ -26,6 +26,7 @@ - (instancetype)initWithEmission:(AREmission *)emission
 - (void)viewDidLoad;
 {
   [super viewDidLoad];
+    self.automaticallyAdjustsScrollViewInsets = NO;
```
alloy: Why not in init? And has incorrect indentation 😱

orta: ideally shouldn't really be using properties in init, especially ones with side effects

did it in VS Code, will change

orta:
👍

alloy:
Grazie 👍

Issue #378, opened 2016-10-20
Amend order of rails


katarinabatina:
current logic is producing a bit of a weird experience I suggest making this adjustment 
![image](https://cloud.githubusercontent.com/assets/6482926/20306948/1addaef6-ab0b-11e6-8638-295187f5b009.png)

katarinabatina:
Add logic/reasoning

mennenia:
Not sure if the change in #394 is what you had in mind, @katarinabatina?

Moving this to QA, I can shuffle orders some more if you'd like.

katarinabatina:
@mennenia just took a look at your gif, it looks like you have the photography gene showing up above works by artist you follow (but its towards the end of the gif so I might just be getting confused between the beginning and end). 

There are several changes that this new order reflects

- getting genes to show up underneath current auctions/fairs
- getting saved works to show up under galleries you follow
- choosing one artist you follow and displaying an entire rail of their for sale works, followed by a recommended artists (based on artist x)

mennenia:
@katarinabatina Yeah, photography is the last one! It's after Works by Artists You Follow and so on. 

Cool, that's helpful, thanks.

So now I have:

- Recommended works
- Followed artists 
- Related artists
- Followed galleries
- Saved works
- Live auctions
- Followed genes
- Current fairs
- Generic genes

This correct? (e.g. in your comment you mention the artist + related artist after saved works, but originally it was higher up (because you'd have an artist before a gallery). Wasn't sure whether to switch that order!)

alloy:
From the PR #394:

> @alloy - it still puts in the artist rail after 2 artwork rails. Is that ok for now / was there a plan to change that in the future? (e.g. in the new order, we sometimes may want a few other rails before showing trending artists, but it's on a case by case basis).

Yeah that’s very simplistic logic done on the client-side, mainly because the app’s layout for these differ so much from Force. Can you file a separate ticket to track that issue and then we can prioritise it accordingly next Monday?

alloy:
Closing in favor of #399.

PR #394, opened 2016-11-15
[Home] Move saved works rail further to the top

mennenia:
For #378. 

Current order:

- Recommended works
- Followed artists
- Related artists
- Followed galleries
- Saved works
- Live auctions
- Current fairs
- Followed genes
- Generic genes

@alloy - it still puts in the artist rail after 2 artwork rails. Is that ok for now / was there a plan to change that in the future? (e.g. in the new order, we sometimes may want a few other rails before showing trending artists, but it's on a case by case basis).


alloy:
Awesome 👍

PR #396, opened 2016-11-15
Added some pagination responsibility to infinite_scroll_grid

sarahscott:
Fixes https://github.com/artsy/eigen/issues/1984

The `infinite_scroll_grid` was assuming the first page of artworks would be passed in as the `artworks` prop and would start its `page` state at 1. With this fix, the grid will start at page 0 if no artworks are passed in as props.

simulator screen shot nov 15 2016 5 33 47 pm

### DIFF
--- a/Example/Emission/ARRootViewController.m
+++ b/Example/Emission/ARRootViewController.m
@@ -119,7 +119,7 @@
 - (ARCellData *)jumpToGene
 {
   return [self tappableCellDataWithTitle:@"Gene" selection: ^{
-    id viewController = [[ARGeneComponentViewController alloc] initWithGeneID:@"contemporary-pop"];
+    id viewController = [[ARGeneComponentViewController alloc] initWithGeneID:@"website"];
     [self.navigationController pushViewController:viewController animated:YES];
   }];
 }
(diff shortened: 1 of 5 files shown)


Example/Emission/ARRootViewController.m
```diff
@@ -119,7 +119,7 @@ - (ARCellData *)jumpToHomepage
 - (ARCellData *)jumpToGene
 {
   return [self tappableCellDataWithTitle:@"Gene" selection: ^{
-    id viewController = [[ARGeneComponentViewController alloc] initWithGeneID:@"contemporary-pop"];
+    id viewController = [[ARGeneComponentViewController alloc] initWithGeneID:@"website"];
```
sarahscott: Not all that relevant; can revert to contemporary pop if need be

Example/Podfile.lock
```diff
@@ -87,9 +87,9 @@ EXTERNAL SOURCES:
   Artsy+UIFonts:
     :git: https://github.com/artsy/Artsy-UIFonts.git
   Emission:
-    :path: ../
+    :path: "../"
```
sarahscott: Does this seem right?

Example/Emission/ARRootViewController.m
orta: e.g. these

Example/Podfile.lock
orta: as it's the only change, it's probably worth reverting

alloy: Agreed.

alloy:
Can you add a CHANGELOG entry?

PR #400, opened 2016-11-22
[RelatedArtists] Use layoutEvent instead of Dimensions

sarahscott:
Fixes https://github.com/artsy/eigen/issues/2024

I thought this was addressed but I realized it's one of those layout bugs that can only really be seen on-device. I think this is a pattern we should adopt everywhere because it really doesn't seem like `Dimensions` is a reliable source for layout.

### DIFF
--- a/lib/components/related_artists/index.js
+++ b/lib/components/related_artists/index.js
@@ -24,6 +24,8 @@ class RelatedArtists extends React.Component {
           height: 1,
         },
     }
+
+
   }
 
   layoutState(currentLayout) : Object {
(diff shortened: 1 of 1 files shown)


lib/components/related_artists/index.js
```diff
@@ -16,24 +17,30 @@ class RelatedArtists extends React.Component {
 
   constructor(props) {
     super(props)
-    this.state = this.layoutState()
+    this.state = {
+        columns: 0,
+        imageSize: {
+          width: 1,
```
sarahscott: `opaque_image_view` needs some sort of aspect ratio or dimensions to render, so I start with 1.

lib/components/related_artists/index.js
```diff
@@ -16,24 +17,30 @@ class RelatedArtists extends React.Component {
...
+    const isPadHorizontal = width > 900
 
     // TODO: Document what these margins are based on.
     let columnCount
     let margins
     if (isPad) {
       if (isPadHorizontal) {
         columnCount = 4
-        margins = 140
+        margins = 90
```
sarahscott: I really want to make these calculated the way they are [here](https://github.com/artsy/emission/blob/master/lib/components/artwork_grids/generic_grid.js#L38), so that's why this is still a WIP.

lib/components/related_artists/index.js
sarahscott: Happy to hear suggestions though because this does feel a bit inelegant. I'm guessing this could be fixed by doing some work in the image view or by not rendering images at all until after layout?

orta: @alloy ^

mennenia:
Nice one!

mennenia:
It seems something is up with CI (it's timing out) so I'm just gonna go ahead and merge.

PR #403, opened 2016-11-28
Update ESLint and Jest

orta:
Updated: `eslint-plugin-flowtype@2.28.2, babel-eslint@7.1.1, babel-jest@17.0.2, danger@0.6.3, eslint-plugin-flow-vars@0.5.0, eslint-plugin-react@6.7.1, jest@17.0.3, jest-cli@17.0.3, jest-react-native@17.0.3, eslint@3.11.0`

I may make something in Danger to echo this ^

Mainly: Updates ESLint, and Jest to latest. We now have 1 warning in the app. I don't want to fix that though, it could change behavrior, need tests to cover it first. Will look at that soon.


lib/components/__tests__/__snapshots__/tab_view-tests.js.snap
```diff
@@ -6,19 +6,19 @@ exports[`test looks like expected 1`] = `
         Object {
           "alignSelf": "flex-start",
           "borderColor": "red",
-          "borderWidth": 1
+          "borderWidth": 1,
```
orta: looks like newer Jest snapshots add dangling commas, probably to make diffs easier to read

lib/components/buttons/inverted_button.js
```diff
@@ -2,7 +2,7 @@
 'use strict'
 
 import React from 'react'
-import { Animated, Dimensions, StyleSheet, TouchableHighlight, View } from 'react-native'
```
orta: not used

lib/components/gene/__tests__/__snapshots__/about-test.js.snap
```diff
@@ -0,0 +1,197 @@
...
+              "lineHeight": 20,
+              "marginBottom": 20,
+            },
+            Object {
+              "fontFamily": "AGaramondPro-Regular",
+            },
+          ]
+        }>
+        This is some example text
+      </Text>
```
orta: this is the rendered content for the gene about tab

screen shot 2016-11-28 at 16 53 12

basically nothing on this one, this file represents all the tests in the test file, so it's not 1 test 1 file.

lib/components/gene/__tests__/__snapshots__/about-test.js.snap
```diff
@@ -0,0 +1,197 @@
...
+          Object {
+            "backgroundColor": "#e5e5e5",
+            "height": 1,
+            "marginLeft": -20,
+            "marginRight": -20,
+          },
+          Object {
+            "marginBottom": 20,
+          },
+        ]
```
orta: the separator

lib/components/gene/__tests__/__snapshots__/about-test.js.snap
```diff
@@ -0,0 +1,197 @@
...
+      } />
+  </View>
+  <View
```
orta: same as above but now we have related artists

screen shot 2016-11-28 at 16 54 57

lib/components/gene/__tests__/__snapshots__/about-test.js.snap
```diff
@@ -0,0 +1,197 @@
...
+              Object {
+                "color": "#666666",
+                "fontFamily": "AGaramondPro-Regular",
+                "fontSize": 16,
+                "marginTop": 5,
+              }
+            }>
+            6 works for sale
+          </Text>
+        </View>
```
orta: so we get 6 works for sale ^

lib/components/gene/__tests__/about-test.js
```diff
@@ -0,0 +1,37 @@
...
+jest.mock('../../opaque_image_view.js', () => 'ImageView')
+
+it('renders correctly', () => {
+  const gene = {
+    description: 'This is some example text',
+    trending_artists : []
+  }
+  const about = renderer.create(<About gene={gene}/>).toJSON()
+  expect(about).toMatchSnapshot()
+})
```
orta: We can effectively ignore Relay for this, we are just using the props directly ( not triggering a Relay API grab ( due to [this file](https://github.com/artsy/emission/blob/master/lib/__mocks__/react-relay.js) ) )

lib/components/gene/__tests__/about-test.js
```diff
@@ -0,0 +1,37 @@
+import 'react-native'
+import React from 'react'
+import About from '../about'
+
+// Note: test renderer must be required after react-native.
+import renderer from 'react-test-renderer'
+
+jest.mock('../../opaque_image_view.js', () => 'ImageView')
```
orta: this line is important, we need to provide a mock for any native Components - in this case it will be treated as a blank component with the name `ImageView` in the snapshot

lib/components/__tests__/__snapshots__/tab_view-tests.js.snap
mennenia: I'm a big fan of dangling commas actually

lib/components/gene/__tests__/__snapshots__/about-test.js.snap
mennenia: so, the image you've got here is with actual content though right?

What you've got in this file, is a very minimal version of it.

Also, is the idea that, for the test, you write out the view xml? or was this partially generated or something?

lib/components/gene/__tests__/about-test.js
mennenia: so you mock the import basically, and you know that there will be a "box" of that size and positioning there but it's not about what's in that component, right?

orta: yeah pretty much! 

Only difference is that we're looking at the DOM tree, rather than it applied to a screen - so there's no layout information outside of what CSS attributes we've declared

lib/components/gene/__tests__/__snapshots__/about-test.js.snap
orta: This is just me taking a screenshot to give you a reference at what the snapshot represents, you don't really need to do this. This is the real thing, nothing to do with the test itself.

mennenia: because it's comparing the dom, rather than the image :). cool

mennenia:
Cool! This looks great @orta 🎉 

For anyone reading along, you create snapshots like we did before but instead of an image file being generated, you get a JSON / DOM representation of the components.

I got confused for a sec, because I thought the .snap file was actually manually made to specify what you wanted the view to look like, heh. 

Very keen to get started with this.

sarahscott:
@orta Is there a reason to adopt this approach instead of [this one](https://facebook.github.io/react-native/docs/testing.html#integration-tests-ios)?

orta:
You _can_ do both - but you can't do tests that run in the simulator at the same time as working on emission, I'd rather be able to work and write tests simultaneously. 

These tests don't require running the entire iOS infra, and thus would be cross platform and instant.

PR #405, opened 2016-11-30
Update tests for the Gene Component

orta:
I started looking at adding the refine button back, but until I'm sure what the problem was, I couldn't really find any.

This PR doesn't add it back, but it does add back the tests that I removed from the Gene component. I can add the refine button back in another PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,12 @@
+cache: yarn
+
 language: node_js
 node_js:
   - "6.1"
 
 install:
-  - npm install
+  - npm install -g yarn
+  - yarn install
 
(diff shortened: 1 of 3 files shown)


mennenia:
Nice one!

.travis.yml
```diff
@@ -1,9 +1,12 @@
...
 node_js:
   - "6.1"
 
 install:
-  - npm install
+  - npm install -g yarn
+  - yarn install
 
 after_script:
   - npm run flow
```
alloy: Why are these all still using `npm` ?

PR #406, opened 2016-12-01
Gene Refine price fixes

orta:
Initial work on Gene price range fixes, looks like in. at least one I was using `this.props` instead of `this.state` - probably more.


.vscode/launch.json
```diff
@@ -21,6 +21,21 @@
...
+      "address": "localhost",
+      "stopOnEntry": false,
+      "runtimeExecutable": null,
+      "runtimeArgs": [
+        "--debug-brk",
+        "./node_modules/.bin/jest",
+        "-i"
+      ],
+      "cwd": "${workspaceRoot}"
     }
```
orta: I was planning on adding this to vscode-jest, but for now this is a pretty good tool for poking around at the state of your tests.

lib/containers/__tests__/__snapshots__/gene-tests.js.snap
```diff
@@ -205,8 +205,47 @@ exports[`test looks like expected 1`] = `
                 },
               ]
             }>
-            12 works ・ 
+            12 works ・ Painting ・ Below $100,000
           </Text>
```
orta: previously it was undefined, so it didn't show up on the bar

lib/containers/__tests__/gene-tests.js
```diff
@@ -5,6 +5,9 @@ import 'react-native'
 import React from 'react'
 import renderer from 'react-test-renderer'
 
+var refineCallbackPromise = () => Promise.resolve()
+jest.mock('../../native_modules/refine_callback.js', () => { return { triggerRefine: () => refineCallbackPromise() }})
+
```
orta: replaces the function which normally [triggers the native code](https://github.com/artsy/emission/blob/master/lib/native_modules/refine_callback.js#L8) with our own instantly resolved promise

lib/components/gene/__tests__/about-test.js
```diff
@@ -20,7 +21,7 @@ it('shows trending artists correctly', () => {
   const gene = {
     description: 'This is some example text',
     trending_artists : [{
-      _id: '1',
+      __id: 'artist-uuid',
```
orta: this is what we use as a key for the components in the related_artists child array, not including it gives an error

orta:
OK, this and https://github.com/artsy/eigen/pull/1960 are good to go

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>No CHANGELOG added.</td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

orta:
Moving back to WIP, did some QA with @katarinabatina 

- [ ] all mediums should be a choice
-  [ ] Recently added should be replaced with Recently Updated
- [ ] Sold works should not be showing up at all
- [ ] Double check ascending and descending price sorts are correct
- [ ] hide refine results when parameters are not applied (eg. no need to show $0)
- [ ] fix text clipping on filter results

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>Changes were made to package.json, but not to yarn.lock - <i>Perhaps you need to run `yarn install`?</i></td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

orta:
Note: on staging it is possible to see sold works inside the filtered_artworks ( [context](https://artsy.slack.com/archives/collector-gmv-dev/p1485532487001745) ) this is not on production

orta:
Hrm, I have no idea what's going on with that push CI build

orta:
Gonna try re-run the CI

orta:
Interesting, same issue , I've deleted all travis caches

screen shot 2017-01-30 at 10 38 00 am

orta:
Yep - was the cache

alloy:
Should we merge now or do you want to wait and then update against the TS work?

alloy:
@sarahscott @l2succes Should we merge now or do you want to wait and then update against the TS work?

sarahscott:
We can merge now and then update it to TS.

PR #409, opened 2016-12-02
[Testing] Test layout in RelatedArtist + Gene test fix

sarahscott:
Jest doesn't automatically call `onLayout`, so this test started failing after I [updated Related Artists with a more responsive approach](https://github.com/artsy/emission/pull/400).

I put together a proof-of-concept test for injecting `width` into `onLayout` in the `RelatedArtist` component. The nice thing about this approach is we can specify any dimensions we want in the mocked `nativeEvent` and can make them correspond to actual expected values for iPhone and iPad.


lib/components/gene/__tests__/__snapshots__/about-test.js.snap
```diff
@@ -143,15 +143,15 @@ exports[`test shows trending artists correctly 1`] = `
...
               "paddingBottom": 20,
-              "width": 157,
+              "width": 1,
             }
           }>
           <ImageView
             imageURL=""
             style={
               Object {
-                "height": 105,
```
sarahscott: This was happening because the layout would rely on `Dimensions`, which would assume iPhone dimensions if `window` returned nothing.

lib/components/related_artists/__tests__/__snapshots__/layout-test.js.snap
```diff
@@ -0,0 +1,105 @@
...
+      }
+    }>
+    <TouchableWithoutFeedback
+      onPress={[Function]}>
+      <View
+        style={
+          Object {
+            "margin": 5,
+            "paddingBottom": 20,
+            "width": 242,
```
sarahscott: Now we have a calculated value for width based on the mocked `nativeEvent` width.

lib/components/related_artists/__tests__/layout-test.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  let component = renderer.create(<RelatedArtists artists={artists}/>)
+  let tree = component.toJSON()
+
+  // manually trigger onLayout with mocked nativeEvent
+  tree.props.onLayout(mockNativeEvent)
+
+  // re-render
+  tree = component.toJSON()
```
sarahscott: This dance is admittedly a bit cumbersome but could be wrapped?

sarahscott: I took a lot of inspiration from this: https://facebook.github.io/jest/docs/tutorial-react.html#snapshot-testing

lib/components/related_artists/__tests__/layout-test.js
```diff
@@ -0,0 +1,38 @@
+import 'react-native'
+import React from 'react'
+import RelatedArtists from '../'
+
+import renderer from 'react-test-renderer'
+
+jest.mock('../../opaque_image_view.js', () => 'ImageView')
+
+it('lays out correctly', () => {
```
mennenia: so we could make a few of these, for iPhone and iPad, say?

lib/components/related_artists/__tests__/layout-test.js
```diff
@@ -0,0 +1,38 @@
...
+      for_sale_artworks: 2,
+      artworks: 4,
+    },
+  }]
+
+  // mock nativeEvent with desired dimensions
+  const mockNativeEvent = {
+    nativeEvent: {
+      layout: {
+        width: 768,
```
mennenia: and then change the associated widths here - and get different dom outputs

lib/components/related_artists/__tests__/layout-test.js
```diff
@@ -0,0 +1,38 @@
...
+      }
+    }
+  }
+
+  let component = renderer.create(<RelatedArtists artists={artists}/>)
+  let tree = component.toJSON()
```
mennenia: so you have to json-ify the component, trigger onlayout, and then jsonify again? e.g. skipping L29 wouldn't work? :o

lib/components/related_artists/__tests__/layout-test.js
sarahscott: So this is great for testing layout logic in the component itself but not in anything it instantiates. It would be lovely to have some sort of global `LayoutEvent` type thing, so I'll look into that next.

mennenia: ohhh interesting. Thanks for explaining!

lib/components/related_artists/__tests__/layout-test.js
sarahscott: Yeah... that's the annoying thing. I am using the approach [here](https://facebook.github.io/jest/docs/tutorial-react.html#snapshot-testing) for manually triggering a callback.

sarahscott: This could be wrapped in a helper method though. Something like `runOnLayout(component)`.

mennenia:
Great stuff!

sarahscott:
Maybe we can mock [`LayoutEvent`s ](https://github.com/artsy/emission/blob/master/lib/system/events.js) the way we mock `ImageView` [here](https://github.com/artsy/emission/blob/master/lib/components/gene/__tests__/about-test.js#L8)

lib/components/related_artists/__tests__/layout-test.js
alloy: Agreed, the toJSON and tree var are a little confusing. I'd suggest having that helper take the dimensions and do the mocking in there as well.

orta:
Shouldn't need to do mocking, as an event is just a normal javascript object - you can pass in whatever you want, here's a test where I pass in a NativeEvent -> https://github.com/artsy/emission/pull/406/files#diff-0bc03f13698e18df3c3bb6d44c2f6e75R41

lib/components/related_artists/__tests__/layout-test.js
orta: yeah, this all looks 👍

orta:
Aye, looks a tad cumbersome - but that's what they're doing - seems like the right way to do it to me @sarahscott  💯

sarahscott:
I think I can timebox myself for the wrapper method in this PR - it would be a nice precedent to set. The reason I would consider mocking is so that a parent component like the `About` view could instantiate all of its children using the same `LayoutEvent` instead of having to pass that down through props or something. 

I think if we make the assumption that parent components are only testing logic and fonts and maybe their own layouts, and that the children in their snapshots will be layout-less, the approach in this pr is fine. It would be nice if the snapshots were more accurate though.

orta:
@sarahscott did you look at this?

sarahscott:
All set for review ☑️

alloy:
@sarahscott Awesome, loving that test helper 👌 💯 

Just a question about the location on disk, though. I’m still getting used to the “tests next to the lib files” convention of Jest, is `lib/testing_utilities` the convention used in the community? And seeing as the directory is called `testing_utilities` shouldn’t individual helpers be in separate files (that use `export default TheFunction`) rather than having a single `helpers.js` file?

sarahscott:
I was looking at how metaphysics organizes smaller helper functions and modeled the `helpers.js` file after [this](https://github.com/artsy/metaphysics/blob/master/schema/artwork/utilities.js) because I assume there will be other helper methods in the near future. I'm happy to adopt the convention of one-file-per-function if there's a good reason to, though.

I'll look into conventions within the community today. I was thinking the `testing_utilities` folder would be a place for methods that can be used by tests for any container or component, even though the tests themselves are next to the lib files.

orta:
I feel things like these should probably go in the `__tests__` folders, I wouldn't think to look right back at the root, closest to what they are the domain of, so I'd have expected this file in `lib/containers/__test__/helpers/layout.js`

I looked through the Jest source code, and couldn't find any examples of using helper functions in there tests. So I couldn't get anything useful from there.

lib/testing_utilities/helpers.js
```diff
@@ -0,0 +1,19 @@
+import renderer from 'react-test-renderer'
+
+export const setupWithLayout = (component, layout) => {
```
orta: this should probably be more descriptive, and ideally be typed - I wouldn't be able to know from the name what objects to put in, perhaps something like?

```js
/** Renders a React Component after applying a nativeEvent to an onLayout function passed in to props */
export const jsonAfterApplyingNativeLayoutDimentions = (component: any, layout: { width?: number, height?:number } ) => {

```

sarahscott:
But what if a component (instead of a container) needs those methods? Would we traverse from the `components` folder into the `containers` folder's `__tests__`?

sarahscott:
I know I'm going to need the exact same method for testing components. If we really don't want a `lib`- or project-wide `testing_utilities` folder, I think the only other way would be to make a separate `utilities` folder or file for both `components` and `containers`. The reason I didn't do this here is because that would require code duplication. If it would really improve our testing experience though, I'm open to that.

lib/testing_utilities/helpers.js
```diff
@@ -1,6 +1,7 @@
 import renderer from 'react-test-renderer'
 
-export const setupWithLayout = (component, layout) => {
+/** Renders a React Component with specified layout using onLayout callback */
+export const renderWithLayout = (component: any, layout: { width?: number, height?:number }) => {
```
alloy: Inconsistent spacing in layout type info.

sarahscott:
@orta bump

orta:
Alright - yeah, that makes sense, those two are both siblings on the same hierarchy, so IMO it should probably go in `lib/__tests_/helpers/file.js`

sarahscott:
I think using a `__tests__` folder inside `lib` would be confusing because it implies there are actual tests in that folder when actually there are only utilities. I'd like to keep the folder named as it is for now and if it proves to be an issue we can change it later.

orta:
Personally, I'm not sold, but if you're sure 👍 

But you shouldn't move the `__mocks__` folder into that folder, I'm surprised my tests pass at all with that changed.

sarahscott:
I see - `__mocks__` is [supposed to go in the same parent directory as `node_modules`](https://facebook.github.io/jest/docs/manual-mocks.html) (it was previously inside `lib`). 

@alloy do you have an opinion on the name for this shared directory of testing utilities? Is `__tests__` significantly clearer?

orta:
> I see - __mocks__ is supposed to go in the same parent directory as node_modules (it was previously inside lib).

ah nice one 👍 

My point is to _not_ make a generic `testing_utilities` folder, but to say it provides functions whose scope is from the lib folder downwards, and that like all testing related code - you can find it inside a `__tests__` folder.

alloy:
@sarahscott I agree with @orta 👍

orta:
👍

PR #411, opened 2016-12-03
Use node 6.7 on travis

orta:
https://github.com/yarnpkg/yarn/issues/764 seems to imply it is likely to help ( see https://github.com/yarnpkg/yarn/issues/764#issuecomment-256995983 )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ cache: yarn
 
 language: node_js
 node_js:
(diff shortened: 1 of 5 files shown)


orta:
Cool - the first fail is legit, as master is broken - and if this greens then we're good to go

orta:
Yeah - rockin'

PR #414, opened 2016-12-14
[Testing] Added test coverage to Artist container

sarahscott:
@xtina-starr first emission pr review 🎉 . All it does is add some tests to our Artist view:

simulator screen shot dec 14 2016 8 26 54 am

Many of our tests are similar to what you use in force with the exception of snapshots. The snapshots are the [React trees generated by Jest](https://facebook.github.io/jest/blog/2016/07/27/jest-14.html) and they're useful for catching changes in the component. Since these are the first snapshots for the Artist container, they're not really that interesting, but over time the [diffs become more useful](https://github.com/artsy/emission/pull/409/files#diff-b4f897fde7641a36650c3d7e54682bad).

### DIFF
--- a/lib/containers/__tests__/artist-tests.js
+++ b/lib/containers/__tests__/artist-tests.js
@@ -1,0 +1,102 @@
+// @flow
+
+'use strict'
+import 'react-native'
+import React from 'react'
+import renderer from 'react-test-renderer'
+
+import Artist from '../artist'
+
+jest.mock('../../components/artist/shows', () => 'Shows')
+jest.mock('../../components/artist/artworks', () => 'Artworks')
+jest.mock('../../components/artist/header.js', () => 'Header')
+jest.mock('../../components/artist/about.js', () => 'About')
+
+jest.mock('../../components/spinner.js', () => 'ARSpinner')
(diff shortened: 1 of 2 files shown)


lib/containers/__tests__/artist-tests.js
```diff
@@ -0,0 +1,102 @@
...
+jest.mock('../../components/artist/about.js', () => 'About')
+
+jest.mock('../../components/spinner.js', () => 'ARSpinner')
+jest.mock('../../components/opaque_image_view.js', () => 'AROpaqueImageView')
+jest.mock('../../components/switch_view.js', () => 'ARSwitchView')
+
+describe('availableTabs', () => {
+  it('returns nothing if artist has no metadata, shows, or works', () => {
+    const artist = new Artist()
+    artist.props = artistProps(false)
```
sarahscott: This `artistProps` method can be found down [here](https://github.com/artsy/emission/pull/414/files#diff-39671116ef994b68054be6fca954e02dR94)

xtina-starr:
Went over this PR with @orta and it looks good 👍  🎉

sarahscott:
Dankeschön  @xtina-starr and @orta !

PR #424, opened 2017-01-09
[Testing] Snapshot test coverage of Artist components

sarahscott:
- [x] Basic snapshot coverage for About, Articles, Biography, and Header
- [x] Coverage for remaining components

Should be wrapped up by sprint planning tomorrow 😄. This will give us a solid foundation upon which to add more tests. I am going to add logic tests subsequently because adding them for all of our components at once is a big undertaking.


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>No CHANGELOG added.</td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

lib/components/artist/__tests__/about-tests.js
```diff
@@ -0,0 +1,19 @@
...
+
+import About from '../about'
+
+jest.mock('../../opaque_image_view.js', () => 'AROpaqueImageView')
+
+it('renders properly', () => {
+  const artist = {
+    has_metadata: true,
+    is_display_auction_link: true,
+    articles: [],
```
mennenia: so, we're testing and rendering each component separately? So no articles here (within 'about'), but instead in its own test, as per below? (`lib/components/artist/__tests__/articles-tests.js`)

lib/components/artist/shows/__tests__/show-tests.js
```diff
@@ -0,0 +1,41 @@
...
+    status: 'running',
+    partner: {
+      name: 'Gallery'
+    },
+    location: {
+      city: 'Berlin'
+    }
+  }
+
+  const showStyles = {
```
mennenia: Interesting! How come we override the styles here? :o

lib/components/artist/__tests__/header-tests.js
```diff
@@ -0,0 +1,21 @@
...
+
+jest.mock('../../spinner.js', () => 'ARSpinner')
+
+it('renders properly', () => {
+  const artist = {
+      name: 'Marcel Duchamp',
+      nationality: 'French',
+      birthday: '11/17/1992',
+      counts: {
+        follows: 22,
```
mennenia: What are your thoughts on testing the follow button? or do we have examples of followed and unfollowed artists thoughout the snap suite? (I guess so, come to think of it)

lib/components/artist/shows/__tests__/show-tests.js
orta: It normally comes in from it's parent component ([VariableSizeShowsList](https://github.com/artsy/emission/blob/master/lib/components/artist/shows/variable_size_shows_list.js#L62])) and has no defaults to fall back to.

This test will implicitly validate that any style passed in via props would be re-applied on the component

mennenia: I see - great :)

lib/components/artist/__tests__/about-tests.js
sarahscott: For now, yes. I'm covering every component with the most basic snapshot but eventually they should have more comprehensive tests. I figure a very basic foundation for all components is better than having several components that are extremely well-tested and others that are not tested at all.

There are just so many components and so little time.

lib/components/artist/__tests__/header-tests.js
sarahscott: I definitely agree that we should test the follow button for those cases. It could use some logic tests but for now I'm just doing snapshots.

lib/components/artist/__tests__/about-tests.js
mennenia: totally! 👍

mennenia:
Merging! 💃

PR #426, opened 2017-01-11
[Testing] Snapshot coverage for artwork grid components

sarahscott:
These are basic snapshot tests for the artwork grids; they all definitely need additional logic tests but that will be done in a new PR (feel free to take them on 😉 ).

### DIFF
--- a/lib/components/artwork_grids/__tests__/infinite-scroll-grid-tests.js
+++ b/lib/components/artwork_grids/__tests__/infinite-scroll-grid-tests.js
@@ -1,0 +1,40 @@
+import 'react-native'
+
+import React from 'react'
+import { renderWithLayout } from '../../../tests/render_with_layout'
+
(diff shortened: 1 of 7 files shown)


lib/components/artwork_grids/__tests__/infinite-scroll-grid-tests.js
```diff
@@ -0,0 +1,40 @@
...
+jest.mock('../../spinner.js', () => 'ARSpinner')
+
+it('renders properly', () => {
+  const artworks = [ artwork(), artwork(), artwork() ]
+
+  const grid = renderWithLayout(<InfiniteScrollArtworksGrid artworks={artworks} />, { width: 768})
+  expect(grid).toMatchSnapshot()
+})
+
+var artwork = () => {
```
orta: hrm, ideally our linters be complaining that this should be a `const`, it's not something being mutated

PR #430, opened 2017-01-18
[ArtistRail] fixes text position when bio is missing

l2succes:
Fixes https://github.com/artsy/eigen/issues/1997


lib/components/home/artist_rails/__tests__/artist_card-tests.js
```diff
@@ -0,0 +1,42 @@
...
+    formatted_artworks_count: "14 works, 3 for sale",
+    href: "/artist/juan-gris",
+    id: "juan-gris",
+    image: {
+      _urlZizbi: "https://d32dm0rphc51dk.cloudfront.net/wGMxL6TvlSORJzEHZsK9JA/large.jpg"
+    },
+    name: "Juan Gris",
+  }
+  const tree = renderer.create(<ArtistCard artist={artist} />).toJSON()
+  expect(tree).toMatchSnapshot()
```
orta: I wonder if we can go better on making the expectation say the tests intentions, e.g. is there a way to say something like "expect(tree).toNotInclude(<Thing/>)`

orta: Interesting, looks like [react-test-renderer](https://github.com/thejameskyle/react-test-renderer#examples) comes with an API for this sort of thing.

It'd look something like this: 
```js
 const artist = {
    ...
    }

    const tree = renderer.create(<ArtistCard artist={artist} />)
    const heading = tree.find(element => element.type === 'Text' && [some more checks]);
    expect(heading).toBeFasly()
  })
```

However realistically, this test is probably better done as a simple unit test without snapshots [like these](https://github.com/artsy/emission/blob/master/lib/containers/__tests__/home-tests.js#L20-L27)

```js
const artist = ...  
const artistCard = new ArtistCard({artist})
expect(artistCard.renderMetadata).toBeFalsy()
```

This is because your snapshot changes could break because of completely different reasons to what the test is about. If someone renames an element in the ArtistCard, your test breaks, making it feel more brittle than it is. Then it's really quick for you to C&P that tests and add ones for the other bio `<Text>` parts

orta: FWIW, I cover some of these concepts like "when is it a good time to use snapshots vs units"  in https://github.com/orta/pragmatic-testing/

l2succes: Right, ok let me look into `react-test-renderer` see what I can use it to solidify that test.

l2succes: wut!? lol, you just so happen to have written a book on pragmatic testing 🙌🏿 Thanks man, I'll read it

l2succes:
@orta I believe that have addressed the points you brought up concerning the test, let me know if/how I could improve it further.

lib/components/home/artist_rails/__tests__/artist_card-tests.js
```diff
@@ -0,0 +1,42 @@
...
+  expect(metadata[0].props.children).toEqual(props.artist.name.toUpperCase())
+  expect(metadata[1].props.children).toEqual(props.artist.formatted_artworks_count)
+})
+
+const artistProps = (hasBio?: boolean = true) => {
+  return {
+    artist: {
+      __id: "QXJ0aXN0Omp1YW4tZ3Jpcw==",
+      formatted_artworks_count: "14 works, 3 for sale",
+      formatted_nationality_and_birthday: hasBio ? "Spanish, 1887–1927" : "",
```
orta: I'm intrigued on why this passes, shouldn't a `""` pass with this statement `if (artist.formatted_nationality_and_birthday) {`

lib/components/home/artist_rails/__tests__/artist_card-tests.js
```diff
@@ -0,0 +1,42 @@
...
+
+it('hides bio when metadata missing correctly', () => {
+  const props = artistProps(false)
+  const artistComponent = new ArtistCard(props)
+  const metadata = artistComponent.renderMetadata()
+
+  expect(metadata).toHaveLength(2)
+  expect(metadata[0].props.children).toEqual(props.artist.name.toUpperCase())
+  expect(metadata[1].props.children).toEqual(props.artist.formatted_artworks_count)
+})
```
orta: you might want to consider renaming the test to be something like `"only shows name and artwork count when bio is omitted"` as that is what the expectations are saying here 👍

orta: The concept feels good though 💯

lib/components/home/artist_rails/__tests__/artist_card-tests.js
l2succes: `""` is a falsy value in Javascript so it bypasses the if statement  [See here](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)

lib/components/home/artist_rails/__tests__/artist_card-tests.js
l2succes: Ok sounds good 👍🏿

lib/components/home/artist_rails/__tests__/artist_card-tests.js
orta: new to me - thanks :D

l2succes:
I need to create a branch instead to get CI checks to all pass. Closing this.

PR #330, opened 2016-10-04
Gene View Controller Polish

orta:
* Adds an Eigen-like back nav ( it doesn't go away on scroll, unlike eigen's ) is that feature deprecated? I'm not seeing it in any RN component's API

* Tightens up the UI, so it looks like most of the screenshots now, with normal margins/padding etc

* The header / sticky bits is still definitely a WIP, I've tried to show off how it feels in the GIF

PR #432, opened 2017-01-18
[Testing] Snapshots for gene components

sarahscott:


lib/components/artist/header.js
```diff
@@ -4,7 +4,6 @@
 import React from 'react'
 import Relay from 'react-relay'
 import { NativeModules, StyleSheet, View, Dimensions } from 'react-native'
-const { ARTemporaryAPIModule } = NativeModules
```
sarahscott: It seems like jest isn't able to recognize what this is unless we use `NativeModules.ARTemporaryAPIModule`.

orta:
Might not be worth merging this, as I think I have a bunch of these tests in #330

orta:
Eh, on second thought,s - seems 👍

PR #435, opened 2017-01-23
Adds react native debugger command and instructions

l2succes:
screen shot 2017-01-23 at 1 00 01 pm


alloy:
Awesome :+1:

package.json
```diff
@@ -73,6 +74,7 @@
     "jest": "^18.1.0",
     "jest-cli": "^17.0.3",
     "jest-react-native": "^17.0.3",
+    "react-native-debugger-open": "^0.3.2",
```
orta: I see no lock file changes for this?

l2succes: Yeah, actually I'll remove it for now as I included a launch command in `package.json` instead.

sarahscott:
🔥

Issue #302, opened 2016-09-21
[React] Check if the native animations driver is good enough for our needs.


alloy:
At the time of writing, the RN version we use (0.32.1) does not fully support the animations we’re doing on the home view when following an artist. However, the native animations module is under development as we speak, so need to check this each time after upgrading RN.

sarahscott:
@alloy bump

alloy:
We should definitely still give this a try, but it will be more involved than just flicking a switch and probably also more involved than a newcomer could chew on so marking this as a maintenance item.

PR #437, opened 2017-01-27
[WIP] Upgrade to latest React Native

alloy:
- [x] This uses my RN fork that has a fixed podspec: https://github.com/facebook/react-native/pull/12089
- [x] ~~This switches to use [Yoga](https://github.com/facebook/yoga) directly where we do CSS calculations in our native image module.~~ RN views have support for aspect ratios built-in nowadays.
- [ ] ~~While the build is now green, a segfault occurs in the native image module around the CSS code, but this is just because I will need to take some time to update our use of it other than fixing the symbols.~~ Fixed, it was because we were overriding the now built-in aspect ration property.
- [ ] Make sure we use a list view that does not exhibit the issue of going completely blank after being hidden for a while.
- [ ] See if built-in animations work #302


yimingtang:
Hi, thank you for your [article](http://artsy.github.io/blog/2016/08/24/On-Emission/) on emission. I'm following emission's approach in my project.

It seems that Artsy uses its own CocoaPods [specs](https://github.com/artsy/Specs). When I tried to setup my own React pod, I failed.

Since React's Yoga dependency version is something like `0.42.0.React` and you put the podspec file in react-native repository, React can't pass CocoaPods lint tool. It shows that CocoaPods can't find the specific version of Yoga pod. The official Yoga pod version is something like `1.1.0`. That prevents me from upgrading to the latest React Native.

At present, I have to use CocoaPods and NPM together. It's too tedious.

``` ruby
react_path = '../node_modules/react-native'
yoga_path = File.join(react_path, 'ReactCommon/yoga')

pod 'MyReactNativePod'
pod 'React', :path => react_path, :subspecs => ['RCTWebSocket']
pod 'Yoga', :path => yoga_path
```

I wish my Podfile neat again.

``` ruby
pod 'MyReactNativePod'
```

My question is how are you going to deal with this problem. Since you integrated emission with eigen by CocoaPods only. I think that you will have the same issue.

alloy:
The React Native podspecs are not going into the master spec repo atm, so you do need to use either the CocoaPods+NPM way or a private spec repo.

If you wish to setup your own spec repo, you can make use of [this script](https://github.com/facebook/react-native/blob/master/scripts/process-podspecs.sh) to push both the Yoga and React podspecs. use it like so: `$ env SPEC_REPO=MySpecRepoName SKIP_LINT=1 ./scripts/process-podspecs.sh` and then commit those new podspecs in your spec repo.

Having said that, you could also just make use of our spec repo.

Once you have a spec repo setup, you can integrate your pod in your app just with `pod "MyReactNativePod"`.

yimingtang:
Awesome! Thank you so much.

By doing so, I also made a little change 😄 

``` ruby
# React.podspec

# Use a forked version
s.source                  = { :git => "https://github.com/username/react-native.git", :tag => "v#{s.version}" }

```

alloy:
Ah yeah, good point, can do that too 👍

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>No CHANGELOG added.</td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

alloy:
Alright, opted to merge this now, rather than later, so that nobody has to go through the merge hell that this was.

Still need to continue to try to use the new experimental FlatList, but in the meantime be sure to keep an eye on whether or not you see the home view going blank (the old bug).

alloy:
Closing in favour of https://github.com/artsy/emission/pull/453

PR #438, opened 2017-01-30
[WIP] Typescript conversion

l2succes:
This migrates the codebase from flow to typescript for type definitions.
Fixed just enough to get top level containers to run in the simulator. 

_You may need to run `tsc --watch` if files are not automatically being compiled by VS Code (didn't work for me but will fix)._

To-do list:

- [ ] Add types to remaining classes
- [x] Fix tests
- [ ] Fix hot reloading


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>No CHANGELOG added.</td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

l2succes:
@sarahscott I'm currently working on defining types and fixing errors for the `Gene` container and its sub-components. So feel free to work on any of the other containers. I should have that work pushed to this branch hopefully by EOD so you can be aligned on conventions.

l2succes:
@orta Sounds good, I added the hack for now but hot-reloading doesn't work yet so either I'll try to fix it or switch to using your fork.

sarahscott:
@l2succes fantastic! I'll start on the Artist container/components.

src/lib/components/tab_view.tsx
```diff
@@ -1,14 +1,27 @@
...
+export default class TabView extends React.Component<Props, {}> {
+  static propTypes = {
+    ...SwitchView.propTypes
+  }
+
   render() {
     const { children, ...props } = this.props
     return (
       <View>
-        <SwitchView style={styles.switch} {...props} />
```
sarahscott: This spread operator gave me a lot of trouble - I did appreciate the elegance of this class though so if someone knows how to pass the props in this way without ts errors, it'll be very appreciated 😄 .

src/lib/components/artist/articles/article.tsx
```diff
@@ -1,14 +1,24 @@
...
+      name
+    }
+    thumbnail_image: {
+      url
+    }
+    thumbnail_title: string
+  }
+}
+
+class Article extends React.Component<Props, {}> {
```
alloy: I think the second `state` type should be `null` if there’s no state.

tsconfig.json
```diff
@@ -1,11 +1,27 @@
...
-        "allowJs": true
+        "jsx": "react",
+        "target": "ESNext",
+        "module": "es2015",
+        "rootDir": "src",
+        "outDir": "build",
+        "allowJs": true,
+        "sourceMap": true,
+        "moduleResolution": "node",
+        "strictNullChecks": true
```
orta: this is the good stuff

tsconfig.json
```diff
@@ -1,11 +1,27 @@
...
         "node_modules",
         "Pod/Assets",
         "Example/Build",
-        "externals/"
-    ]
+        "externals",
+        "build",
+        "index.ios.js"
+    ],
+    "compileOnSave": true
```
orta: this we want to eliminate via https://github.com/orta/systems-theory/issues/1 and https://github.com/facebook/react-native/issues/11932

Example/EmissionTests/TestApps.js
```diff
@@ -1,24 +1,21 @@
-/* @flow */
-'use strict';
-
-import Emission from 'emission';
-import React from 'react';
-import ReactNative from 'react-native';
-const { View } = ReactNative;
-const { TestModule } = ReactNative.NativeModules;
+import Emission from 'emission'
+import React from 'react'
```
alloy: Shouldn’t this be?

```
import * as React from 'react'
```

l2succes: Yeah that's correct, still working on converting all the tests. I'll have it updated soon

orta: nah, just checked, this doesn't need to change - it's still a JS file, and should stay one probably.  It's only referenced by the iOS Sim Test Runner.

src/lib/components/load_failure_view.tsx
```diff
@@ -1,19 +1,17 @@
...
 import {
   requireNativeComponent,
 } from 'react-native'
 
-export default class LoadFailureView extends React.Component {
+export default class LoadFailureView extends React.Component<any,any> {
+  static propTypes = {
+    // A callback that is called when the user requests a retry.
+    onRetry: React.PropTypes.func,
+  }
```
alloy: You should move this to a Props interface instead.

l2succes: We need to have propTypes defined for native components but yes this should be defined in the Props interface as well

alloy: Ah right, forgot about that 👍 

This should probably be documented in the README:

* JS components: only a TS interface (which inherits from x and y)
* Native components: also add prop types

l2succes:
Closing this PR in favour of the following one https://github.com/artsy/emission/pull/453

PR #441, opened 2017-02-08
@alloy - Wire artwork grid up to artworks connection [WIP] #dontmerge

broskoski:
Very much just getting the hang of things here but wanted to get some WIP up.

I think the structure is a little off right now, and I haven't settled on a good arrangement of where everything should live. `InfiniteScrollGrid` should definitely handle the scroll progress check, but I haven't decided if it's better to try and keep it as a dumb component and just send it artworks or if it should actually maintain its own query. Right now it's firing back to the `Artworks` component, which is updating the relay variable and triggering a render, but it is obviously doing that for _both_ `for_sale_artworks` and `not_for_sale_artworks`. I guess smart component it is for `InfiniteScrollGrid`, thanks for being my 🦆 

One funny gotcha I've encountered so far, you can't actually access attributes on `artwork` from outside the component who's fragment it belongs to...so for example the check for `artwork.image` returns null _unless_ you include that attribute in the query outside the fragment. I'm sure there's some better way of doing this that I don't know about.


lib/components/artist/artworks/index.js
```diff
@@ -60,48 +66,21 @@ class Artworks extends React.Component {
...
-          }
-          partner {
-            name
-          }
-          href
-        }
-      }
-    }
-  `
-  }
```
alloy: YES 🔥

lib/components/artwork_grids/infinite_scroll_grid.js
```diff
@@ -314,47 +313,5 @@ const styles = StyleSheet.create({
...
-          }
-          href
-        }
-      }
-    }
-  `
-}
-
-export default InfiniteScrollArtworksGridContainer
+export default InfiniteScrollArtworksGrid
```
alloy: Oh sweet lord, YES 🔥

alloy:
> I guess smart component it is for `InfiniteScrollGrid`

Agreed 👍 

> One funny gotcha I've encountered so far, you can't actually access attributes on artwork from outside the component who's fragment it belongs to...so for example the check for artwork.image returns null unless you include that attribute in the query outside the fragment. I'm sure there's some better way of doing this that I don't know about.

This is correct behaviour, it’s called “data masking” and ensures that a component’s behaviour can’t break due to another component’s data requirements changing.

I think it’s one of the more awesome features of Relay :bowtie: 

https://facebook.github.io/relay/docs/thinking-in-relay.html#data-masking

lib/components/artist/artworks/index.js
```diff
@@ -23,6 +24,13 @@ class Artworks extends React.Component {
     this.state = { completedForSaleWorks: false }
   }
 
+  onFetchMore(cb = () => {}) {
+    console.log('onfetchmore, this', this)
+    this.props.relay.setVariables({
+      pageSize: this.props.relay.variables.pageSize + PageSize
```
alloy: The `pageSize` variable should probably be renamed to reflect that it’s the _total_ amount of entries fetched.

Example/Podfile
```diff
@@ -11,7 +11,7 @@ target 'Emission' do
   pod 'React', :path => react_path, :subspecs => %w(RCTWebSocket)
 
   # Got to make it look right
-  pod 'Artsy+UIFonts'
+  pod 'Artsy+UIFonts', git:  'https://github.com/artsy/Artsy-OSSUIFonts.git', branch: 'always_artsy'
```
orta: I don't want use shipping this change ^

alloy:
Errrr…

lib/components/artist/artworks/index.js
```diff
@@ -27,81 +23,55 @@ class Artworks extends React.Component {
     const for_sale_count = this.props.artist.counts.for_sale_artworks
     const other_count = this.props.artist.counts.artworks - for_sale_count
     if (for_sale_count === 0) {
-      return this.renderSection('Works',
-                                other_count,
-                                this.props.artist.not_for_sale_artworks,
-                                { availability: 'IS_NOT_FOR_SALE' },
-                                this.resolveQuery,
-                                 null)
+      return this.renderSection({
```
broskoski: I think this is easier to parse, but I've been known to be wrong from time to time.

lib/components/artist/artworks/index.js
```diff
@@ -124,19 +94,17 @@ const styles = StyleSheet.create({
...
           artworks
           for_sale_artworks
         }
-        for_sale_artworks: artworks(sort: partner_updated_at_desc, filter: IS_FOR_SALE, size: 10) {
-          ${ArtworksGrid.getFragment('artworks')}
-        }
-        not_for_sale_artworks: artworks(sort: partner_updated_at_desc, filter: IS_NOT_FOR_SALE, size: 10) {
-          ${ArtworksGrid.getFragment('artworks')}
-        }
+        ${ArtworksGrid.getFragment('artist')}
```
broskoski: One fragment to rule them all

lib/components/artwork_grids/infinite_scroll_grid.js
```diff
@@ -88,64 +61,45 @@ class InfiniteScrollArtworksGrid extends React.Component {
...
     }
 
     this._sentEndForContentLength = null
-  }
-
-  // Initial setup
-  componentWillMount(){
-    if (this.state.artworks.length === 0) {
-      this.fetchNextPage()
+    if (this.props.filter) {
```
broskoski: This is the only semi-annoying thing here, we set variables on construction which as far as I can tell does not trigger any additional queries, it's just....not as nice as it could be.

Example/Podfile
broskoski: I shall remove

broskoski:
Getting closer, need to fix these tests now and figure out why metaphysics is inaccurately reporting `pageInfo.hasNext` as false.

broskoski:
Getting some weird errors here:
```
externals/metaphysics/schema/artist/index.js:3
  3: import { pageable, getPagingParameters } from 'relay-cursor-paging';
                                                   ^^^^^^^^^^^^^^^^^^^^^ relay-cursor-paging. Required module not found
externals/metaphysics/schema/artist/index.js:7
  7: } from 'graphql-relay';
            ^^^^^^^^^^^^^^^ graphql-relay. Required module not found
```
Does `yarn sync externals` not update the packages in metaphysics?

alloy:
It does a `npm install`, maybe the issue is that it needs to be `yarn install` nowadays and some deps are not being installed? https://github.com/artsy/emission/blob/0a5c6454ced2136817d21464ba9ad5eb0f572940/package.json#L19

data/schema.js
```diff
@@ -1,4 +1,4 @@
-var getbabelRelayPlugin = require('babel-relay-plugin');
-var schema = require('./schema.json');
+var getbabelRelayPlugin = require('babel-relay-plugin')
+var schema = require('./schema.json')
 
-module.exports = getbabelRelayPlugin(schema.data);
\ No newline at end of file
+module.exports = getbabelRelayPlugin(schema.data, { substituteVariables: true })
```
alloy: Oh, what’s this about?

lib/components/artist/artworks/index.js
alloy: Agreed 👍

lib/components/artist/artworks/index.js
```diff
@@ -104,7 +98,7 @@ const styles = StyleSheet.create({
 
 export default Relay.createContainer(Artworks, {
   initialVariables: {
-    pageSize: PageSize
+    totalSize: PageSize
```
alloy: This variable is no longer needed here, right?

lib/components/artist/artworks/index.js
```diff
@@ -113,26 +107,7 @@ export default Relay.createContainer(Artworks, {
...
-          edges {
-            node {
-              image {
-                id
-              }
-              ${Artwork.getFragment('artwork')}
-            }
-          }
-        }
+        ${ArtworksGrid.getFragment('artist')}
```
alloy: You probably need to pass in the filter variable here, right?

And I assume the fact that there’s only 1 grid now is just because you’re trying to get 1 to work first?

lib/components/artwork_grids/infinite_scroll_grid.js
alloy: Yeah that seems right and I agree it’s less nice.

One thing, the filter prop probably shouldn’t be optional?

lib/components/artwork_grids/infinite_scroll_grid.js
```diff
@@ -85,72 +73,51 @@ class InfiniteScrollArtworksGrid extends React.Component {
...
+      })
     }
   }
 
-  // Reset detection
-  componentDidUpdate() {
-    if (this.state.artworks.length === 0) {
-      this.fetchNextPage()
+  shouldComponentUpdate(nextProps, nextState){
+    if (this.props === undefined) {
```
alloy: When does this happen?

lib/components/artwork_grids/infinite_scroll_grid.js
alloy: Btw, I’m thinking, if you’d move this down to https://github.com/artsy/emission/pull/441/files/f2e35653d3af57d27663c4a93b34e89a71ec3b9e..3d7c0d0693b1d9fc4f590c95d43e4f37e5adae14#diff-4ae33428e5de3c1de5112e081353e759R105, then maybe it’s a bit nicer as at least the setting of variables is done in one place?

alloy: …because if I got it right in my mind, the variable isn’t needed _until_ we’d try to fetch more.

lib/components/artwork_grids/infinite_scroll_grid.js
broskoski: Ah, good catch. Leftover from before.

lib/components/artist/artworks/index.js
broskoski: It is, its the initial page size.

lib/components/artist/artworks/index.js
alloy: So good

lib/components/artist/artworks/index.js
broskoski: Nope, we set the variables in the component constructor. And both grids are working now but Metaphysics is giving 'alternative facts' about `hasNextPage`, so the Other works grid renders too soon.

data/schema.js
broskoski: Shall remove, was trying to get something wild working before I took the current approach (use two identical fragments, but set them as a different variables in the queries).

lib/components/artist/artworks/index.js
```diff
@@ -37,7 +37,7 @@ class Artworks extends React.Component {
       const otherWorks = []
       const showOtherWorks = (other_count > 0) && (for_sale_count < 10 || this.state.completedForSaleWorks)
       if (showOtherWorks) {
-        otherWorks.push(<Separator style={styles.sectionSeparator} key="separator" />)
+        otherWorks.push()
```
alloy: Why is this needed?

lib/components/artist/artworks/index.js
```diff
@@ -68,6 +70,7 @@ class Artworks extends React.Component {
           <SerifText style={styles.text}>{title}</SerifText> <SerifText style={[styles.text, styles.count]}>({count})</SerifText>
         </SerifText>
         <ArtworksGrid
+          artworks={this.props.for_sale_artworks}
```
alloy: This line shouldn’t be included, it was my (incorrect) example of using an aliased field.

PR #451, opened 2017-02-22
@alloy - Fix Gene view

broskoski:
cc @orta 

screen shot 2017-02-22 at 1 27 03 pm


broskoski:
Sigh, forgot that it needs to be a branch PR.

PR #462, opened 2017-03-20
Updates notifications using native module

sarahscott:
Adds a way to replace [this](https://github.com/artsy/eigen/blob/master/Artsy/View_Controllers/WorksForYou/ARWorksForYouViewController.m#L170)

### DIFF
--- a/src/lib/containers/works_for_you.tsx
+++ b/src/lib/containers/works_for_you.tsx
@@ -1,7 +1,7 @@
 import * as React from "react"
 import { ListView, ListViewDataSource, NativeModules, ScrollView, StyleSheet, TextStyle, View, ViewStyle }
   from "react-native"
-const { ARTemporaryAPIModule } = NativeModules
+const { ARTemporaryAPIModule, ARWorksForYouModule } = NativeModules
 import * as Relay from "react-relay"
 
 import Events from "../native_modules/events"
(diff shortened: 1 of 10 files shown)


src/lib/containers/works_for_you.tsx
```diff
@@ -32,6 +36,19 @@ export class WorksForYou extends React.Component<Props, State> {
 
   }
 
+  componentDidMount() {
+    NativeModules.ARTemporaryAPIModule.markNotificationsRead((error) => {
+      if (error) {
+        console.error(error)
+      } else {
+        Events.postEvent(this, {
```
sarahscott: optional analytics ✨

alloy:
This looks great 👍 

However, the ‘TemporaryAPIModule’ is only meant for networking, i.e. ‘API’ refers to Gravity. It is meant to be temporary in the sense that it would get removed once we have mutations in MP.

So can you add a dedicated native module for it instead? E.g. ‘ARWorksForYouModule’

sarahscott:
Ah yeah that  makes sense. Will do. How temporary is the `ARTemporaryAPIModule`? Are there plans to make metaphysics mutable soon?

alloy:
No planned timeframe, just whenever we get to it.

sarahscott:
Alright, 🍏 and ready for review!

alloy:
Looks good! 👌

PR #468, opened 2017-03-24
Fixes display of artworks in non-expanding grids in artwork rails

ashfurrow:
Related to https://github.com/artsy/eigen/issues/2237


ashfurrow:
Not sure why Danger is failing:

> The command "yarn danger" exited with 1.

PR #479, opened 2017-04-10
Remove flow definition for jest

orta:
This flow type definition got missed on the TypeScript port, we're now using `@types` instead 👍


sarahscott:
👍

PR #478, opened 2017-04-07
Update Dangerfile to use master

orta:
I'll see if I can get Danger to raise the error, then fix the error.

Not certain if using yarn this way will compile the TS for Danger, so it might not work at all.

PR #482, opened 2017-04-12
[WFU] Added special notification capabilities

sarahscott:
- [x] Added special notification based on artist ID passed in from native app
- [x] Tests

### DIFF
--- a/Pod/Classes/ViewControllers/ARWorksForYouComponentViewController.m
+++ b/Pod/Classes/ViewControllers/ARWorksForYouComponentViewController.m
@@ -9,9 +9,14 @@
 
 - (instancetype)initWithSelectedArtist:(NSString *)artistID emission:(AREmission *)emission;
 {
+    NSDictionary *initialProperties = @{};
+    if (artistID) {
+        initialProperties = @{ @"selectedArtist": artistID };
+    }
+
     if ((self = [super initWithEmission:emission
                             moduleName:@"WorksForYou"
-                     initialProperties:@{ @"selectedArtist": artistID }])) {
+                     initialProperties:initialProperties])) {
         _selectedArtist = artistID;
     }
     return self;
--- a/Pod/Classes/ViewControllers/ARWorksForYouComponentViewController.m
+++ b/Pod/Classes/ViewControllers/ARWorksForYouComponentViewController.m
@@ -9,14 +9,9 @@
 
 - (instancetype)initWithSelectedArtist:(NSString *)artistID emission:(AREmission *)emission;
 {
-    NSDictionary *initialProperties = @{};
-    if (artistID) {
-        initialProperties = @{ @"selectedArtist": artistID };
-    }
-
     if ((self = [super initWithEmission:emission
                             moduleName:@"WorksForYou"
-                     initialProperties:initialProperties])) {
+                      initialProperties:artistID ? @{ @"selectedArtist": artistID } : nil])) {
         _selectedArtist = artistID;
     }
     return self;
--- a/Pod/Classes/ViewControllers/ARWorksForYouComponentViewController.m
+++ b/Pod/Classes/ViewControllers/ARWorksForYouComponentViewController.m
@@ -2,9 +2,19 @@
 
 @implementation ARWorksForYouComponentViewController
 
-- (instancetype)initWithEmission:(AREmission *)emission;
+- (instancetype)initWithSelectedArtist:(NSString *)artistID;
 {
-    return [super initWithEmission:emission moduleName:@"WorksForYou" initialProperties:nil];
+    return [self initWithSelectedArtist:artistID emission:nil];
+}
+
+- (instancetype)initWithSelectedArtist:(NSString *)artistID emission:(AREmission *)emission;
+{
+    if ((self = [super initWithEmission:emission
+                            moduleName:@"WorksForYou"
+                     initialProperties:@{ @"selectedArtist": artistID }])) {
+        _selectedArtist = artistID;
+    }
+    return self;
 }
 
 @end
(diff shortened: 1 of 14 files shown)


src/lib/app_registry.tsx
```diff
@@ -76,8 +76,12 @@ class Home extends RootContainer {
 class WorksForYou extends RootContainer {
   constructor(props) {
     super(props)
+
+    // This really has to be a boolean for Relay to understand it
+    const showSpecialNotification = props.selectedArtist && props.selectedArtist !== ""
```
alloy: You can use the ol’ double-bang to ensure it’s a boolean:

```
$ node
> !!""
false
```

Which means you can probably make Relay happy like so:

```js
this.route = new Routes.WorksForYou({
  selectedArtist: props.selectedArtist,
  showSpecialNotification: !!props.selectedArtist,
})
```

alloy: On second thought, you don’t even really need to set the `showSpecialNotification` flag from here, you can instead use Relay’s `prepareVariables` to derive it from `selectedArtist`: https://facebook.github.io/relay/docs/api-reference-relay-container.html#preparevariables

Something like the following, I think?

```js
  prepareVariables:({ selectedArtist }) => {
    return {
      selectedArtist,
      showSpecialNotification: !!selectedArtist,
    };
  },
```

src/lib/components/works_for_you/notification.tsx
```diff
@@ -13,7 +13,9 @@ interface Props extends RelayProps {}
 
 export class Notification extends React.Component<Props, any> {
   handleArtistTap() {
-    const artistHref = this.props.notification.artworks[0].artists[0].href
+    // Special notifications will pass down an artistHref. Otherwise, grab it from the artworks
```
alloy: What does “special” refer to in this comment?

src/lib/containers/works_for_you.tsx
```diff
@@ -203,33 +224,75 @@ const styles = StyleSheet.create<Styles>({
...
+              }
             }
           }
         }
+
+        selectedArtist: artist(id: $selectedArtist)
+                          @include(if: $showSpecialNotification) {
+          name
+          image {
+            resized(height: 80, width: 80) {
```
alloy: Does this size take retina screens into account?

src/lib/containers/works_for_you.tsx
```diff
@@ -203,33 +224,75 @@ const styles = StyleSheet.create<Styles>({
...
+            image {
+              aspect_ratio
+              url(version: "large")
+            }
+            partner {
+              name
+            }
+            href
+          },
+          href
```
alloy: Would probably read better to move this to the top next to `name`.

src/lib/relay/routes.tsx
```diff
@@ -54,15 +54,20 @@ class Gene extends Relay.Route {
 
 class WorksForYou extends Relay.Route {
   static queries = {
-    me: (component, params) => Relay.QL`
+    viewer: (component, params) => Relay.QL`
       query {
-        me {
-          ${component.getFragment("me", params)}
+        viewer {
+          ${(component.getFragment("viewer", params))}
```
alloy: I think this would remove the need to use the `viewer` namespace at all in the rest of the code:

```graphql
query {
  viewer {
    me {
      ${component.getFragment("me", params)}
    }
  }
}
```

sarahscott: I could do that, but then the `artworks` would have to be `me.selectedArtist` or something which isn't entirely true since the `artist` doesn't reside under `me` in graphql.

sarahscott: I actually don't even know if that would work since you can only have one fragment in a root and I need more than just the `me` fragment.

src/lib/containers/works_for_you.tsx
sarahscott: Not yet - I could use some recommendations for that in RN.

src/lib/components/works_for_you/notification.tsx
sarahscott: It just refers to this special-case situation where we make a notification from an artist's recently published works instead of through our usual notifications api.

sarahscott: I suppose if i decoupled the header from the notification, I could make a new `specialNotification` component that uses the notification header and the artworks grid and that might be more in line with the React way of doing things. But it would also be slightly more redundant.

src/lib/containers/works_for_you.tsx
alloy: Because of 🐔 and 🥚 problems, I moved the resize/crop URL logic from MP into Emission in our image view component. What it does is wait to be laid out so that it knows its actual size, multiply by screen pixel ratio, and then fetch the actual resized/cropped image.

src/lib/relay/routes.tsx
alloy: Hmm, even if just for experimentation purposes and have a better understanding, I’d really like to pair on this with you for a little bit.

Pod/Classes/ViewControllers/ARWorksForYouComponentViewController.m
```diff
@@ -2,9 +2,24 @@
...
+- (instancetype)initWithSelectedArtist:(NSString *)artistID;
 {
-    return [super initWithEmission:emission moduleName:@"WorksForYou" initialProperties:nil];
+    return [self initWithSelectedArtist:artistID emission:nil];
+}
+
+- (instancetype)initWithSelectedArtist:(NSString *)artistID emission:(AREmission *)emission;
+{
+    NSDictionary *initialProperties = @{};
+    if (artistID) {
```
sarahscott: fixing this

alloy:
I suppose Danger is failing because you haven’t rebased and pulled in this fix #478. Tests are green etc, so works for me 👍

PR #491, opened 2017-04-24
[WIP][Analytics] first draft

mennenia:
Would love your opinions! @alloy @sarahscott @l2succes @mzikherman @xtina-starr @yuki24 @broskoski @orta

(Just tagging the whole team here 😄 )

FYI
- Segment hasn't been added yet, hence some commented code in the analytics file
- DSL is very up for grabs (also note that currently not being distinguished between following and unfollowing, that can easily be done in there though)
- Old analytics code is still in the follow method :D
- Decorator could do with a few more checks (Autobind mentions loops in IE11, and there are some measures they take to prevent that from happening that I haven't ported)

### DIFF
--- a/src/analytics/analytics.tsx
+++ b/src/analytics/analytics.tsx
@@ -1,0 +1,17 @@
+// import segment
+import * as DSL from "./dsl"
+
+export function sendIntent(analyticsConstant: string, contextProps: any) {
+  // send it
+  let properties = DSL.dsl[analyticsConstant].intent(contextProps)
+  console.log(properties)
+  // Segment.track(analyticsConstant, properties)
+}
+
+export function sendSuccess(analyticsConstant: string, contextProps: any) {
+  // send it
+}
+
+export function sendFail(analyticsConstant: string, contextProps: any, error: Error) {
+  // send it
+}
--- a/src/analytics/dsl.tsx
+++ b/src/analytics/dsl.tsx
@@ -1,0 +1,12 @@
+export const dsl = {
+  "Follow Artist": {
+    intent: (object) => ({ artist_id: object.artist.id }),
+    success: (object) => ({ artist_id: object.artist.id }),
+    fail: (object) => ({ artist_id: object.artist.id }),
+  },
+  "Unfollow Artist": {
+    intent: (object) => ({ artist_id: object.artist.id }),
+    success: (object) => ({ artist_id: object.artist.id }),
+    fail: (object) => ({ artist_id: object.artist.id }),
+  },
+}
--- a/src/lib/components/artist/header.tsx
+++ b/src/lib/components/artist/header.tsx
@@ -1,11 +1,13 @@
 import * as React from "react"
 import * as Relay from "react-relay"
+import eventHandler from "../../helpers/decorators"
 
 import {
   Dimensions,
   NativeModules,
   StyleSheet,
   TextStyle,
+  Touchable,
   View,
   ViewStyle,
 } from "react-native"
@@ -52,29 +54,35 @@ class Header extends React.Component<HeaderProps, State> {
     })
   }
 
-  handleFollowChange = () => {
+  @eventHandler("Follow Artist")
+  handleFollowChange() { // this now needs to be a normal method, not a property
     const newFollowersCount = this.state.following ? (this.state.followersCount - 1) : (this.state.followersCount + 1)
-    ARTemporaryAPIModule.setFollowArtistStatus(!this.state.following, this.props.artist._id, (error, following) => {
-      if (error) {
-        console.error(error)
-      } else {
-        Events.postEvent(this, {
-          name: following ? "Follow artist" : "Unfollow artist",
-          artist_id: this.props.artist._id,
-          artist_slug: this.props.artist.id,
-          // TODO At some point, this component might be on other screens.
-          source_screen: "artist page",
-        })
-      }
-      this.setState({ following, followersCount: newFollowersCount })
+    return new Promise((resolve, reject) => {
+      ARTemporaryAPIModule.setFollowArtistStatus(!this.state.following, this.props.artist._id, (error, following) => {
+        if (error) {
+          console.error(error)
+          reject(error)
+        } else {
+          Events.postEvent(this, {
+            name: following ? "Follow artist" : "Unfollow artist",
+            artist_id: this.props.artist._id,
+            artist_slug: this.props.artist.id,
+            // TODO At some point, this component might be on other screens.
+            source_screen: "artist page",
+          })
+          resolve()
+        }
+        this.setState({ following, followersCount: newFollowersCount })
+      })
+      this.setState({ following: !this.state.following, followersCount: newFollowersCount })
     })
-    this.setState({ following: !this.state.following, followersCount: newFollowersCount })
   }
 
   render() {
     const artist = this.props.artist
+
     return (
-      <View style={{paddingTop: 20}}>
+      <View style={{ paddingTop: 20 }}>
         <Headline style={[styles.base, styles.headline]}>
           {artist.name}
         </Headline>
@@ -88,10 +96,10 @@ class Header extends React.Component<HeaderProps, State> {
   renderFollowButton() {
     if (this.state.following !== null) {
       return (
-        <View style={styles.followButton}>
-            <InvertedButton text={this.state.following ? "Following" : "Follow"}
-                            selected={this.state.following}
-                            onPress={this.handleFollowChange} />
+        <View style={styles.followButton} >
+          <InvertedButton text={this.state.following ? "Following" : "Follow"}
+            selected={this.state.following}
+            onPress={this.handleFollowChange} />
         </View>
       )
     }
--- a/src/lib/helpers/decorators.tsx
+++ b/src/lib/helpers/decorators.tsx
@@ -1,0 +1,53 @@
+import * as Analytics from "../../analytics/analytics"
+
+// This heavily uses ideas and approaches from @autobind
+// https://github.com/andreypopp/autobind-decorator
+export default function eventHandler(analyticsConstant: string) {
+  // TODO: disable tslint arrow function rule
+  return function bindMethod(target: any, propertyKey: string, descriptor: PropertyDescriptor) {
+    let originalMethod = descriptor.value
+
+    if (typeof originalMethod !== "function") {
+      throw new Error(`@eventHandler can only be applied to methods not: ${typeof originalMethod}`)
+    }
+
+    // normally in decorators, you overwrite the descriptor's value (= function), and return as is
+    // because we're binding, we're creating a whole new descriptor
+    let newDescriptor = {
+      configurable: true,
+      get() {
+        if (this === target.prototype || this.hasOwnProperty(propertyKey)) {
+          return originalMethod
+        }
+
+        let boundFn = originalMethod.bind(this)
+
+        let analyticsWrappedMethod
+        if (analyticsConstant) {
+          let contextualProps = this.props
+          Analytics.sendIntent(analyticsConstant, contextualProps)
+          analyticsWrappedMethod = function (...args: any[]) {
+            const resultPromise = boundFn.apply(this, args)
+            resultPromise
+              .then(() => {
+                Analytics.sendSuccess(analyticsConstant, contextualProps)
+              })
+              .catch(error => {
+                Analytics.sendFail(analyticsConstant, contextualProps, error)
+                console.error("FAILED", error)
+              })
+            return resultPromise
+          }.bind(this)
+        }
+
+        Object.defineProperty(this, propertyKey, {
+          value: analyticsConstant ? analyticsWrappedMethod : boundFn,
+          configurable: true,
+          writable: true,
+        })
+        return analyticsConstant ? analyticsWrappedMethod : boundFn
+      },
+    }
+    return newDescriptor
+  }
+}
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,7 +8,8 @@
         "allowJs": true,
         "pretty": true,
         "moduleResolution": "node",
-        "strictNullChecks": false
+        "strictNullChecks": false,
+        "experimentalDecorators": true
     },
     "include": [
         "typings/index.d.ts",


src/lib/components/artist/header.tsx
```diff
@@ -1,11 +1,13 @@
 import * as React from "react"
 import * as Relay from "react-relay"
+import eventHandler from "../../helpers/decorators"
 
 import {
   Dimensions,
   NativeModules,
   StyleSheet,
   TextStyle,
+  Touchable,
```
mennenia: This will be removed

src/lib/components/artist/header.tsx
```diff
@@ -1,11 +1,13 @@
 import * as React from "react"
 import * as Relay from "react-relay"
+import eventHandler from "../../helpers/decorators"
```
mennenia: does it make sense to put it there?

src/lib/components/artist/header.tsx
```diff
@@ -52,29 +54,35 @@ class Header extends React.Component<HeaderProps, State> {
...
-        Events.postEvent(this, {
-          name: following ? "Follow artist" : "Unfollow artist",
-          artist_id: this.props.artist._id,
-          artist_slug: this.props.artist.id,
-          // TODO At some point, this component might be on other screens.
-          source_screen: "artist page",
-        })
-      }
-      this.setState({ following, followersCount: newFollowersCount })
+    return new Promise((resolve, reject) => {
```
mennenia: Still need to implement a way to enforce promises in any event handlers. Any thoughts?

tsconfig.json
```diff
@@ -8,7 +8,8 @@
         "allowJs": true,
         "pretty": true,
         "moduleResolution": "node",
-        "strictNullChecks": false
+        "strictNullChecks": false,
+        "experimentalDecorators": true
```
mennenia: Introducing decorators! 💃

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Running your Dangerfile has Failed</td>
    </tr>
  </tbody>
</table>




## Danger has errored

Error: TypeError

```sh
TypeError: Cannot read property 'length' of null
    at reactComponentForPath (/home/travis/build/artsy/dangerfile.ts:87:12)
    at Array.map (native)
    at Object.<anonymous> (/home/travis/build/artsy/dangerfile.ts:94:52)
    at Runtime._execModule (/home/travis/build/artsy/emission/node_modules/jest-runtime/build/index.js:447:13)
    at Runtime.requireModule (/home/travis/build/artsy/emission/node_modules/jest-runtime/build/index.js:295:14)
    at /home/travis/build/artsy/emission/node_modules/danger/source/runner/DangerfileRunner.ts:92:13
    at ensureCleanDangerfile (/home/travis/build/artsy/emission/node_modules/danger/source/runner/DangerfileRunner.ts:129:3)
    at Object.<anonymous> (/home/travis/build/artsy/emission/node_modules/danger/source/runner/DangerfileRunner.ts:91:3)
    at step (/home/travis/build/artsy/emission/node_modules/danger/distribution/runner/DangerfileRunner.js:32:23)
    at Object.next (/home/travis/build/artsy/emission/node_modules/danger/distribution/runner/DangerfileRunner.js:13:53)
```


<p align="right">
</p>

src/lib/components/artist/header.tsx
```diff
@@ -52,29 +54,35 @@ class Header extends React.Component<HeaderProps, State> {
     })
   }
 
-  handleFollowChange = () => {
+  @eventHandler("Follow Artist")
```
yuki24: Decorators! 👍

src/lib/components/artist/header.tsx
alloy: Can you use TS to enforce a signature on decorators? E.g. you’d require a function with the following signature:

```ts
() => Promise<any>
```

src/lib/helpers/decorators.tsx
```diff
@@ -0,0 +1,53 @@
...
+            const resultPromise = boundFn.apply(this, args)
+            resultPromise
+              .then(() => {
+                Analytics.sendSuccess(analyticsConstant, contextualProps)
+              })
+              .catch(error => {
+                Analytics.sendFail(analyticsConstant, contextualProps, error)
+                console.error("FAILED", error)
+              })
+            return resultPromise
```
alloy: Definitely add a test that ensures we don’t return the promise resulting from the then/catch calls just above and test that us introducing a catch call doesn’t omit errors from the returned `resultPromise`.

src/analytics/dsl.tsx
```diff
@@ -0,0 +1,12 @@
+export const dsl = {
+  "Follow Artist": {
+    intent: (object) => ({ artist_id: object.artist.id }),
+    success: (object) => ({ artist_id: object.artist.id }),
+    fail: (object) => ({ artist_id: object.artist.id }),
```
alloy: We should probably also yield the `error` object to the block.

src/analytics/dsl.tsx
```diff
@@ -0,0 +1,12 @@
...
+  },
+  "Unfollow Artist": {
+    intent: (object) => ({ artist_id: object.artist.id }),
+    success: (object) => ({ artist_id: object.artist.id }),
+    fail: (object) => ({ artist_id: object.artist.id }),
+  },
+}
```
alloy: In retrospect I wonder if this large object that mimics ARAnalytics is the best idea, because that large object sometimes becomes a bit of a hassle to maintain and is really only done because otherwise objc makes DSLs cumbersome. Additionally we probably want to add multiple analytics files at some point, so being able to call those functions from anywhere in the project is more flexible.

```js
import * as analytics from "./dsl"

analytics.on("Follow Artist", {
  intent: (object) => ({ artist_id: object.artist.id }),
  success: (object) => ({ artist_id: object.artist.id }),
  fail: (object, error) => ({ artist_id: object.artist.id }),
})
```

alloy: Or shorter DSL usage:

```js
import { on } from "./dsl"

on("Follow Artist", {
  intent: (object) => ({ artist_id: object.artist.id }),
  success: (object) => ({ artist_id: object.artist.id }),
  fail: (object, error) => ({ artist_id: object.artist.id }),
})
```

alloy:
> Decorator could do with a few more checks (Autobind mentions loops in IE11, and there are some measures they take to prevent that from happening that I haven't ported)

I didn’t realise binding might be such a complex task. If you think it makes sense, we could also just rely on the [autobind-decorator](https://github.com/andreypopp/autobind-decorator) implementation for the binding part?

alloy:
Great stuff @mennenia, well done 👏

mennenia:
> I didn’t realise binding might be such a complex task. If you think it makes sense, we could also just rely on the autobind-decorator implementation for the binding part?

@alloy , I wanted to, it was a bit of a toss up. Because the analytics is kind of interwoven with the creating of the function / binding, I wasn't sure to either do it is like it is now, or autobind, get the function back, and then modify again

I decided to copy it to test it for this wip. Happy to move over to just using the lib :).

src/analytics/dsl.tsx
mennenia: definitely!

src/lib/helpers/decorators.tsx
mennenia: :+1:

src/lib/components/artist/header.tsx
mennenia: will try!

src/analytics/dsl.tsx
mennenia: Ohh I like this! (Also, such snazzy syntax)

I don't think I quite understand your idea of the multiple files though. There's one analytics file that actually sends the events and these are the methods that the decorator knows to access. E.g., send an event that this thing succeeded.

So, if we were to have multiple files for the different components etc, how would this context specific code end up in the analytics file?

alloy:
@mennenia Gotcha, makes sense 👍

src/analytics/dsl.tsx
alloy: The `analytics` module should keep a singleton of all registered handlers:

```js
const handlers = new Map()

export function on(eventName, eventHandler) {
  handlers.set(eventName, eventHandler)
}

export function sendIntent(analyticsConstant: string, contextProps: any) {
  // send it
  let properties = handlers.get(analyticsConstant).intent(contextProps)
  console.log(properties)
  // Segment.track(analyticsConstant, properties)
}
```

alloy: We should also think about how we can best add typings to those props. Do we want to reference specific components for these analytics event hooks? If so, we could maybe do something like:

```js
import { for, on } from "./analytics"

import Artist from "../containers/artist"

for(Artist, artist => {
  artist.on("Follow Artist", {
    intent: (object) => ({ artist_id: object.artist.id }),
    success: (object) => ({ artist_id: object.artist.id }),
    fail: (object, error) => ({ artist_id: object.artist.id }),
  })
})
```

That way we can use generics to get the props interface from the component and type the `object` yielded to each function as the same props type.

alloy: I guess these are indeed kinda bound to specific components because we also use their props 🤔 If so, the above example could work.

src/lib/components/artist/header.tsx
l2succes: I just read through the [decorators typescript docs](https://www.typescriptlang.org/docs/handbook/decorators.html). There isn't a way to enforce a return type on method decorator. My initial thought was using the reflection API in order to get the return type of the method at runtime

However, based on the Promise/A+ spec, we don't need to enforce the method to return a `Promise`. just use `Promise.resolve(valueOrPromiseItDoesn'tMatter)`, you can read more about the logic behind that [here](http://stackoverflow.com/a/27746324)

Useful Links:
- http://blog.wolksoftware.com/decorators-reflection-javascript-typescript
- http://stackoverflow.com/a/27746324

src/lib/helpers/decorators.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+        let boundFn = originalMethod.bind(this)
+
+        let analyticsWrappedMethod
+        if (analyticsConstant) {
+          let contextualProps = this.props
+          Analytics.sendIntent(analyticsConstant, contextualProps)
+          analyticsWrappedMethod = function (...args: any[]) {
+            const resultPromise = boundFn.apply(this, args)
+            resultPromise
```
l2succes: As mentioned above, you can do

`Promise.resolve(boundFn.apply(this, args)).then(...)` instead, it'll safely ensure that `.then` gets called regardless of the return type of your boundFunction

l2succes:
Nice work! ❤️ the decorator API usage 😊

src/analytics/analytics.tsx
```diff
@@ -0,0 +1,17 @@
+// import segment
+import * as DSL from "./dsl"
+
+export function sendIntent(analyticsConstant: string, contextProps: any) {
+  // send it
+  let properties = DSL.dsl[analyticsConstant].intent(contextProps)
```
l2succes: As @alloy mentioned earlier, if we store handlers in a map, we can check the constant was registered properly.
```
const handlers = new Map()
if(handlers.has(analyticsConstant)) {
  handlers.get(analyticsConstant).intent(contextProps)
}
```

src/analytics/dsl.tsx
l2succes: You could add an interface to specify the handlers signature
```
const dsl: [string: AnalyticsHandler] = {...}

interface AnalyticsHandler {
  intent: (object: any) => any,
  success: (object: any) => any,
  error: (object: any, error: Error) => any
}
```
or if we end up using the `Analytics.on` API style
```
class Analytics {
  on(contant: string, handler: AnalyticsHandler) {...}
}
```

src/lib/helpers/decorators.tsx
```diff
@@ -0,0 +1,53 @@
...
+              .then(() => {
+                Analytics.sendSuccess(analyticsConstant, contextualProps)
```
l2succes: Not necessary but assuming the `Analytics.sendSuccess` method returns a value or promise. We should return a promise here

`            resultPromise
               .then(() => {
                 return Analytics.sendSuccess(analyticsConstant, contextualProps)
               })
`
that way we can chain promises or check the value in tests.

sarahscott:
This is really exciting and will make analytics so much lighter and nicer to implement! Thank you @mennenia 🥇 !

src/lib/components/artist/header.tsx
alloy: @l2succes I think you misunderstood, we need these functions to return a promise so that they can signal whether an event succeeded or failed.

src/analytics/analytics.tsx
mennenia: oh that's cool!

src/lib/helpers/decorators.tsx
```diff
@@ -0,0 +1,53 @@
+import * as Analytics from "../../analytics/analytics"
+
+// This heavily uses ideas and approaches from @autobind
+// https://github.com/andreypopp/autobind-decorator
+export default function eventHandler(analyticsConstant: string) {
+  // TODO: disable tslint arrow function rule
+  return function bindMethod(target: any, propertyKey: string, descriptor: PropertyDescriptor) {
+    let originalMethod = descriptor.value
```
alloy: @mennenia The original method is actually also in `target`, so maybe the way to set the return promise type expectation you need to do something like:

```ts
function bindMethod(target: () => Promise<any>, propertyKey: string, descriptor: PropertyDescriptor) {
  ...
}
```

src/lib/components/artist/header.tsx
l2succes: Ah!! Got it, so it makes sense to always return a promise then. 
We can still add a runtime check to make sure the return type is `Promise`

alloy: True, we should probably throw an error if the returned object isn’t a Promise.

mennenia:
Thanks for all the comments, tips, and suggestions everyone! I will have a further go and try and implement some of them (might poke you again when I do haha).

orta:
Can this be closed now that #693 is getting there?

PR #500, opened 2017-05-03
[doc] More elaborate debugging guide

alloy:
It became painfully clear in [this Slack chat](https://artsy.slack.com/archives/C02CAB1KW/p1493776090098445) that being able to debug e.g. GraphQL queries that Relay emits was hard to do for people not contributing to this repo regularly. Hopefully this should remedy that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,12 +48,7 @@ There is a comprehensive document covering [our setup here](docs/vscode.md).
 
 ### Debugging
 
-You can use React Native debugger which is a standalone app to inspect views as well as the standard chrome debugging tools.
-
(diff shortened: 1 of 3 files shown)


dblock:
Thanks for this. I was close, missing the whole shake story.

It would be good to be able to debug eigen itself like this, having to rely on potentially a different version of emission seems problematic.

alloy:
That would require leaving all debug tooling and symbols in the Emission builds, which I’d prefer we don’t. We tag releases, so it shouldn’t be hard to test the right version.

PR #503, opened 2017-05-03
Fixed issues found when upgrading to TypeScript 2.3.

DanielRosenwasser:
This meant I needed to upgrade to `@types/react-native@0.43.9`, but it looks like you've already moved onto `0.43.7` anyhow.


src/lib/containers/home.tsx
```diff
@@ -83,7 +83,7 @@ export class Home extends React.Component<Props, State> {
   render() {
     return (
       <ListView dataSource={this.state.dataSource}
-                renderScrollComponent={(props: ScrollViewStatic) => {
+                renderScrollComponent={(props: ScrollViewProps) => {
```
DanielRosenwasser: I believe I can remove the annotation here if you'd like.

orta:
Awesome, thanks! 

This looks good to me, it even works with 2.2 which is what we're running with now in CI. 

We don't expect tests for OSS contributors to pass yet, so this is 👍

src/lib/components/text/headline.tsx
```diff
@@ -1,11 +1,11 @@
 import * as React from "react"
-import { StyleSheet, Text, TextProperties, TextStatic } from "react-native"
+import { StyleSheet, Text, TextProperties } from "react-native"
 
 export default class Headline extends React.Component<TextProperties, any> {
   render() {
     const content = (this.props.children || "") as string
     return (
-      <Text {...this.props as TextStatic} style={[styles.default, this.props.style, styles.required]} >
```
alloy: YES! I’m so glad to get rid of these things that felt kinda wrong.

alloy:
Thanks @DanielRosenwasser! 🙏

PR #510, opened 2017-05-05
[Consignments] Initial work on adding a Me Profile page

orta:
Currently a bit blocked on getting the initial relay query off the floor. Going to call it for this second, but wanted to at least have my work out there.


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>No CHANGELOG added.</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/__tests__/app_registry-tests.tsx`
- `src/lib/containers/__stories__/__tests__/index-tests.ts`
- `src/lib/relay/__tests__/routes-tests.tsx`
- `src/lib/containers/__stories__/__tests__/my_profile.story-tests.tsx`

If these files are supposed to not exist, please update your PR body to include "Skip New Tests".</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>Please assign someone to merge this PR, and optionally include people who should review.</td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

orta:
Assigning Maxim instead for more fun, @mennenia ask me anything you don't get

data/schema.graphql
```diff
@@ -194,32 +194,32 @@ type ArtistCarousel {
 
 type ArtistCounts {
   artworks(
-    # Returns a `String` when format is specified. e.g.`"0,0.0000"`
+    # Returns a `String` when format is specified. e.g.`'0,0.0000''`
```
mennenia: Why are the quotes different here? :o

data/schema.graphql
```diff
@@ -1264,6 +1287,15 @@ enum BuyerOutcomeTypes {
...
 type BuyersPremium {
+  # A globally unique ID.
+  __id: ID!
+
+  # A type-specific ID.
+  id: String!
+
+  # A type-specific Gravity Mongo Document ID.
+  _id: String!
+
```
mennenia: so many ids lol x_x

What are the types here? it's an id specific to a type of... category? :o

data/schema.graphql
```diff
@@ -2499,6 +2560,7 @@ type OrderedSet {
 
 type organizer {
   profile_id: ID
+  profile: Profile
```
mennenia: is "Profile" here the same as what we were talking about the other week? Metadata to an entity?

src/lib/app_registry.tsx
```diff
@@ -91,7 +91,16 @@ class WorksForYou extends RootContainer {
   }
 }
 
+class MyProfile extends RootContainer {
```
mennenia: So, here we're talking about "my account" as it were. Is there a specific reason we're calling it profile? I know it's quite obvious here, but in Eigen for example we have so many references to profiles (in the context of metainformation, if I understood correctly). I'm wondering whether it might save us confusion in the future if we call it account or ... dno, something? 😅

src/lib/containers/__tests__/my_profile-tests.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react"
+import * as renderer from "react-test-renderer"
+
+import { MyProfile } from "../my_profile"
+
+it("looks like expected", () => {
+  const props = {
+      me: {
+        name: "Danger McShane",
```
mennenia: 😄

mennenia:
Looks good! I added some Qs I have, but overall, good to merge from what I gather :)

data/schema.graphql
orta: Been converting all of the source code to use [prettier](https://github.com/prettier/prettier/), in doing so we switched from `'` to `"` - think I missed a spot and it got caught as I was working.

data/schema.graphql
orta: lols yeah, we have 3 types of IDs, let's think of it in terms of an Artist:

* `id` - `banksy` the usual slug-likeID
* `_id` - The mongo ID ( every JSON object in Mongo has a UUID)
* `__id` - I believe in _some_ models we had already used `_id` to _not_ mean the mongoID and so we had to go a step further and have a _really_ UUID (nearly always the mongoID, but if we have multiple DBs then in theory we can have clashes)

src/lib/app_registry.tsx
orta: Not really anything special in the naming, just wanted a name so I could move on. `MyAccount` is probably a better one IMO. This can be a blocker on the PR.

data/schema.graphql
orta: Yep ( [this object](https://github.com/artsy/metaphysics/blob/master/schema/profile.js) ) which gets attached to things

src/lib/app_registry.tsx
mennenia: In that case, yeah, I think it'd be a good change to make now while it's still fairly easy :D

data/schema.graphql
mennenia: Ahh ok :) thanks!

data/schema.graphql
mennenia: 👍

data/schema.graphql
alloy: It’s the `id` field that was already in use by us in a non-globally unique way, and as `_id` also already existed we arrived at `__id`.

For more info see https://facebook.github.io/relay/graphql/objectidentification.htm

data/schema.graphql
alloy: Oh, is TypeScript officially supported now in prettier?

orta: sooooonnn, these are the derived graphql exports from metaphysics

mennenia:
Aweeesome, thanks @orta !! Will merge on 🍏

PR #513, opened 2017-05-06
[Dev] Adds husky for linting ts files, and ensuring the entire thing builds

orta:
To avoid accidentally pushing builds that fail, I've added:

* TSLint to run on creating a new commit
* A project build on push

The first will only really catch issues inside files which have changed, so it'll be faster. This won't highlight issues in related files, so when you do a push it will run the entire project build (took me 4 seconds before this commit was pushed.)

Interesting side effect is that this forces you into using the right node version, I had to install 6.9 in order for the linters etc to pass.


alloy:
I’m ok with this, just will have to figure out why `lint-staged` hasn’t been working for me in MP.

PR #522, opened 2017-05-08
[Dev] Adds the ability to log out to the app

orta:
Exposes the `AuthenticationManager` to the `ARRootViewController` so that it can call logout from the same context.


DanielRosenwasser:
Thanks so much for this change, we really appreciate it!! 😄

alloy:
I think the CI/PR failure is because of the token being revoked after Travis’ leakage.

Going to merge this as is. Thanks for jumping on this @orta 🙏

PR #523, opened 2017-05-08
[Dev] Adds an API call to ensure authentication is valid on launch

orta:
Sibling PR to #522 - it's an alternative, or it can enhance. 

When the debugging app launches, it makes a call for the user's profile. If that API request passes because authorisation is 👍 then it will let you into the app. If it doesn't, then it will present the setup interface.

### DIFF
--- a/Example/Emission/AppDelegate.m
+++ b/Example/Emission/AppDelegate.m
@@ -70,7 +70,15 @@ randomBOOL(void)
   if ([auth isAuthenticated]) {
     NSString *accessToken = [auth token];
     if (accessToken) {
-      [self setupEmissionWithUserID:[auth userID] accessToken:accessToken keychainService:service];
+
+      [auth validateStoredCredentialsPassing:^{
+        [self setupEmissionWithUserID:[auth userID] accessToken:accessToken keychainService:service];
+      } failing:^{
+        [auth presentAuthenticationPromptOnViewController:rootVC completion:^{
+          NSLog(@"Logged in successfully :)");
+          [self setupEmissionWithUserID:[auth userID] accessToken:[auth token] keychainService:service];
+        }];
+      }];
     }
   } else {
     [auth presentAuthenticationPromptOnViewController:rootVC completion:^{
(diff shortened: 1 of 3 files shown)


Example/Emission/AppDelegate.m
```diff
@@ -70,7 +70,15 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
     if (accessToken) {
-      [self setupEmissionWithUserID:[auth userID] accessToken:accessToken keychainService:service];
+
+      [auth validateStoredCredentialsPassing:^{
+        [self setupEmissionWithUserID:[auth userID] accessToken:accessToken keychainService:service];
+      } failing:^{
+        [auth presentAuthenticationPromptOnViewController:rootVC completion:^{
+          NSLog(@"Logged in successfully :)");
+          [self setupEmissionWithUserID:[auth userID] accessToken:[auth token] keychainService:service];
+        }];
```
orta: could be DRYer (its the same as line 84) - but it's not really enough to worry me at this scale

Example/Emission/AuthenticationManager.m
```diff
@@ -104,6 +108,13 @@ - (void)authenticateWithEmail:(NSString *)email password:(NSString *)password ke
             NSParameterAssert(accessToken);
 
             NSError *error = nil;
+            if([SAMKeychain passwordForService:service account:userID]) {
+              [SAMKeychain deletePasswordForService:service account:userID error:&error];
+              if (error) {
+                NSLog(@"%@", error);
+              }
+            }
```
orta: hint of paranoia, but we'd get error reports this way

Example/Emission/AuthenticationManager.m
```diff
@@ -119,5 +130,20 @@ - (void)authenticateWithEmail:(NSString *)email password:(NSString *)password ke
...
+                                                               clientSecret:@"3a33d2085cbd1176153f99781bbce7c6"];
+  auth.router.authToken = [[ArtsyToken alloc] initWithToken:self.token expirationDate:[NSDate distantFuture]];
+
+  NSURLRequest *request = [auth.router requestForUserDetails];
+  [networkOperator JSONTaskWithRequest:request success:^(NSURLRequest *_, NSHTTPURLResponse *__, id JSON) {
+    pass();
+  } failure:^(NSURLRequest *_, NSHTTPURLResponse *response, NSError *error, id JSON) {
+    fail();
+  }];
+}
```
orta: pulls out some of the authentication objects to make it use the current setup for validation

orta:
Alright, calling this as not needed 👍

PR #530, opened 2017-05-11
[Home] Rails without artworks will not be rendered

### LINKED ISSUE #516: empty auction and fair rails
image

The auction is live and has lots, the fair on the otherhand is not live to the public but has been published with works so not entirely sure how we want to handle that case. The native fair view isn't getting any data yet but on /force we're able to support the full microsite. Sounds like there's probably some rules around when a fair is published vs it's start date that need to be addressed.

sarahscott:
We don't have a specific issue for this, but it provides a safeguard just in case we ever have rails without artworks.

This can be replicated with the `test` rail in this pr (will be removed before merge), but the basic idea is that the spinner will show until it is confirmed that the `results` array does not exist or is empty.

simulator screen shot may 12 2017 11 39 10

Skip New Tests

Fixes #516


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>Missing Test Files:

- `src/lib/components/home/artwork_rails/__tests__/artwork_rail-tests.tsx`

If these files are supposed to not exist, please update your PR body to include "Skip New Tests".</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>Could not find stories using these components:

- `ArtworkRail`

</td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

alloy:
In Slack you mentioned that you weren’t sure if this lead to the best result, something about the spinner being useless. Can you elaborate/show a vid of what you mean? Or is this already fixed?

src/lib/components/home/artwork_rails/artwork_rail.tsx
```diff
@@ -207,13 +207,21 @@ class ArtworkRail extends React.Component<Props & RelayPropsWorkaround, State> {
...
     const sideMargin = isPad ? 40 : 20
     const style: any = { marginLeft: sideMargin, marginRight: sideMargin }
-    if (!(this.props.relay && this.props.relay.variables.fetchContent)) {
+    if (!hasFetched) {
       // if there's no content, set minHeight to prevent spinners from all the rails showing
       style.minHeight = 400
     }
+
+    const hasArtworks = hasFetched && this.props.rail.results && this.props.rail.results.length
+    if (!hasArtworks) {
```
alloy: Isn’t this going to return `null` while it’s still fetching and thus not show the spinner like it should?

sarahscott: I think my variable name is misleading - `hasFetched` is actually more like `hasStartedFetching`. It's set [here](https://github.com/artsy/emission/blob/master/src/lib/components/home/artwork_rails/artwork_rail.tsx#L63). I can change the name so it makes more sense.

l2succes: @sarahscott You could rewrite the check that way so it'll only check if the results array is empty after it's done fetching data.

```javascript
    if (hasFetched) {
      const hasArtworks = this.props.rail.results && this.props.rail.results.length
      if (!hasArtworks) {
        // if the data has been fetched but there are no results, hide the whole thing
        return null
      }
    } else {
      // if there's no content, set minHeight to prevent spinners from all the rails showing
      style.minHeight = 400
    }

    return (
      <View accessibilityLabel="Artwork Rail" style={{ paddingBottom: this.state.expanded ? 0 : 12 }}>
        <Header rail={this.props.rail} handleViewAll={this.handleViewAll}/>
        <View style={style}>
          { this.renderModuleResults() }
        </View>
      </View>
    )
```

orta:
It'd be awesome to get a test or two for this behaviour change 👍

sarahscott:
Yeah, the rails are tricky to test but I've made an issue for that [here](https://github.com/artsy/emission/issues/534). It's next on my list. @l2succes , are we beta-ready?

l2succes:
Yes! we're good. let's go ahead and merge to cut a beta release we can spend time adding tests next

src/lib/components/home/artwork_rails/artwork_rail.tsx
```diff
@@ -207,13 +207,21 @@ class ArtworkRail extends React.Component<Props & RelayPropsWorkaround, State> {
...
       return null
     }
+    const hasStartedFetching = this.props.relay && this.props.relay.variables.fetchContent
 
     const sideMargin = isPad ? 40 : 20
     const style: any = { marginLeft: sideMargin, marginRight: sideMargin }
-    if (!(this.props.relay && this.props.relay.variables.fetchContent)) {
+    if (!hasStartedFetching) {
       // if there's no content, set minHeight to prevent spinners from all the rails showing
       style.minHeight = 400
```
alloy: This doesn’t look right. The rails should always start at min 400, currently they only show part of the spinner for a brief period of time.

PR #545, opened 2017-05-17
[Home] Fixes pull-to-refresh scroll position bug

sarahscott:
PTR now looks like this: 

https://artsy.slack.com/files/sarahscott/F5FD4QWNB/scroll.mov

### DIFF
--- a/src/lib/containers/home.tsx
+++ b/src/lib/containers/home.tsx
@@ -21,6 +21,7 @@ type DataSourceRow = {
 }
 
 interface Props extends ViewProperties, RelayProps {
+  trigger1pxScrollHack?: boolean
 }
 
 interface State {
@@ -84,8 +85,12 @@ export class Home extends React.Component<Props, State> {
     })).then(stopRefreshing, stopRefreshing)
   }
 
-  componentDidUpdate() {
-    this.listView.scrollTo({ y: this.currentScrollOffset + 1, animated: false })
+  componentDidUpdate(previousProps: Props) {
+    const didTrigger1pxScrollHack = (!!previousProps.trigger1pxScrollHack) === false
+                                    && this.props.trigger1pxScrollHack === true
+    if (didTrigger1pxScrollHack) {
+      this.listView.scrollTo({ y: this.currentScrollOffset + 1, animated: false })
+    }
   }
 
   render() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 1.3.4
 
+- Fixes pull-to-refresh scroll position bug on Home - sarah
 - Fixes loading state UI for artwork rail - luc
 
 ### 1.3.3


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>No CHANGELOG added.</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>Please assign someone to merge this PR, and optionally include people who should review.</td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/lib/containers/home.tsx
```diff
@@ -85,7 +85,10 @@ export class Home extends React.Component<Props, State> {
   }
 
   componentDidUpdate() {
-    this.listView.scrollTo({ y: this.currentScrollOffset + 1, animated: false })
+    if (this.currentScrollOffset < 0) {
```
alloy: So `componentDidUpdate` gets triggered after the refresh network fetch has occurred not immediately after triggering the refresh, right?

src/lib/containers/home.tsx
```diff
@@ -85,7 +85,10 @@ export class Home extends React.Component<Props, State> {
...
+      this.currentScrollOffset = 0
+    }
+    this.listView.scrollTo({ y: this.currentScrollOffset + 1, animated: !this.currentScrollOffset })
```
alloy: The `animated: !this.currentScrollOffset` doesn’t reveal the intent well, how about storing the result of `this.currentScrollOffset < 0` in `const didPullToRefresh` and use that for the `if` and this `animated` parameter?

src/lib/containers/home.tsx
sarahscott: It is triggered both before and after the request (the first `componentDidUpdate` console log there happens immediately after you release the drag gesture)

screen shot 2017-05-17 at 3 39 05 pm

alloy: Does that mean that the list will scroll back to 0 immediately after releasing the drag gesture, as opposed to after the network fetch completed?

sarahscott: Yes. It ensures the PTR looks nice each time, but it does not ensure the PTR works exactly the same as it did before. I can put some more time into making it only return to 0 upon fetch if that's necessary.

sarahscott:
Ready for review

src/lib/containers/home.tsx
```diff
@@ -21,6 +21,7 @@ type DataSourceRow = {
 }
 
 interface Props extends ViewProperties, RelayProps {
+  trigger1pxScrollHack?: boolean
```
alloy: Oh, if this can actually be `undefined` (because you marked it as optional) than I guess comparing just to `false` like we do in `componentDidUpdate` could lead to a false positive. In which case changing it to something like this should work:

```js
(!!previousProps.trigger1pxScrollHack) === false
```

sarahscott: Done!

sarahscott:
Merging!

PR #565, opened 2017-05-24
[Consignments] Adds a component for Artist Search

orta:
Adds support for rendering the Artist Search component.

Design:

screen shot 2017-05-24 at 13 54 57

### Changes from spec

* ~~Changed "DONE" to "BACK" - I'd expect that when you hit an artist it would count that as a selection and go back. Hitting "done" when you got no results didn't make sense in comparison to "back". If it's not transactional, done doesn't fit IMO~~

* Added a spinner for when we're getting artist results

* Changed the height just enough to indicate that there could be more results to scroll. You never know when your search gives you the 4th item that you actually want.

### Bugs

- [ ] For some reason the placeholder text isn't grey, I set it to green to test and it's not showing, I'll throw a bit more time at it but it's not a badge worth dying for.
- [ ] No X to clear button yet

### GIF

/cc @katarinabatina <hr data-danger="yep"/>

### Tested on Device?

- [ ] @orta
- [ ] @ashfurrow

<details>
  <summary>How to get set up with this PR?</summary>
  <p>&nbsp;</p>
   <p><b>To run on your computer:</b></p>
<pre><code>git fetch origin pull/565/head:orta-565-checkout
git checkout orta-565-checkout
yarn install
cd example; pod install; cd ..
open -a Simulator
yarn start</code></pre>
   </p>
   <p>Then run <code>xcrun simctl launch booted net.artsy.Emission</code> once a the simulator has finished booting</p>
   <p><b>To run inside Eigen (prod or beta) or Emission (beta):</b> Shake the phone to get the Admin menu.</p>
   <p>If you see <i>"Use Staging React Env" </i> - click that and restart, then follow the next step.</p>
   <p>Click on <i>"Choose an RN build" </i> - then pick the one that says: "X,Y,Z"</p>
   <p>Note: this is a TODO for PRs, currently  you can only do it on master commits.</p>
</details>

### DIFF
--- a/src/lib/components/consignments/__stories__/consignments-search.story.tsx
+++ b/src/lib/components/consignments/__stories__/consignments-search.story.tsx
@@ -1,0 +1,102 @@
+import * as React from "react"
+import { View } from "react-native"
+
+import { storiesOf } from "@storybook/react-native"
+import { camelCase } from "lodash"
+
+import Search, { ArtistQueryData } from "../components/artist-search-results"
+
+const noQuery: ArtistQueryData = {
+  query: null,
+  searching: false,
+  results: [],
+}
+
+const queryNoResults: ArtistQueryData = {
+  query: "ba",
+  searching: false,
+  results: [],
+}
+
+const queryNoResultsSearching: ArtistQueryData = {
+  query: "ba",
+  searching: true,
+  results: [],
+}
+
+const query1Result: ArtistQueryData = {
+  query: "ban",
+  searching: false,
+  results: [
+    {
+      name: "Banksy",
+      id: "banksy",
+      image: { url: "https://d32dm0rphc51dk.cloudfront.net/X9vVvod7QY73ZwLDSZzljw/square.jpg" },
+    },
+  ],
+}
+
+const query1ResultSearching: ArtistQueryData = {
+  query: "ban",
+  searching: true,
+  results: [
+    {
+      name: "Banksy",
+      id: "banksy",
+      image: { url: "https://d32dm0rphc51dk.cloudfront.net/X9vVvod7QY73ZwLDSZzljw/square.jpg" },
+    },
+  ],
+}
+
+const query2Results: ArtistQueryData = {
+  query: "bank",
+  searching: false,
+  results: [
+    {
+      name: "Banksy",
+      id: "banksy",
+      image: { url: "https://d32dm0rphc51dk.cloudfront.net/X9vVvod7QY73ZwLDSZzljw/square.jpg" },
+    },
+        {
+      name: "Banksy's Brother",
+      id: "thanksy",
+      image: { url: "https://d32dm0rphc51dk.cloudfront.net/zKBFBZPGN-V5TljXtsSuEg/square.jpg" },
+    },
+  ],
+}
+
+const artistGen = (name: string) => ({
+    name,
+    id: camelCase(name),
+    image: { url: "https://d32dm0rphc51dk.cloudfront.net/X9vVvod7QY73ZwLDSZzljw/square.jpg" },
+
+})
+
+const query4Results: ArtistQueryData = {
+  query: "bank",
+  searching: false,
+  results: [
+    artistGen("Ada Van Hoorebeke"), artistGen("Kathleen Adair Brown"),
+    artistGen("Linda Adair"), artistGen("Hector Adalid"),
+  ],
+}
+
+export const allStates = [
+  { "No query" : noQuery },
+  { "Looking for new results" : queryNoResultsSearching },
+  { "Found no results" : queryNoResults },
+  { "Found one result" : query1Result },
+  { "Found one result, and searching" : query1ResultSearching },
+  { "Found two results" : query2Results },
+  { "Found four results" : query4Results },
+]
+
+const stories = storiesOf("Consignments - Search")
+allStates.forEach(element => {
+  const name = Object.keys(element)[0]
+  stories.add(name, () =>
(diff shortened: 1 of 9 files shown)


katarinabatina:
I don't think back works in this context either. I wanted to create consistency between all the screens by having a done button. If we think it's confusing I would add a back button and remove the lower button.

orta:
cool 👍 - I'll switch it to back and once it's in context of running through the setup if it feels weird it can easily be changed

src/lib/components/consignments/__stories__/consignments-search.story.tsx
```diff
@@ -0,0 +1,102 @@
...
+    image: { url: "https://d32dm0rphc51dk.cloudfront.net/X9vVvod7QY73ZwLDSZzljw/square.jpg" },
+
+})
+
+const query4Results: ArtistQueryData = {
+  query: "bank",
+  searching: false,
+  results: [
+    artistGen("Ada Van Hoorebeke"), artistGen("Kathleen Adair Brown"),
+    artistGen("Linda Adair"), artistGen("Hector Adalid"),
```
ashfurrow: Any reason to keep these on the same line?

src/lib/components/consignments/__stories__/consignments-search.story.tsx
```diff
@@ -0,0 +1,102 @@
...
+  searching: true,
+  results: [],
+}
+
+const query1Result: ArtistQueryData = {
+  query: "ban",
+  searching: false,
+  results: [
+    {
+      name: "Banksy",
```
ashfurrow: Classic Orta.

src/lib/components/consignments/components/artist-search-results.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const noResults = (props) => {
+   if (!props.query || props.searching) { return null }
+   return <UnknownLabel>
+    Unfortunately we are not accepting consignments for works by <UnknownName>{props.query}</UnknownName>
+   </UnknownLabel>
+}
+
+export default (props: ArtistQueryData) =>
+  <View>
```
ashfurrow: Aren't container views in JSX usually wrapped in `()`? Or is that just in prediction?

orta: nope -  it's optional, maybe that's a typescript thing? But the `()` are only needed for something where it's ambiguous like a `() => { name: "thing"}` where it doesn't know whether the `{` refers to an object or a function opener

src/lib/components/consignments/__stories__/consignments-search.story.tsx
orta: Nope, one day prettier will come and maybe make that a one liner

src/lib/components/consignments/components/artist-search-results.tsx
```diff
@@ -0,0 +1,106 @@
+import * as React from "react"
+import { ActivityIndicator, ScrollView, View } from "react-native"
+
+import styled from "styled-components/native"
+import colors from "../../../../data/colors"
+import fonts from "../../../../data/fonts"
+
+import Button from "../../buttons/inverted_button"
+
+const Input = styled.TextInput`
```
l2succes: styled-components tho 👌🏿

src/lib/components/consignments/components/artist-search-results.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const rowForResult = (result) =>
+  <Result key={result.id}>
+    <Image source={{uri: result.image.url}} />
+    <Text>{result.name}</Text>
+  </Result>
+
+const noResults = (props) => {
+   if (!props.query || props.searching) { return null }
+   return <UnknownLabel>
```
l2succes: @orta we could use the same tslint config from reaction-force for code formatting including jsx
(https://github.com/artsy/reaction-force/blob/master/tslint.json)

src/lib/components/consignments/components/artist-search-results.tsx
alloy: Good question!

> Why do we need the parentheses around the return statement (line 3)? This is because of JavaScript's automatic semicolon insertion. Without the parentheses, JavaScript would ignore the following lines and return without a value. If the JSX starts on the same line as the return, then parentheses are not needed.

– http://buildwithreact.com/tutorial/components

src/lib/components/consignments/__stories__/consignments-search.story.tsx
```diff
@@ -0,0 +1,102 @@
...
+      id: "banksy",
+      image: { url: "https://d32dm0rphc51dk.cloudfront.net/X9vVvod7QY73ZwLDSZzljw/square.jpg" },
+    },
+        {
+      name: "Banksy's Brother",
+      id: "thanksy",
```
alloy: lol

src/lib/components/consignments/components/artist-search-results.tsx
alloy: So it also doesn’t really matter in the case of an arrow function without curly brackets and thus the result of the last expression in the function is going to be returned.

src/lib/components/consignments/components/artist-search-results.tsx
orta: Agree 👍 - yeah, that one is Emissions, then with some more improvements

orta:
Depending on when #568 is merged, this will need changing to point to master instead.

orta:
Alright, merging, it's had some 👍s and I can handle the cancel etc in a polish pass

Issue #577, opened 2017-05-31
Look into replacing Enzyme with ReactTestUtils


sarahscott:
Enzyme doesn't support the new version of React, and it could become difficult to maintain in the future. We should try to use the ReactTestUtils instead.

alloy:
Can you add the link to the Enzyme ticket?

sarahscott:
https://github.com/airbnb/enzyme/issues/953

alloy:
Done in #583

orta:
Either this or https://github.com/artsy/emission/issues/559 should probably still be active, it's tough make non-brittle tests that are not just straight up "render to snapshots" or "render to svg"

I investigated into react's shallow renderer 

```js
describe("User Interaction", () => {
  it("calls goto Artist on to the artist button", () => {
    const testFunc = jest.fn()

    const r = shallow.createRenderer()
    r.render(<TODO goToArtist={testFunc} />)
    const e = r.getRenderOutput()

    e.props.children[1].props.onPress()
    expect(testFunc).toBeCalled()
})
```

Which does something, but I can't find a safe way to find the thing I'm looking for when the objects all look like this:

```js
{ '$$typeof': Symbol(react.element),
      type: [Function: Button],
      key: null,
      ref: null,
      props:
       { onPress:
          { [Function: mockConstructor]
            _isMockFunction: true,
            getMockImplementation: [Function],
            mock: [Getter/Setter],
            mockClear: [Function],
            mockReset: [Function],
            mockReturnValueOnce: [Function],
            mockReturnValue: [Function],
            mockImplementationOnce: [Function],
            mockImplementation: [Function],
            mockReturnThis: [Function],
            mockRestore: [Function] },
         children: [ [Object], [Object], [Object] ] },
      _owner: null,
      _store: {} }
```

orta:
also ReactTestUtils only are for react-dom I think

alloy:
Not this ticket, I think, because this ticket was about _removing_ enzyme.

And what’s the problem with ReactTestUtils being liited to react-dom?

orta:
React-dom is only for the web

orta:
Alright, I'll bring https://github.com/artsy/emission/issues/559 back to life

alloy:
Hrm, afaik ReactTestUtils is supposed to work fine with RN.

orta:
They require a DOM, which RN doesn't have, for example:

```
import * as React from "react"
[...]

PR #583, opened 2017-06-02
Clean dependencies #trivial

### LINKED ISSUE #559: Improve the ability to introspect rendered react trees
Right now we're using jest-snapshots to say "this is the whole structure" of a react tree. We use those tests to verify that is what we want, but those tests are a really thick brush and say nothing about behaviours that you expect.

I'd like to write some tests that verify the behaviour of the react tree, for example:

```js
it("shows an image with the number of images when passed image data", () => {
 state = {
   photos: ["url1", "url2", "url3"],
 }
 const todo = renderer.create(<TODO {...state}/>).toJSON()
 const image = getComponent(todo, "Background Button[1] ImagePreview")
 expect(image).toBeTruthy()

 const text = getComponent(image, "ImagePreview ImageDarkener InlineCopy")
 expect(text.text).toEqual(3)
})
```

So we could consider looking into what the minimal parts of [Enzyme](https://github.com/airbnb/enzyme) can be found that work with react-test-renderer.

alloy:
Closes #559

# Number of deps before

1004

# Number of deps after

965


orta:
I thought Enzyme was in active use? https://github.com/artsy/emission/issues/559

alloy:
Enzyme is a bottleneck to upgrading, it doesn’t work the React 16 alpha that RN uses atm and it’s likely to happen again in the future: https://github.com/artsy/emission/issues/577.

I don’t know what was previously using it, it was probably just left behind after somebody tried to use it.

orta:
Looks like it got removed from the codebase two days ago 👍

sarahscott:
👌

PR #588, opened 2017-06-06
[Dev] Deploy the head commit for a PR to our S3 bucket

orta:
Bundles the JS from head of every PR and places it into an S3 bucket with the name `[PR_ID].js`. Meaning that we can make assumptions from inside Emission about the JS existing. Ergo: we can have testflight Emission testing PRs.

New commits on the same PR will overwrite the JS.

PR #590, opened 2017-06-07
[Dev] Allow any PR to be used as a build inside Emission

orta:
Allows anyone using the app to pick any current PR, assuming it has passed CI enough to upload the JS from #588 

Here's the flow:

The app normally looks like this
screen shot 2017-06-07 at 09 13 32

You want to look at the JS for a PR, so you click "Use a PR of Emission":
screen shot 2017-06-07 at 09 52 08

The grey'd out ones are unavailable to use because the JS isn't there, shouldn't be many in the future - but if we ship failing typescript then it won't bundle.

You pick one:
screen shot 2017-06-07 at 09 42 59

App restarts, and it shows a spinner while it downloads the JS from S3
screen shot 2017-06-07 at 09 43 05

This happens every time, so if the PR updates, you just kill the app and load it back up. Now the main menu looks like this:

screen shot 2017-06-07 at 09 43 16

And in this case, in the storybook browser, you'll get the stories for that PR, in my case that's the "Consignments - bottom aligned".

screen shot 2017-06-07 at 09 43 24

---

So why the S3 instead of apphub for PRs? 

Mainly, that this is really gonna mess up our AppHub builds - we'd be creating and deleting them a lot or it'd be full of builds. I know I regularly amend my commits on PRs and so we'd need to delete all apphub builds for the PR other than head on each CI run. It turns out the API is now closed-ish, so I'd have to email for details on how to seach and delete builds.

Shipping a to a single JS file per-PR seemed the simplest way to build it.

### DIFF
--- a/Example/Emission/PRNetworkModel.m
+++ b/Example/Emission/PRNetworkModel.m
@@ -1,0 +1,49 @@
+#import "PRNetworkModel.h"
+
+@implementation PRNetworkModel
+
+- (void)getPRs:(void (^_Nonnull)(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error))completionHandler
+{
+  NSURL *prListURL = [NSURL URLWithString:@"https://api.github.com/repos/artsy/emission/pulls"];
+  NSURLSessionDataTask *task = [[NSURLSession sharedSession] dataTaskWithURL:prListURL completionHandler:completionHandler];
+  [task resume];
+}
+
+- (void)verifyJSAtPRNumber:(NSUInteger)number completion:(void (^_Nonnull)(BOOL exists))completionHandler
+{
+  NSURL *url = [NSURL URLWithString:[NSString stringWithFormat:@"https://s3.amazonaws.com/artsy-emission-js/%@.js", @(number)]];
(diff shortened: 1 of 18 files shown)


Example/Emission/PRNetworkModel.m
```diff
@@ -0,0 +1,49 @@
...
+      completionHandler(nil, error);
+      return;
+    }
+
+    NSURL *fileURL = [self fileURLForPRJavaScript];
+    [data writeToURL:fileURL atomically:YES];
+    completionHandler(fileURL, nil);
+  }];
+  [task resume];
+}
```
alloy: Not that it matters much, but I was wondering if adding download code ourselves is really needed rather than passing the S3 URL directly to RN?

alloy:
Great stuff @orta 👏 🎉

Example/Emission/PRNetworkModel.m
orta: I thought of this and actually tried it first, ends up infinitely recursing, not actually sure why TBH

screen shot 2017-06-07 at 10 18 34

alloy: Hah, well, so much for putting on the academic hat.

damassi:
Really amazing, nice work @orta 🎉

PR #594, opened 2017-06-07
[MSG] update last message timestamp with real data

l2succes:
Sister PR: https://github.com/artsy/metaphysics/pull/649
Related: https://github.com/artsy/collector-experience/issues/295
simulator screen shot jun 7 2017 1 45 05 pm

### DIFF
--- a/src/lib/components/inbox/conversations/conversation_snippet.tsx
+++ b/src/lib/components/inbox/conversations/conversation_snippet.tsx
@@ -1,6 +1,6 @@
+import * as moment from "moment"
 import * as React from "react"
 import * as Relay from "react-relay"
-import * as moment from "moment"
 
 import {
   MetadataText,
--- a/src/lib/components/inbox/conversations/conversation_snippet.tsx
+++ b/src/lib/components/inbox/conversations/conversation_snippet.tsx
@@ -1,5 +1,6 @@
 import * as React from "react"
 import * as Relay from "react-relay"
+import * as moment from "moment"
 
 import {
   MetadataText,
@@ -80,6 +81,7 @@ export interface Conversation {
   from_email: string | null
   to_name: string | null
   last_message: string | null
+  last_message_at: string | null
   artworks: Array<{
     id: string | null
     href: string | null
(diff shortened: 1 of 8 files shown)


src/lib/components/inbox/conversations/conversation_snippet.tsx
```diff
@@ -109,7 +111,7 @@ export class ConversationSnippet extends React.Component<Props, any> {
     const artworkDate = `${artwork.date}`
     const artworkArtist = `${artwork.artist.name} · `
     const conversationText = conversation.last_message.replace(/\n/g, " ")
-    const date = "11:00 AM"
+    const date = moment(conversation.last_message_at).fromNow(true)
```
mennenia: from the example data below, I assume that this can return things like "a few seconds"? 

From what I remember, though not scoped [in this ticket](https://github.com/artsy/collector-experience/issues/295), I thought that anything less than a day will say the time, and anything more than a day will say "x days ago". 

Whatcha think?

Happy to ask for clarification and taking it on as a separate item, and getting this in as is :)

l2succes: @mennenia that's essentially what this call does. You can see exactly what it'll return for various durations here https://momentjs.com/docs/#/displaying/fromnow/

l2succes: So I think that'll do the trick but we can get some feedback from @katarinabatina later on after merging :)

mennenia: yeah I saw that but, and please correct me if I'm wrong, if it's 1pm now, and last message was 2 hours ago, 11am, will the string be `11:00 AM` or `2 hours ago`?

mennenia:
also, patience! :P haha

src/lib/components/inbox/conversations/conversation_snippet.tsx
mennenia: from what I gather it would be `2 hours ago`, but my comment was that from the designs it looks like it should be `11:00 AM`

l2succes: Ah right, it would say `2 hours ago` and you're saying it should say `11:00 AM` I get your point. Ok I can do that

l2succes:
@mennenia lol word, got a bit ahead of myself, I'll push a change to that branch anyway

mennenia:
@l2succes nah man! not necessary to fix now, but was just trying to explain the reason I was asking. Let's just wait and see what the spec is ^_^

PR #598, opened 2017-06-08
[Dev] Adds a recommended styled-components vscode extension - #trivial

orta:
I dunno what best practices are in styled-components yet, but I added an extension that will color the text correctly.

In this PR, I also changed all of the consignments components to use `x-y` instead of `xY` to be consistent with reaction-force, and added a `;` - as that fixes the syntax colouring :)

Before:
screen shot 2017-06-08 at 13 18 19

After:
screen shot 2017-06-08 at 13 18 57

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,6 +2,7 @@
 	// See http://go.microsoft.com/fwlink/?LinkId=827846
 	// for the documentation about the extensions.json format
 	"recommendations": [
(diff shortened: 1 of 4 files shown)


orta:
This PR isn't meant to do https://github.com/artsy/emission/issues/599 - just to be a small starting point.

mennenia:
Awesome! 👍

PR #600, opened 2017-06-08
[Dev] Remove yarn run from the package.json + danger tslint warnings

orta:
This should raise a tonne of warnings, as a test of Danger.

`yarn run` eats `'`s which we were using for globs. This means that our tslint check in the CI wasn't working as we hoped. So This PR _does not fix them_ - I will do that after I'm sure the Danger side works. 

I've made Danger now also post the tslint results to the PR, it's better than looking inside the CI for why it's red.

#trivial


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Please assign someone to merge this PR, and optionally include people who should review.
  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Changes were made to package.json, but not to yarn.lock - <i>Perhaps you need to run `yarn install`?</i>
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

dangerfile.ts
```diff
@@ -161,3 +162,23 @@ if (pr.body.replace(neuterMarkdownTicks, "-") !== newBody.replace(neuterMarkdown
...
 }
+
+// No `yarn run` inside the Package.json
+// It would make the shell glob instead of CLI tools which do a better job.
+// See: https://github.com/yarnpkg/yarn/issues/3595
+const packageText = fs.readFileSync("package.json", "utf8")
+if (packageText.includes("yarn run")) {
+  const url = "https://github.com/yarnpkg/yarn/issues/3595"
+  fail(`You have a \`yarn run\` inside the package.json. This is probably a mistake, see ${url}.`)
+}
```
orta: _strictly speaking_ this would not have caught this issue, I couldn't think of a way to detect the implicit `yarn run` but it definitely protects against it in the future

orta:
Looks like I had an older version of tslint in my `node_modules` 👍 - at least the danger warning and the tslint inline things are useful

orta:
Looks like this: not too much text, which is fine

screen shot 2017-06-08 at 15 33 33

src/lib/containers/__tests__/works_for_you-tests.tsx
```diff
@@ -99,21 +99,8 @@ interface NotificationsResponse {
...
-              image: {
-                resized: {
-                  url: string
-                }
-              }
-            }
           }
-        ]
+        >
       }
```
orta: @mennenia your interface here was wrong, you're actually putting an array inside `edge` - I've turned it into an array 👍

package.json
```diff
@@ -121,7 +122,7 @@
   "lint-staged": {
     "*.@(ts|tsx)": [
       "tslint --fix",
-      "yarn prettier-write --",
+      "npm run prettier-write --",
```
orta: I'd been having global prettier issues, probably due to my work when I first started using it for dev (so we could test the typescript implementation) seems fine after this

PR #609, opened 2017-06-12
[MSG] Adds Relay support to Conversation container

### LINKED ISSUE #605: [MSG] Get Relay data up and running for Conversation container

sarahscott:
Fixes #605 

simulator screen shot jun 12 2017 2 03 07 pm

### DIFF
--- a/src/lib/app_registry.tsx
+++ b/src/lib/app_registry.tsx
@@ -110,6 +110,14 @@ class Inbox extends RootContainer {
   }
 }
 
+class Conversation extends RootContainer {
+  constructor(props) {
+    super(props)
+    this.component = Containers.Conversation
+    this.route = new Routes.Conversation({ conversationID: props.conversationID })
+  }
+}
+
 AppRegistry.registerComponent("Consignments", () => Consignments)
 AppRegistry.registerComponent("Artist", () => Artist)
 AppRegistry.registerComponent("Home", () => Home)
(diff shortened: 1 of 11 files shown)


src/lib/relay/routes.tsx
```diff
@@ -18,6 +18,24 @@ class Artist extends Relay.Route {
   static routeName = "ArtistRoute"
 }
 
+class Conversation extends Relay.Route {
+  static queries = {
+    me: (component, params) => Relay.QL`
```
sarahscott: @alloy not sure if you know a way around this - it would be nice to use the existing `MyAccount` route but the Conversation container needs a `conversationID` parameter. I could make it optional on `MyAccount` but that might be worse than just having a unique `Conversation` route. I definitely can't do this:

```graphql
me {
    conversation(id: $conversationID) {
        ${component.getFragment("me", params)}
    }
}
```

because roots can only have one fragment and no fields. Would be nice though.

alloy: I do not, but also haven’t tried. I’d just add the extra route for now, though, maybe with Relay Modern this gets simplified anyways.

src/lib/app_registry.tsx
```diff
@@ -110,11 +110,19 @@ class Inbox extends RootContainer {
   }
 }
 
+class Conversation extends RootContainer {
+  constructor(props) {
+    super(props)
+    this.component = Containers.Conversation
+    this.route = new Routes.Conversation({ conversationID: props.conversationID })
```
mennenia: 🎉

src/lib/components/inbox/conversations/conversation_snippet.tsx
```diff
@@ -103,7 +102,7 @@ export class ConversationSnippet extends React.Component<Props, any> {
     const conversation = this.props.conversation
     const artwork = conversation.artworks[0]
 
-    const galleryName = conversation.to_name
+    const partnerName = conversation.to_name
```
mennenia: 👍

PR #615, opened 2017-06-13
[Dev] Update danger

orta:
Main update for us is that it adds a status report at the bottom showing whether green or red, but won't fail the build on a dangerfiile fail (the same behaviour as ruby Danger)

#trivial


orta:
Weird, I guess I didn't make status reports available for this token, ah well. https://github.com/danger/danger-js/issues/279

Issue #555, opened 2017-05-22
Storybooks isn't reloading the view via HMR


orta:
You can't press save and see the changes, you'd need to change between different stories. The HMRer is definitely doing it's work, you can see the notification - it's just that the storybook's content doesn't get updated.

Storybooks 3.0 is coming out soon, perhaps we can try that if a simple `yarn upgrade [storybooks]` doesn't help

orta:
This is what I mean by proof:

![screen shot 2017-05-22 at 13 40 20](https://cloud.githubusercontent.com/assets/49038/26309263/3b6d6454-3ef4-11e7-9724-e54d16472d23.png)

orta:
I [updated storybooks to 3.0.0](https://github.com/storybooks/storybook/issues/1112) and don't see HMR changes to our views inside storybooks.

The same type of changes work fine inside the app.

orta:
Look a look at the default settings inside storybooks-rn: https://github.com/storybooks/storybook/blob/master/app/react-native/src/server/config/webpack.config.js

Changing out config to 
```js
const path = require('path')
const webpack = require('webpack')

const { OccurenceOrderPlugin, includePaths, excludePaths } = require('@storybook/react-native/dist/server/config/utils')

module.exports = {
  devtool: "#inline-source-map", // Otherwise getting errors about e.g. `Relay` not being defined.
  resolve: {
    extensions: [".js", ".jsx", ".ts", ".tsx"],
  },
  module: {
    loaders: [
      {
        test: /\.tsx?$/,
        exclude: /node_modules/,
        loaders: ["awesome-typescript-loader?configFileName=./tsconfig.json&silent=true&transpileOnly=true&target=es6&useBabel=true&useCache=true"],
      },
    ],
  },
  plugins: [
    new OccurenceOrderPlugin(),
    new webpack.HotModuleReplacementPlugin(),
  ],
}
```

Didn't help - think that might only work for the server anyway.

orta:
#562 could help a bit here

orta:
This works for maxim

🌊

orta:
We discovered that live reloading works for maxim, not HMR

orta:
This works for me, in storybooks, on the Welcome page when I make changes to the text

orta:
Got it - a PureComponent does not get HMR

alloy:
Sigh. Are there docs for that?

orta:
https://github.com/facebook/react-native/issues/9152

[...]

PR #620, opened 2017-06-14
Adds Storybooks for the Artist search screen in consignments

orta:
Also:

* Simplifies the CHANGELOG, as it caused a bad deploy earlier, I also ran a markdown formatter on it
* ~~Adds a Danger rule that adds labels based on tags in commits~~
* Removes some test files that were the result of a bad merge somewhere (the files used kebab-case, not snake_case )
* Removes the comment about PureComponents, as they were removed due to #555


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Running your Dangerfile has Failed
  </td>
    </tr>
  </tbody>
</table>




## Danger has errored

Error: Error

```sh
undefined
```


<p align="right">
</p>

CHANGELOG.md
```diff
@@ -1,172 +1,180 @@
...
 
 ### 1.3.8 & 1.3.9
 
-- [dev] Moves native view controllers to use iOS7+ status bar api - ash
+-   [dev] Moves native view controllers to use iOS7+ status bar api - ash
 
 ### 1.3.6
 
-- Fixes missing compiled bundle - alloy
+-   Fixes missing compiled bundle - alloy
```
alloy: Why all these extra indents?

orta: Good question in general, seems to be recommend as the default in remark: https://github.com/wooorm/remark

src/lib/components/consignments/components/artwork_consignment_todo.tsx
```diff
@@ -86,45 +86,58 @@ const DoneButton = () =>
     <Image source={require("../images/black-tick.png")} />
   </ImageBG>
 
+const Button = (props: TouchableHighlightProperties) =>
```
l2succes: If you set the type of the stateless component the following way

```javascript
const Button: React.SFC<TouchableHighlightProperties> =
 ```

`props.children` will be defined

orta: Interesting, this does look like it should do it.

I _think_ this is more of a case that touchable highlight doesn't expect children (it can only have one child, which can have more)  as adding the specifier didn't help

screen shot 2017-06-14 at 13 36 29

orta:
Alright - that's green, I started getting used to having my web hooks verify tests too :D

orta:
@l2succes poke

l2succes:
Ok sounds good! merging 👌🏿

Issue #648, opened 2017-07-07
[MSG] Fix warnings emitted by React on the zero inbox state


alloy:
There are 7 warnings when we show the inbox zero state. They are about listviews rendering with empty header sections, at least 5 of them are in the `Conversations` component.

Issue #646, opened 2017-07-07
[MSG] Pull to refresh control is shown when loading the Inbox


alloy:
When opening the Inbox view, we immediately show the pull to refresh control. Because in most cases this won’t actually load (much) new data the control almost immediately hides again, which makes the overall experience one that makes the list appear to bounce.

The pull to refresh control should _only_ be shown when the user initiates a pull to refresh.

l2succes:
FIxed with https://github.com/artsy/emission/issues/649

Issue #644, opened 2017-07-07
[MSG] Inbox pull-to-refresh does not actually update


alloy:
Even though the Inbox view has pull-to-refresh support, it doesn’t actually fetch new data after the initial fetch. It should probably tell Relay to force fetch, otherwise it’s going to be serving from its cache.

Reproduction for new conversation:

1. Load the Inbox view
2. Make an inquiry on https://staging.artsy.net
3. Pull to refresh the Inbox view
4. No new conversation shows up

Reproduction for existing:

1. Load the Inbox view
2. Reply to a conversation from http://cms-staging.artsy.net/conversations
3. Pull to refresh the Inbox view
4. The existing conversation status and last message snippet isn’t updated

PR #649, opened 2017-07-07
[Inbox] Fixes pull to refresh

l2succes:
Fixes: https://github.com/artsy/emission/issues/648 https://github.com/artsy/emission/issues/646 https://github.com/artsy/emission/issues/644

Skip new tests

### DIFF
--- a/src/lib/Components/Inbox/Conversations/index.tsx
+++ b/src/lib/Components/Inbox/Conversations/index.tsx
@@ -97,11 +97,12 @@ export class Conversations extends React.Component<Props, State> {
       if (readyState.done) {
         this.setState({
           fetchingNextPage: false,
+          initialLoadDone: true,
           dataSource: this.state.dataSource.cloneWithRows(this.conversations),
         })
 
         if (!this.props.me.conversations.pageInfo.hasNextPage) {
-          this.setState({ completed: true, initialLoadDone: true })
+          this.setState({ completed: true })
         }
       }
     })
--- a/src/lib/Components/Inbox/Conversations/index.tsx
+++ b/src/lib/Components/Inbox/Conversations/index.tsx
@@ -25,6 +25,7 @@ interface State {
   dataSource: ListViewDataSource | null
   fetchingNextPage: boolean
   completed: boolean
+  initialLoadDone: boolean
 }
 
 export class Conversations extends React.Component<Props, State> {
@@ -39,6 +40,7 @@ export class Conversations extends React.Component<Props, State> {
       dataSource,
       completed: false,
       fetchingNextPage: false,
+      initialLoadDone: false,
     }
   }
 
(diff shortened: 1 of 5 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/Inbox/Conversations/index.tsx
```diff
@@ -112,8 +114,12 @@ export class Conversations extends React.Component<Props, State> {
         initialListSize={10}
         scrollEventThrottle={10}
         onEndReachedThreshold={10}
+        enableEmptySections={true}
         refreshControl={
-          <RefreshControl refreshing={this.state.fetchingNextPage} onRefresh={() => this.fetchData(false)} />
+          <RefreshControl
+            refreshing={this.state.fetchingNextPage && this.state.initialLoadDone}
+            onRefresh={() => this.fetchData(false)}
```
sarahscott: If this is all that's needed to refresh, that's awesome, but on Home we had to do a [`forceFetch` of each individual module](https://github.com/artsy/emission/blob/master/src/lib/Containers/Home.tsx#L62) to get accurate results so that's something to look out for.

l2succes: This is calling `fetchData` which in turn calls`forceFetch`  (https://github.com/artsy/emission/pull/649/files#diff-1a3b226b50eea831f2935ab045a6415aL94). The list updates fine so I think we're good. Home has nested listed in each row so that could also be why it needs more code to refresh everything.

src/lib/Components/Inbox/Conversations/index.tsx
```diff
@@ -99,7 +101,7 @@ export class Conversations extends React.Component<Props, State> {
         })
 
         if (!this.props.me.conversations.pageInfo.hasNextPage) {
-          this.setState({ completed: true })
+          this.setState({ completed: true, initialLoadDone: true })
```
sarahscott: Isn't it possible for there to be more pages after the initial load? i.e. is `initialLoadDone` only being set to `true` when `hasNextPage` is `false`?

l2succes: Yes, that's right. I'll move it out of the if statement :)

l2succes: We really need these test accounts with data at various states to test our code more thoroughly 😩... soon

sarahscott:
👏 👏 👏  merging!

src/lib/Components/Inbox/Conversations/index.tsx
alloy: So how does this actually fix the Inbox not updating previously?

PR #632, opened 2017-06-27
Adds support for looking through a user's photo library

orta:
End goal is this:
screen shot 2017-06-27 at 11 28 41

Not quite there yet as it just uses stubbed images inside the storybooks, I mainly wanted a PR so I could review the changes.

PR #652, opened 2017-07-07
[Dev] Remove all warnings from the project

orta:
We had let a few warnings slip into the project, which stopped our Eigen deployment - https://github.com/artsy/eigen/pull/2379#issuecomment-313712896

This happened because Emission's view controllers are a dev Pod, so their warnings were hidden in a mess of a warnings from all pods. 

So with this PR, I've removed all warnings from other pods, then made Emission's pods be warn as errors. As Emission's CI still runs on linux for speed, this means we won't accidentally ship warnings to downstream apps.


Example/Podfile.lock
```diff
@@ -140,7 +140,7 @@ CHECKOUT OPTIONS:
     :git: https://github.com/artsy/Artsy-Authentication.git
 
 SPEC CHECKSUMS:
-  AppHub: 03788112dd48c42b60d51321c0ffff4ef15a4432
+  AppHub: '03788112dd48c42b60d51321c0ffff4ef15a4432'
```
orta: Fixed in https://github.com/artsy/emission/pull/632

Pod/Classes/ViewControllers/ARInboxComponentViewController.m
```diff
@@ -16,20 +16,6 @@ - (instancetype)initWithInbox;
...
-    RCTRootView *rootView = self.view.subviews.firstObject;
-}
-
-- (void)viewDidDisappear:(BOOL)animated
-{
-    [super viewDidDisappear:animated];
-
-    RCTRootView *rootView = self.view.subviews.firstObject;
-}
-
```
orta: Discussion: https://artsy.slack.com/archives/C02BAQ5K7/p1499445667492158

alloy:
Taking this, looks good 👍

PR #664, opened 2017-07-10
[Inbox] Active Bids fixes

### LINKED ISSUE #662: [MSG] Active Bids aren’t tappable
They should probably take you to the artwork view.

### LINKED ISSUE #663: [MSG] Active Bids doesn’t reload
After bidding on a work and pull to refresh the bid is not shown. Only restarting the app works.

alloy:
Fixes #662 
Fixes #663

Also makes it so the `Inbox` always force fetches data on load so that the `ZeroInboxState` hides when bids/conversations exist.

Will comment inline.

Skip New Tests.

### DIFF
--- a/src/lib/Components/Inbox/Bids/ActiveBid.tsx
+++ b/src/lib/Components/Inbox/Bids/ActiveBid.tsx
@@ -134,32 +134,13 @@ export default Relay.createContainer(ActiveBid, {
       fragment on LotStanding {
         is_leading_bidder
         active_bid {
-          id
           max_bid {
-            cents
             display
           }
           sale_artwork {
-            lot_label
             lot_number
-            position
             reserve_status
-            counts {
-              bidder_positions
-            }
-            sale {
-              live_start_at
-              end_at
-              is_live_open
-              is_closed
-            }
-            highest_bid {
-              cents
-              display
-            }
             artwork {
-              id
-              title
               image {
                 url
               }
@@ -176,32 +157,13 @@ interface RelayProps {
   bid: {
     is_leading_bidder: boolean | null
     active_bid: {
-      id: string
       max_bid: {
-        cents: number | null
         display: string | null
       } | null
       sale_artwork: {
-        lot_label: string | null
         lot_number: string | null
-        position: number | null
         reserve_status: string | null
-        counts: {
-          bidder_positions: boolean | number | string | null
-        } | null
-        sale: {
-          live_start_at: string | null
-          end_at: string | null
-          is_live_open: boolean | null
-          is_closed: boolean | null
-        } | null
-        highest_bid: {
-          cents: number | null
-          display: string | null
-        } | null
         artwork: {
-          id: string
-          title: string | null
           image: {
             url: string | null
           } | null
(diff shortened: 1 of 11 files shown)


src/lib/Components/Inbox/Bids/ActiveBid.tsx
```diff
@@ -134,32 +134,13 @@ export default Relay.createContainer(ActiveBid, {
...
-              is_live_open
-              is_closed
-            }
-            highest_bid {
-              cents
-              display
-            }
             artwork {
-              id
-              title
```
alloy: This component was massively over-fetching data. Is that because of a certain workflow you using while building out these components? Are there other places we should check to ensure we don’t over-fetch?

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

- `src/lib/Components/Inbox/Bids/__tests__/index-tests.tsx`

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Containers/Inbox.tsx
```diff
@@ -2,51 +2,79 @@ import * as React from "react"
...
+      : <ZeroStateInbox />
   }
 }
 
+// FIXME The `lot_standings` snippet is copy-pasted from the `ActiveBids` component so it doesn’t fetch data that’s not
+//       really needed, MP should really just expose something like `has_active_bids` and ensure that it doesn’t perform
+//       extra backend reuqests to determine if it has active bids and resolve the `ActiveBids` query.
+//
+//       The same applies to the `conversations` snippet.
+//
```
alloy: See TODO.

src/lib/Components/Inbox/Bids/ActiveBid.tsx
l2succes: In this particular case, it was in part because the dev process involved a lot of trial and error to figure out where the display data came from. 

Also, since there are still some unknowns about active bids, namely where rows should be routing to (artwork view, auction page, etc...), we decided to hold on for that because the auctions team was changing auction urls. The plan was to update the fragment after we figure the new URI format. Anyway, sorry I should've left a comment explaining that.

src/lib/Components/Inbox/Bids/ActiveBid.tsx
```diff
@@ -95,6 +99,10 @@ class ActiveBid extends React.Component<RelayProps, State> {
     }
   }
 
+  handleTap() {
+    SwitchBoard.presentNavigationViewController(this, this.props.bid.active_bid.sale_artwork.artwork.href)
```
l2succes: I recall talking about routing to an auction lot VC for the bid in question but we can update that once we have the URI

l2succes:
I actually had it implemented that exact same way before but I changed it because pagination was broken on the Messages section for some reason. I'll just double-check it still works before merging.

src/lib/Components/Inbox/Bids/ActiveBid.tsx
alloy: Alrighty. So this isn’t an issue elsewhere?

l2succes: Not that I know of, I think it's an isolated case but we can always check other relay containers in the app to make sure we're not over-fetching data.

l2succes:
Ok, I tested the code on device, pagination still works and new data get loaded on pull-to-refresh so we're good. Merging

PR #671, opened 2017-07-11
[Podspec] Move RCTAnimation dependency to spec from Podfile.

alloy:
Fixes https://github.com/artsy/collector-experience/issues/402


mennenia:
🎉

PR #673, opened 2017-07-11
[MSG] Fix missing icons

### LINKED ISSUE #665: [MSG] Missing icons
image_0

image_0-2

alloy:
Fixes #665 

I’m not entirely sure why these assets need to be in the source root for them to be picked up by `react-native bundle`, but I didn’t feel like debugging this too much for now.

Skip New Tests


alloy:
FYI @l2succes @orta @sarahscott

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

- `src/lib/Components/Inbox/Conversations/__tests__/ZeroStateInbox-tests.tsx`

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

orta:
I think this file will also need a rename instead of telling Danger to skip the warning: `src/lib/Components/Inbox/Conversations/__tests__/ZeroStateInbox-tests.tsx`

orta:
Would it be bette to instead have a pre-comile script that moves all image files there before a deploy? Then you can still keep the images in the place where they right folders.

mennenia:
I think it would be pretty great. For now I'm going to merge this, just to get the images in :).

alloy:
I recall that in the past we spent a lot of time on this too, I think it was related to being able to find the images at runtime, we probably settled on this root images dir because of that. I’d definitely prefer to colócate images with the source, but I don’t wanna potentially dive into a rabbit hole on this specific issue atm.

PR #679, opened 2017-07-13
[Example] Fix archive build.

alloy:
It would fail due to the unused `testBundlePath` variable after stripping all assertions.


orta:
👍

PR #682, opened 2017-07-13
[MSG] Network connectivity state handling in Conversations

### LINKED ISSUE #683: [Messaging] Network connectivity banner in Conversation
Part of #625 

simulator screen shot jul 13 2017 4 15 34 pm

sarahscott:
- [x] Get a banner to display when network goes down
- [x] Tests

Fixes #683 

simulator screen shot jul 13 2017 4 15 34 pm

### DIFF
--- a/src/lib/Containers/Conversation.tsx
+++ b/src/lib/Containers/Conversation.tsx
@@ -4,11 +4,12 @@ import * as Relay from "react-relay"
 
 import { MetadataText, SmallHeadline } from "../Components/Inbox/Typography"
 
-import { FlatList, ImageURISource, ViewProperties } from "react-native"
+import { FlatList, ImageURISource, NetInfo, ViewProperties } from "react-native"
 import ReversedFlatList from "react-native-reversed-flat-list"
 
 import styled from "styled-components/native"
 import colors from "../../data/colors"
+import fonts from "../../data/fonts"
 import Composer from "../Components/Inbox/Conversations/Composer"
 import Message from "../Components/Inbox/Conversations/Message"
 import ArtworkPreview from "../Components/Inbox/Conversations/Preview/ArtworkPreview"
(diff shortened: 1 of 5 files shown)


src/lib/Containers/Conversation.tsx
```diff
@@ -64,14 +81,20 @@ interface Props extends RelayProps {
...
   constructor(props) {
     super(props)
+
+    // Assume if the component loads, connection exists (this way the banner won't flash unnecessarily)
     this.state = {
       sendingMessage: false,
+      isConnected: true,
     }
+
+    this.handleConnectivityChange = this.handleConnectivityChange.bind(this)
```
damassi: One thing you can do to simplify component constructors and binding is to use ES7 class properties and fat arrows, which are supported by TypeScript and remove a lot of boiler plate.   

This:

```ts
constructor(props) {
  super(props)

  // Assume if the component loads, connection exists (this way the banner won't flash unnecessarily)
  this.state = {
    sendingMessage: false,
    isConnected: true,
  }

  this.handleConnectivityChange = this.handleConnectivityChange.bind(this)
}
```

can be simplified to 

```ts
export class Conversation extends React.Component<Props, State> {
  // Instance variables can be set directly on the class
  state = {
    sendingMessage: false,
    isConnected: true,
  }

  // And the fat arrow binds to the surrounding scope so no need for `.bind(this)`
  handleConnectivityChange = (event) => {
    ...
  }
}
```

I wish this were more widely known because it saves so many LOCs!

alloy: We tried to use ES7 class properties in the past, but they were too experimental and they lead to hard to debug code in the Chrome Dev Tools.

damassi: They're fully supported by Typescript and part of the spec; what would be the issue now? https://www.typescriptlang.org/docs/handbook/classes.html

orta: This could have been fixed in Chrome/node by now also: https://github.com/Microsoft/vscode-react-native/issues/273

mennenia:
works really nicely!

src/lib/Containers/Conversation.tsx
mennenia: feel free to continue discussing, but for the time being I'm going to merge the PR :)

alloy: It may well be fixed by now, indeed, but let’s verify that before using it again.

alloy:
Nice 👌

PR #687, opened 2017-07-20
[WIP][MSG] Pending state and recommit for Messages in poor connectivity

sarahscott:
What I would like to do is create the pending state for Messages that are sent under conditions of poor connectivity: 

screen shot 2017-07-20 at 10 43 28 am

Unfortunately, what I've put together so far results in this error once network returns and `transaction.recommit()` is called:

simulator screen shot jul 20 2017 10 40 38 am

I think the Message component is created with the optimistic update but then destroyed when the transaction fails, taking the transaction with it. The question is, how can we get the optimistically updated component to persist and try again when the network connection returns?

Here are some things I've been looking at and trying to make sense of:
https://facebook.github.io/relay/docs/api-reference-relay-container.html#hasoptimisticupdate-example
https://github.com/facebook/relay/issues/1402
http://mgiroux.me/2016/relays-apply-update-function/

@l2succes and I talked about possibly waiting on this until we have support for local storage, and then creating a queue of pending messages. Otherwise, we can give Relay Modern a shot, but I'm not sure if it'll solve the problem or not.

### DIFF
--- a/src/lib/Components/Inbox/Conversations/Message.tsx
+++ b/src/lib/Components/Inbox/Conversations/Message.tsx
@@ -84,6 +84,11 @@ export class Message extends React.Component<Props, any> {
   render() {
     const { artworkPreview, initials, message, senderName } = this.props
     const isSent = this.props.relay ? !this.props.relay.hasOptimisticUpdate(message) : true
+    const transactions = this.props.relay.getPendingTransactions(message)
+
+    if (transactions && transactions.length) {
+      console.log(transactions[0].getStatus())
+    }
 
     return (
       <Container>
--- a/src/lib/Containers/Conversation.tsx
+++ b/src/lib/Containers/Conversation.tsx
@@ -81,6 +81,7 @@ interface Props extends RelayProps {
 interface State {
   sendingMessage: boolean
   isConnected: boolean
+  transaction?: any
 }
 
 export class Conversation extends React.Component<Props, State> {
@@ -137,6 +138,11 @@ export class Conversation extends React.Component<Props, State> {
 
   handleConnectivityChange(isConnected) {
     this.setState({ isConnected })
+
+    if (isConnected && this.state.transaction) {
+      this.setState({ sendingMessage: true })
+      this.state.transaction.recommit()
+    }
   }
 
   render() {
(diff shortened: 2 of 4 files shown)


src/lib/Components/Inbox/Conversations/Message.tsx
```diff
@@ -84,6 +84,11 @@ export class Message extends React.Component<Props, any> {
   render() {
     const { artworkPreview, initials, message, senderName } = this.props
     const isSent = this.props.relay ? !this.props.relay.hasOptimisticUpdate(message) : true
+    const transactions = this.props.relay.getPendingTransactions(message)
+
+    if (transactions && transactions.length) {
```
sarahscott: This is never anything but `COMMITTING`; I never get a `COMMIT_FAILED`. See https://facebook.github.io/relay/docs/api-reference-relay-container.html#getpendingtransactions-example

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Containers/Conversation.tsx
```diff
@@ -152,24 +158,23 @@ export class Conversation extends React.Component<Props, State> {
...
+            onFailure: failedTransaction => {
+              // TODO Actually handle errors
+              console.error(failedTransaction.getError())
+              this.setState({ sendingMessage: false, transaction: failedTransaction })
+            },
+            onSuccess: () => {
+              this.setState({ sendingMessage: false })
+            },
+          })
           this.setState({ sendingMessage: true })
```
alloy: You’re not storing the transaction in the state like [the article you linked me to](http://mgiroux.me/2016/relays-apply-update-function/) does, maybe that’s (part of) the issue?

sarahscott: It's the same error whether the transaction is stored every time or only when `onFailure` is called.

sarahscott: Also I'm not sure if the code in that article has actually been tested. It's one of the only things I could find on how to deal with failed transactions.

alloy:
Is it really needed to recommit the same transaction? Can’t you create a new one when the network comes back?

sarahscott:
In that case, we would lose the `pending` state in the design spec. The message would either exist as a successfully submitted message or not exist at all.

alloy:
Oh right, this

> the Message component is created with the optimistic update but then destroyed when the transaction fails

Hmm 🤔 

I probably won’t be able to say much more without actually diving into it myself, so let me know if you want to do so yourself more first.

sarahscott:
Closing

PR #699, opened 2017-07-28
Adds Provenance for Consignments flow

ashfurrow:
There is one problem with this PR but it may be something with React Native we need to fix. The placeholder text is limited to one line:

simulator screen shot jul 28 2017 2 42 57 pm

When inspected in Reveal, placeholders are implemented internally to RN as a `UILabel` but the multiline property of the text input isn't passed through. 

screen shot 2017-07-28 at 2 43 36 pm

Not sure if that's a blocker for this PR; I'm keen to hear next steps.

### DIFF
--- a/src/lib/Components/Consignments/Components/DoneButton.tsx
+++ b/src/lib/Components/Consignments/Components/DoneButton.tsx
@@ -10,7 +10,7 @@ export interface DoneButtonProps {
 const render = (props: DoneButtonProps) => {
   const doneButtonStyles = {
     backgroundColor: "black",
-    marginBottom: 20,
+    marginBottom: 0,
     paddingTop: 18,
     height: 56,
   }
--- a/src/lib/Components/Consignments/Components/DoneButton.tsx
+++ b/src/lib/Components/Consignments/Components/DoneButton.tsx
@@ -1,0 +1,23 @@
+import * as React from "react"
+
+import BottomAlignedButton from "./BottomAlignedButton"
+
+export interface DoneButtonProps {
+  onPress: () => void
+}
+
+const render = (props: DoneButtonProps) => {
+  const doneButtonStyles = {
+    backgroundColor: "black",
+    marginBottom: 20,
+    paddingTop: 18,
+    height: 56,
+  }
+  return <BottomAlignedButton onPress={props.onPress} bodyStyle={doneButtonStyles} buttonText="DONE" />
+}
+
+export default class DoneButton extends React.Component<DoneButtonProps, null> {
+  render() {
+    return render(this.props)
+  }
+}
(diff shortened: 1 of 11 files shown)


src/lib/Components/Consignments/Components/DoneButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+  children?: any[]
+}
+
+const render = (props: DoneButtonProps) => {
+  const doneButtonStyles = {
+    backgroundColor: "black",
+    marginBottom: 0,
+    paddingTop: 18,
+    height: 56,
+  }
```
ashfurrow: Please note that I removed at `marginBottom: 20` from these styles since it looked all funky. Let me know if it should still be there 👍

orta:
Yeah, I feel like the multi-line is a blocker, think it might be worth implementing your own version over it . Though this does feel. a bit odd - see https://facebook.github.io/react-native/docs/textinput.html#textinput

screen shot 2017-07-28 at 14 49 08

orta:
Will take a quick look at the input src in RN

orta:
Ccode-wise it is the [same RCTTextField](https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInput.js#L639-L660) for both placeholder/main string (it gets given the string if [value is empty](https://github.com/facebook/react-native/blob/master/Libraries/Components/TextInput/TextInput.js#L615) ) 

We aren't on master, so it could be a bug fixed since our release

orta:
So yeah, right now, make your own.

orta:
Ah yeah, Danger isn't running because this is on a fork

orta:
Maybe I can move Emission to Peril for non-build related stuff, seems to be going fine for emission

ashfurrow:
@orta okay, got this figured out here: https://github.com/artsy/emission/pull/699/commits/f80b2616d13e0d6afdb417e62ac21c45a821a90e

Thanks to @craigspaeth for helping bootstrap me on that 👍

ashfurrow:
Tests passed, just Danger failing the build.

src/lib/Components/Consignments/Components/DoneButton.tsx
orta: These should be consistent - so either all of them gets the change, or they all stay the same 👍 

I wonder if that difference is the spell checker bar?

orta: Ah yeah, this is its own class - I'll move other to it in another pr

src/lib/Components/Consignments/Components/DoneButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+      {props.children}
+    </BottomAlignedButton>
+  )
+}
+
+export default class DoneButton extends React.Component<DoneButtonProps, null> {
+  render() {
+    return render(this.props)
+  }
+}
```
orta: in case you didn't know - we do this extra Class at the end so [that HMR works](https://facebook.github.io/react-native/blog/2016/03/24/introducing-hot-reloading.html)

PR #700, opened 2017-07-28
Remove annoying logs around proptypes and createClass from RN codebase

orta:
I finally get some peace and quiet in the test runner

screen shot 2017-07-28 at 15 59 23

Most of these errors are fixed in the latest version of RN - so we can delete this on the next update.

### DIFF
--- a/src/setupJest.ts
+++ b/src/setupJest.ts
@@ -1,3 +1,18 @@
+const originalConsoleError = console.error
+
+// Remove on the next React-Native update.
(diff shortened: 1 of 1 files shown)


mennenia:
Sounds good! 

As an FYI, I'll just cc/ @sarahscott and @alloy, as they usually update Emission to the latest RN :).

alloy:
@sarahscott Yeah we should get on upgrading soon, so that we’ve had extensive time to test before shipping.

PR #709, opened 2017-08-10
[Consignments] Adds the final submission screen

orta:
This adds the a screen that is the final step in the submission process. Handles all the UI work, but none of the logic towards shipping the final build.

Made a nice GIF of my SVGs for testing too:

### DIFF
--- a/src/lib/Components/Consignments/Screens/FinalSubmissionQuestions.tsx
+++ b/src/lib/Components/Consignments/Screens/FinalSubmissionQuestions.tsx
@@ -1,0 +1,86 @@
+import * as React from "react"
+
+import { LayoutAnimation, NavigatorIOS, Route, ScrollView, View, ViewProperties } from "react-native"
+import ConsignmentBG from "../Components/ConsignmentBG"
+import { BodyText, LargeHeadline, Subtitle } from "../Typography"
+
+import { ArtistResult, ConsignmentSetup } from "../"
+import TODO from "../Components/ArtworkConsignmentTodo"
+import Text from "../Components/TextInput"
+import Toggle from "../Components/Toggle"
+import Artist from "./Artist"
+import Provenance from "./Provenance"
+import SelectFromPhotoLibrary from "./SelectFromPhotoLibrary"
+import Welcome from "./Welcome"
+
+interface Props extends ViewProperties {
+  navigator: NavigatorIOS
+  route: Route // this gets set by NavigatorIOS
+  setup: ConsignmentSetup
+}
+
+export default class FinalSubmissionQuestions extends React.Component<Props, ConsignmentSetup> {
+  constructor(props) {
+    super(props)
+    this.state = props.setup
+  }
+
+  submitWork = () => {
+    // OK
+  }
+
+  updateEdition = () => {
+    // React Native's Typings are wrong here, I want to pass in
+    // no arguments.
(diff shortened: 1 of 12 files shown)


src/lib/Components/Consignments/Screens/FinalSubmissionQuestions.tsx
```diff
@@ -0,0 +1,86 @@
...
+    // no arguments.
+    const animate = LayoutAnimation.easeInEaseOut as any
+    animate()
+
+    this.setState({
+      editionInfo: this.state.editionInfo ? null : {},
+    })
+  }
+
+  updateSigned = () => this.setState({ signed: this.state.signed ? false : true })
```
sweir27: Could you do `{ signed: !Boolean(this.state.signed) }`? (or just `!this.state.signed` if it's coming back as truthy already)

src/lib/Components/Consignments/Screens/FinalSubmissionQuestions.tsx
```diff
@@ -0,0 +1,86 @@
...
+          </View>
+          <View style={{ padding: 10 }}>
+            <View style={{ flexDirection: "row", paddingVertical: 6, alignItems: "center" }}>
+              <BodyText style={{ paddingLeft: 10, flex: 1, textAlign: "left" }}>Is this an edition?</BodyText>
+              <Toggle selected={!!this.state.editionInfo} left="NO" right="YES" onPress={this.updateEdition} />
+            </View>
+
+            {this.state.editionInfo
+              ? <View style={{ flexDirection: "row", paddingVertical: 6 }}>
+                  <Text text={{ placeholder: "Edition Size", keyboardType: "phone-pad" }} style={{ margin: 10 }} />
```
sweir27: Can we do some client-side validation here? We only accept edition sizes that are numbers in convection (`edition_number` is a string to accommodate values like `12a/100`).

src/lib/Components/Consignments/Components/__tests__/TextInput-tests.tsx
```diff
@@ -9,6 +9,6 @@ it("shows an activity indicator when searching ", () => {
 })
 
 it("does not have an activity when searching ", () => {
-  const component = renderer.create(<Text text={{ value: "My mocked" }} searching={true} />).toJSON()
+  const component = renderer.create(<Text text={{ value: "My mocked" }} searching={false} />).toJSON()
```
orta: oh yeah, this shouldn't have been commited 👍

src/lib/Components/Consignments/Screens/FinalSubmissionQuestions.tsx
orta: I bet the `Boolean` one might work, the later didnt'

src/lib/Components/Consignments/Screens/FinalSubmissionQuestions.tsx
orta: The `phone-pad` only allows numbers to be inputted., but it looks like the edition number will need to be a string - so I'll change that 👍

orta:
Alright 👍 feedback addressed, thanks!

orta:
Ah yeah, by the time Peril ran the branch had been deleted, need to take that into account

PR #713, opened 2017-08-14
remove unneeded placeholderBackgroundColor from props

erikdstock:
This PR resolves a minor storybook bug in displaying a single/multi-item article.

In `OpaqueImageView` the placeholder background color is processed with `processColor` if needed, but passed into props as a regular hex color string if not. This causes an error in the storybook.


orta:
Ah yeah, forks & Danger. Really should get this on peril.

erikdstock:
closing to PR from upstream repo

PR #719, opened 2017-08-20
[Dev/Admin] Update the admin/beta/dev UI

orta:
We have betas being released *mostly* automatically now (sometimes Xcode fails, so it needs too be re-built). Both in terms of the native code via TestFlight, and master JS commits via AppHub as of last Friday.

So, I downloaded a copy and played around with it. I think it's in a pretty reasonable state but it uses a lot of dev terminology which could be simplified.

So here's what a beta version looks like now:

screen shot 2017-08-20 at 12 24 58

If you wanted to choose a PR build:

screen shot 2017-08-20 at 12 35 23

(the tick indicates that it's definitely available to use)

Then you end up with a much simpler UI:

screen shot 2017-08-20 at 12 35 32

This is because there's no guarantee that any of the other things will work. You get access to the storybooks and you're good to go. Reverting from the PR build brings you back to a similar main menu.

---

I also removed the additional view controllers from the betas, it should encourage us to think a bit more about story hierarchy and coverage. Plus stories are dynamically updated per commit, whereas the view controllers are using native code and we should be moving away from requiring those changes unless necessary.

They're still there in dev mode though.

screen shot 2017-08-20 at 12 40 03

### DIFF
--- a/src/lib/Components/Inbox/Conversations/__tests__/Conversations-tests.tsx
+++ b/src/lib/Components/Inbox/Conversations/__tests__/Conversations-tests.tsx
@@ -21,7 +21,7 @@ const meProps = {
           from: { name: "Jean-Luc Collecteur", email: "luc+messaging@artsymail.com" },
           to: { name: "ACA Galleries" },
           last_message: "Karl and Anna... Fab!",
-          last_message_at: moment().subtract(30, "minutes").toISOString(),
+          last_message_at: moment().subtract(1, "year").toISOString(),
           created_at: "2017-06-01T14:14:35.538Z",
           items: [
             {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 -   AREmission bridging updates which require integration work in Eigen - orta
 -   You can choose any URL for gravity/metaphysics - orta
 -   Adds an env var for google maps API - orta
+-   Updates the beta/dev UI - orta
 
 ###### Consignments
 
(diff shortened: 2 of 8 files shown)


orta:
CI fails are interesting, will take a look

alloy:
Looks like we got ourselves a little time dependent speccy.

orta:
Interesting, I found the code: `last_message_at: moment().subtract(30, "minutes").toISOString(),` - unsure that it takes an additional 4 minutes to do anything

orta:
Switched the 30m to 1 year, a few minutes either side isn't going to change the text then ;)

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

orta:
Added the changelog to the ignored files in spellchecker https://github.com/artsy/artsy-danger/commit/2e132ca5f2e6b07b3e1b8c228d767b4df3a207e0

orta:
🥗

src/lib/Components/Inbox/Conversations/__tests__/Conversations-tests.tsx
```diff
@@ -21,7 +21,7 @@ const meProps = {
           from: { name: "Jean-Luc Collecteur", email: "luc+messaging@artsymail.com" },
           to: { name: "ACA Galleries" },
           last_message: "Karl and Anna... Fab!",
-          last_message_at: moment().subtract(30, "minutes").toISOString(),
+          last_message_at: moment().subtract(1, "year").toISOString(),
```
alloy: This test would ideally mock the `moment` module rather than letting it do actual time calculations.

alloy:
Nice! 👌 

I had one piece of feedback re the time issue in the previously failing test. Feel free to address that feedback in another PR (if at all) and merge this now.

orta:
Agree WRT to the mocks, but I wanna deploy a release - so merging.

PR #722, opened 2017-08-22
[Dev] Move all conversations test to show 1 year instead of 30m as 30m sometimes goes awry

orta:
It's not a perfect solution, like mocking moment, but I wanted to get on with consignments work.


alloy:
Fair 👍

PR #723, opened 2017-08-23
Extracted connectivity banner

sarahscott:
Now we can use it anywhere 😄 

screen shot 2017-08-23 at 3 12 55 pm


mennenia:
looks great :)

Issue #677, opened 2017-07-12
Composer Interface Tweaks


katarinabatina:
Remaining work: Nice 1px solid separator line. Japanese dots sent back to the icebox as a universal design component.

------

Type size of user labels appears to be 9 or 10px, should be 11px
![image](https://user-images.githubusercontent.com/6482926/28098116-6ba6fa80-6682-11e7-9742-d0da45efb351.png)

Sketch makes it hard to express "Japanese dots" as a line quality but outlined here is the desired dotted line versus what we have now 
![image](https://user-images.githubusercontent.com/6482926/28098204-f0b6f658-6682-11e7-985e-316f1e4c19aa.png)

 time stamp and user spacing 
![image](https://user-images.githubusercontent.com/6482926/28098298-7c03f3dc-6683-11e7-91da-1559b8fc4f58.png)

mennenia:
@katarinabatina 

so the font is actually 11pt... I tried changing it to 12, but then it looks too big, so I settled on 11.5, which looks a bit like what you're going for. 

Do I then also change the "date" label to the larger font? Currently, the timestamp looks a bit off compared to the name:

<img width="325" alt="screen shot 2017-07-14 at 14 48 59" src="https://user-images.githubusercontent.com/373860/28214801-97164f68-68a3-11e7-8eb1-299bc11d6a12.png">

Spacing has been fixed ☑️ 

The dotted line thing looks like it's impossible in react native, at least from CSS. Even the dotted style we're using, are actually just a bunch of squares :/. You cannot specify the gap between them unfortunately.

I can probably create a custom view where we create a bunch of miniature circles, and line them up, etc. I don't know much about the performance impact of this approach, but I can check further. I would probably leave that for a bit later though, if that's ok?

katarinabatina:
@mennenia yes change for time stamp too

bloomberg uses these dotted lines a lot, I feel like there's gotta be a baseline css style that allows us to do this? https://www.bloomberg.com/markets

Maybe it's about making the rule "less than a pixel" ?

mennenia:
Sorry, I should have clarified: it exists in web CSS, but there's only select CSS styles that RN makes available and can transform into iOS components and views. RN is also in control what a style then translates to or appears as. 

In this case, `border-style` exists for both RN and web. We [set the border to dotted](https://github.com/artsy/emission/blob/master/src/lib/Containers/Conversation.tsx#L50), and I have no further properties or control as to what that then looks like (which is the square like, frequent, dot we see).

[...]

PR #729, opened 2017-08-26
Dotted line component

erikdstock:
tracked in https://github.com/artsy/emission/issues/677

This PR adds a simple horizontal `DottedLine` component to be used as a separator. Here is the actual aligned with the spec:
image

With scaled dots (`scale={5}`):
image
With color (`color={colors['purple-regular']`}
image

I'm not sure if anything is lacking for me to have it display in a conversation- whether i need to log in as a certain user or if these storybook views simply need to be created. **I'm also wondering - does this need a storybook implementation where it displays in context, or is the existing dotted line story sufficient?** 

Thanks!

### DIFF
--- a/Pod/Classes/DottedLineComponent/ARDottedLineManager.m
+++ b/Pod/Classes/DottedLineComponent/ARDottedLineManager.m
@@ -3,6 +3,8 @@
 
 
 @interface ARDottedLine : UIView
+@property (nonatomic, strong, readwrite) UIColor *color;
+@property (nonatomic, strong, readwrite) UIColor *processedBackgroundColor;
 @end
 
 
@@ -12,48 +14,46 @@
 {
     [super drawRect:rect];
     
-    [[UIColor whiteColor] setFill];
-    [[UIColor grayColor] setStroke];
+    [self.processedBackgroundColor setFill];
+    [self.color setStroke];
     
     CGContextFillRect(UIGraphicsGetCurrentContext(), rect);
-    //
     
     const CGFloat dotDiameter = rect.size.height / 2;
     const CGFloat gapSize = dotDiameter * 5;
     CGFloat pattern[4];
     pattern[0] = 0.0;
     pattern[1] = gapSize;
-//    const CGFloat pattern = 0.5;
+
     
     UIBezierPath *path = [UIBezierPath bezierPath];
     path.lineWidth = dotDiameter;
     path.lineCapStyle = kCGLineCapRound;
-    [path moveToPoint:CGPointMake(0, dotDiameter / 2)];
-    [path addLineToPoint:CGPointMake(rect.size.width, 2)];
+    [path moveToPoint:CGPointMake(0, dotDiameter)];
+    [path addLineToPoint:CGPointMake(rect.size.width, dotDiameter)];
     [path setLineDash:pattern count:2 phase:2];
 
     [path stroke];
 }
 
-//- (void) layoutSubviews {
-//    [super layoutSubviews];
-//    for(UIView* view in self.subviews) {
-//        [view setFrame:self.bounds];
-//    }
-//}
-
-
 @end
 
 
 @implementation ARDottedLineManager
+RCT_CUSTOM_VIEW_PROPERTY(color, NSNumber, ARDottedLine)
+{
+    view.color = [RCTConvert UIColor:json];
+}
+RCT_CUSTOM_VIEW_PROPERTY(processedBackgroundColor, NSNumber, ARDottedLine)
+{
+    view.processedBackgroundColor = [RCTConvert UIColor:json];
+}
 RCT_EXPORT_MODULE();
 
 - (UIView *)view
 {
     
     ARDottedLine *line = [ARDottedLine new];
-//    line.backgroundColor = [UIColor redColor];
     return line;
 }
 
--- a/Pod/Classes/DottedLineComponent/ARDottedLineManager.m
+++ b/Pod/Classes/DottedLineComponent/ARDottedLineManager.m
@@ -1,0 +1,60 @@
+#import "ARDottedLineManager.h"
+
+
+
+@interface ARDottedLine : UIView
+@end
+
+
+@implementation ARDottedLine
+
+- (void)drawRect:(CGRect)rect
+{
+    [super drawRect:rect];
+    
+    [[UIColor whiteColor] setFill];
+    [[UIColor grayColor] setStroke];
+    
+    CGContextFillRect(UIGraphicsGetCurrentContext(), rect);
+    //
+    
+    const CGFloat dotDiameter = rect.size.height / 2;
+    const CGFloat gapSize = dotDiameter * 5;
+    CGFloat pattern[4];
+    pattern[0] = 0.0;
+    pattern[1] = gapSize;
+//    const CGFloat pattern = 0.5;
+    
+    UIBezierPath *path = [UIBezierPath bezierPath];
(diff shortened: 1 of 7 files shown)


src/lib/Components/DottedLine.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from "react"
+import { requireNativeComponent, View } from "react-native"
+
+class DottedLine extends React.Component<any, any> {
+  static propTypes = { dotScale: React.PropTypes.number }
```
damassi: Not sure if emission has updated, but `React.PropTypes` has been deprecated and is available in the `prop-types` package: 

```js
import PropTypes from 'prop-types'
```

erikdstock: Ah yes. I'll switch it- thanks.

Pod/Classes/DottedLineComponent/ARDottedLineManager.m
```diff
@@ -0,0 +1,60 @@
...
+
+- (UIView *)view
+{
+    
+    ARDottedLine *line = [ARDottedLine new];
+//    line.backgroundColor = [UIColor redColor];
+    return line;
+}
+
+@end
```
ashfurrow: There're some extra, commented-out code that could probably be removed.

src/lib/Components/DottedLine.tsx
alloy: Can you add documentation for the `dotScale` prop?

Pod/Classes/DottedLineComponent/ARDottedLineManager.m
```diff
@@ -0,0 +1,60 @@
...
+    CGFloat pattern[4];
+    pattern[0] = 0.0;
+    pattern[1] = gapSize;
+//    const CGFloat pattern = 0.5;
+    
+    UIBezierPath *path = [UIBezierPath bezierPath];
+    path.lineWidth = dotDiameter;
+    path.lineCapStyle = kCGLineCapRound;
+    [path moveToPoint:CGPointMake(0, dotDiameter / 2)];
+    [path addLineToPoint:CGPointMake(rect.size.width, 2)];
```
alloy: Reading this code it looks like you can control the size of the dots by changing the height of the `DottedLine` component. If correct, can you add a second story that demonstrates this?

src/lib/Components/DottedLine.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from "react"
+import { requireNativeComponent, View } from "react-native"
+
+class DottedLine extends React.Component<any, any> {
```
alloy: Can you add a TypeScript interface for the props that describes the `dotScale` prop? And specify that the state will be `null`?

src/lib/Components/__stories__/DottedLine.story.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import DottedLine from "../DottedLine"
+
+storiesOf("App Style/Dotted Line")
+  .addDecorator(story =>
+    <View accessibilityLabel="wrapperView" style={{ height: 10, marginTop: 60, marginLeft: 20, marginRight: 20 }}>
+      {story()}
+    </View>
+  )
+  .add("Flat White", () => {
```
alloy: I think this title is the result from some copy-pasta, right? Can you name this something like ‘With default props’?

erikdstock: Oh man. I kept looking at that name and thinking 'why did we name it this when it is neither flat nor white?'

src/lib/Components/DottedLine.tsx
orta: ( there's some docs on how that works here: https://github.com/artsy/reaction/blob/master/docs/getting_started.md )

erikdstock: yep, thank you!

src/lib/Components/__stories__/DottedLine.story.tsx
alloy: 😂

erikdstock:
Alright, I was able to come back to this and am getting close to finishing up but am not sure about how this should work: basically the container component in the storybook does not constrain or set bounds for the dotted line inside it- if the dots are scaled larger they will expand outside the container. Here i have set the height of that 'wrapperView' artificially low to 2. 

My goal was to have some basic color and scale controls, but I'm also wondering if this isn't worth it.

image

mennenia:
Amazing @erikdstock!!! Looks GREAT 😍

alloy:
@erikdstock I think the component should internally default to the smallest height that looks good (as per the design), but you can definitely expose that property as something that the user can control. To be clear, that means that the `DottedLine` wrapper component would set that height on the `NativeDottedLine`. Does that answer your question?

erikdstock:
@alloy I believe that is what I'm doing right now. The specific issue pictured above is that the story above has a wrapper (unrelated to the DottedLine component itself) and that wrapper does not affect the line. Since it is a line, maybe it shouldn't matter. But to clarify, besides the very low wrapperView height I set above when I was figuring this out, this is the setup you want?
```jsx
// story
    <View accessibilityLabel="wrapperView" style={{ height: 2, marginTop: 60, marginLeft: 20, marginRight: 20 }}>
      {story()}
    </View>

// DottedLine render()
    return <NativeDottedLine style={{ height: this.props.scale * 2 }} color={dotColor} />
```

erikdstock:
Updated with documentation- but I still can't see it in context of the `Conversation`. Is that storybook broken, or can I see it when I am logged in as a different user?

If we don't need that, I removed the WIP line above.

Pod/Classes/DottedLineComponent/ARDottedLineManager.m
```diff
@@ -0,0 +1,55 @@
...
+@end
+
+
+@implementation ARDottedLine
+
+- (void)drawRect:(CGRect)rect
+{
+    [super drawRect:rect];
+    
+    [[UIColor whiteColor] setFill];
```
alloy: I wonder if this should be `clearColor` instead and that’ll show the white background underneath. If it doesn’t then this is fine too 👍

alloy:
When used in the conversation and inspecting the view hierarchy with Xcode’s view debugger, do you see it exist at all and what are its dimensions?

erikdstock:
Had a chanc3e to revisit today- showing as just the height of the dots.
However, these were initially rending with `clearColor()` on a black background, so I set it up with a custom `backgroundColor` prop. In doing that I noticed some XCode documentation pop up saying that backgroundColor defaults to `nil`/transparent so to avoid possibly colliding with something I changed the name to `processedBackgroundColor`. Then it turned black, so I made it white by default. `backgroundColor={null}` in props will achieve the `UIColor *nil` black effect, but I don't know if that is my error or the actual color of a transparent view 🤔. 
image

Actual view:
image

alloy:
Yeah instead of `nil` you should have it default to `[UIColor clearColor]` and let it draw that, by default the graphics context is going to be black.

damassi:
One little design nit -- maybe increase the line padding between each message item so that the margins between appear evenly spaced? Right now the line is much closer to "Karl and Anna" then it is to "ACA Galaries" (or the thumbnail image) which throws off the overall design.

erikdstock:
Updated with @orta to fix the margin in context and remove height/color options- will always now be `gray-medium` on white with a scale to match the spec:
image

erikdstock:
Closing to open a new pr directly from the upstream.

PR #730, opened 2017-08-26
Set up Android for emission

yuki24:
This is a complete redo of https://github.com/artsy/emission/pull/225 and my hackathon project. In orders to minimize the initial diff (although this is still really big) and make it easier for other to understand the entire change, I focused on just setting up the project rather than going all the way to implement integration with existing stuff. In summary, I have made these changes:

 * ran `react-native init android` and moved the entire project up to the root directory
 * updated `compileSdkVersion`, `buildToolsVersion`, `targetSdkVersion`, and the support libraries to `25`, which is compatible with the latest React Native version
    * Android O (latest version as of writing) is not supported: https://github.com/facebook/react-native/pull/15601
 * updated the android gradle plugin to 2.3.3 that has [Instant Run capability](https://developer.android.com/studio/run/index.html#instant-run)

Right now, it still uses the default component generated by the `react-native init` command. Naturally, the next step will be to have it integrate with Emission's existing components.

screen


ashfurrow:
Cool, looks fairly straightforward! Looking forward to the blog post :)

yuki24:
looks like the build is broken now. I'll look into it.

orta:
@yuki24 Emission works off local branches, not forks

yuki24:
It seems like I shouldn't have forked emission. I'll re-send this PR so the build will get proper ENV variables.

PR #732, opened 2017-08-30
[Dev] Separate deploy runs from CI runs

orta:
Separates the Danger/Build/Lint runs from the deployment runs. I think they're also running when they shouldn't - so I took another stab at making them only run at the right place

#trivial

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,22 @@
 cache: yarn
-
 language: node_js
-node_js:
-  - "6.7"
 
-script:
-  - scripts/deploy_commit.sh
-  - yarn ci
-  - yarn danger
-  - scripts/deploy_apphub.sh
+matrix:
+  include:
+  - node_js: '6.7'
+    script:
+      - yarn ci
+      - yarn danger
+
+  # Deployment related tasks
+  - node_js: "node"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,14 +1,19 @@
...
-  - scripts/deploy_apphub.sh
+matrix:
+  include:
+  - node_js: '6.7'
+    script:
+      - yarn ci
+      - yarn danger
+
+  # Deployment related tasks
+  - node_js: "node"
```
alloy: Why is this version requirement different than the one above?

orta: You can't have the same env var for both

alloy: Ah. Do we still do `nvm use` after the VM has started? I’d prefer it that the deployment runs on the same versions we decided to use locally.

orta: We could, that should work. 👍

orta: Yep that works - but THB, I'm not convinced that it's uploading when it should be - will take another look

orta:
OK, this is good to go on green

orta:
Gonna merge so I can check pushes the master commits bit

alloy:
👍

PR #733, opened 2017-09-02
[WIP] React Native Upgrade

sarahscott:
This branch is in a state where everything should be able to run as expected, with the exception of nested scrollviews (e.g. an Artist's Artworks masonry grid). It requires a monkey-patched version of Relay with a small change re: `unstable_batchedUpdates`; I have the change locally and can help someone out with it if they would like to use this branch before it is merged. 

- [x] Upgrade RN
- [x] Upgrade RN to actual v0.48.0 now that it's no longer a release candidate
- [ ] Decide how to deal with the Relay monkey patch (on @alloy's custom branch perhaps?)
- [ ] Fix nested scrollviews (everything should run without crashing but `onScroll` callbacks will not be called; the `RCTScrollView` API has changed considerably so this might take a bit of time)
- [ ] Adapt usage of `NetInfo` to new API (the `change` event is now deprecated)

Please feel free to add to the above list.

### DIFF
--- a/transformer/index.js
+++ b/transformer/index.js
@@ -8,7 +8,7 @@ const { File } = require("babel-core/lib/transformation/file")
 const { SourceMapConsumer } = require("source-map")
 
 const upstream = require("./upstream")
-const { compactMapping } = require("react-native/packager/src/Bundler/source-map")
+const { compactMapping } = require("metro-bundler/src/Bundler/source-map")
 
 /**
  * This is a copy of upstream.transform, but modified to:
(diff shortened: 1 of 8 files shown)


transformer/index.js
```diff
@@ -8,7 +8,7 @@ const { File } = require("babel-core/lib/transformation/file")
 const { SourceMapConsumer } = require("source-map")
```
orta: Might be worth moving to https://github.com/ds300/react-native-typescript-transformer - it is based on our Emission code

alloy: 👍

PR #738, opened 2017-09-06
[Consignments] Links the location button to the location screen and adds tests for the regression

orta:
Tests are so easy to write in React.

#trivial


orta:
Got a much bigger one coming down the line

PR #750, opened 2017-09-12
[conversation] Implements pagination container

l2succes:
This converts our Conversation component to use the pagination container abstraction class from relay modern.

It currently isn't working as needed because the following variables are not being sent to metaphysics in the query `cursor`, `count`. I am currently trying to figure out why. The [docs](https://facebook.github.io/relay/docs/pagination-container.html) don't really cover how to setup the `QueryRenderer` parent container. I suspect it's part of the reason why it isn't working.

Skip New Tests


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Containers/Conversation.tsx
```diff
@@ -276,60 +288,85 @@ function sendConversationMessage(
...
+      return {
+        ...fragmentVariables,
+        count,
+        cursor,
+      }
+    },
+    query: graphql.experimental`
+      query ConversationPaginationQuery($conversationID: String!, $count: Int!, $cursor: String) {
+        me {
+          ...Conversation_me
```
mzikherman: i think this also needs a `@arguments(count: $count, cursor: $cursor)`?

src/lib/Containers/Conversation.tsx
```diff
@@ -276,60 +288,85 @@ function sendConversationMessage(
...
+            initials
           }
-        }
-        items {
-          item {
-            ... on Artwork {
-              __typename
-              href
-              ...ArtworkPreview_artwork
+          messages(first: $count, after: $cursor) {
```
mzikherman: This may need a `@argumentDefinitions(count: { type: "Int", defaultValue: 10 }, cursor: { type: "String", defaultValue: "" })` as well.

mzikherman:
@l2succes made a couple comments about the `argument` thing....docs are sparse but I was looking at https://github.com/artsy/emission/blob/4e4f71879630b4c4252678077d9b563b8d7444c2/src/lib/Components/Inbox/Conversations/index.tsx#L123

That pagination seems to work (I see the subsequent page requests happening).

src/lib/Containers/Conversation.tsx
l2succes: Yes, but not in the query.

l2succes:
Yeah, thanks for the comments. I figured it out though

src/lib/relay/QueryRenderers.tsx
```diff
@@ -60,21 +60,24 @@ export const ArtistRenderer: React.SFC<ArtistRendererProps> = ({ render, artistI
...
   return (
     <QueryRenderer
       environment={environment}
-      query={graphql`
-        query QueryRenderersConversationQuery($conversationID: String!) {
+      query={graphql.experimental`
+        query QueryRenderersConversationQuery($conversationID: String!, $count: Int!, $after: String) {
           me {
-            ...Conversation_me
+            ...Conversation_me @arguments(count: $count, after: $after)
```
alloy: Why do you want to pass `count` and `after` in from here at all? Why does the root query need to know anything about a pagination container somewhere down the tree?

src/lib/Components/Inbox/Conversations/Messages.tsx
```diff
@@ -0,0 +1,176 @@
...
+        conversationID: props.conversation.id,
+        count: paginationInfo.count,
+        after: paginationInfo.cursor,
+      }
+    },
+    query: graphql`
+      query MessagesPaginationQuery($conversationID: String!, $count: Int!, $after: String) {
+        me {
+          conversation(id: $conversationID) {
+            ...Messages_conversation
```
alloy: You probably should use `@arguments` here and add `@argumentDefinitions` in the fragment above, at least that’s what I’ve been doing, still not entirely sure when that’s needed.

alloy: Ohhh, only now see your comments.... sorry for piling on

src/lib/relay/QueryRenderers.tsx
l2succes: I get the following errors back if I don't include those fields in the Conversation query renderer.
screen shot 2017-09-13 at 2 54 49 pm

src/lib/Components/Inbox/Conversations/Messages.tsx
sibelius: it is needed for local variables definitions

alloy: @sibelius Right, and so it scopes them to just these fragments? Looking at the docs, it seems like they indicate that it should be possible to do without the experimental features, is that really true? Because I had a hard time trying that.

src/lib/relay/QueryRenderers.tsx
alloy: That’s probably because you’re missing the `@argumentDefinitions` [here](https://github.com/artsy/emission/pull/750/files#diff-2bff25af92fc79a4ace9f3098d77d65fR102). Look at the other pagination containers for examples.

src/lib/Components/Inbox/Conversations/Messages.tsx
sibelius: graphql.experimental is deprecated on master already

alloy: Makes sense, good to know! 👍

l2succes:
Ok! pagination is now working and a loading indicator is displayed when more messages are fetched.

Closes: https://github.com/artsy/collector-experience/issues/486

l2succes:
@mzikherman I think it's ready to merge

sarahscott:
Looks great 👏  !

sarahscott:
Maybe we can improve the Relay docs using this newfound information?

alloy:
@sarahscott I started a rough list of places we may want to update the docs for: https://github.com/artsy/emission/wiki/Relay-Modern-API-notes

l2succes:
Awesome, thanks for the feedback and help everyone.
@sarahscott that's a great idea, there's definitely a few code examples that can be added to make it easier.

PR #756, opened 2017-09-15
Add support for submitting a consignment at the end of the process

orta:
Now that https://github.com/artsy/metaphysics/pull/739 is on staging we can submit details to the server, so this starts that process, but doesn't finish it


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

package.json
```diff
@@ -97,6 +97,7 @@
     "recursive-readdir-sync": "^1.0.6",
     "relay-compiler": "https://github.com/alloy/relay/releases/download/v1.3.0-artsy/relay-compiler-1.3.0-artsy.1.tgz",
     "relay2ts": "^0.2.0",
+    "snapshot-diff": "^0.2.1",
```
orta: This is a snapshot matcher that compares the diff between two objects, so if you add a single attribute to an object ( you'll see some of those tests below) the snapshot is much more focused on the real change, vs the full object which is more meaningless

src/lib/Components/Consignments/Screens/FinalSubmissionQuestions.tsx
```diff
@@ -31,9 +32,7 @@ export default class FinalSubmissionQuestions extends React.Component<Props, Con
   }
 
   submitWork = () => {
-    this.props.navigator.push({
-      component: Loader,
-    })
+    this.setState({ state: "SUBMITTED" }, () => this.props.submitFinalSubmission(this.state))
```
orta: So that it can go up to metaphysics

src/lib/Components/Consignments/Screens/Metadata.tsx
```diff
@@ -24,21 +24,21 @@ interface Props extends ViewProperties {
...
+  { name: "Mixed Media", value: "MIXED_MEDIA" },
+  { name: "Performance Art", value: "PERFORMANCE_ART" },
+  { name: "Installation", value: "INSTALLATION" },
+  { name: "Video/Film/Animation", value: "VIDEO_FILM_ANIMATION" },
+  { name: "Architecture", value: "ARCHITECTURE" },
+  { name: "Fashion Design and Wearable Art", value: "FASHION_DESIGN_AND_WEARABLE_ART" },
+  { name: "Jewelry", value: "JEWELRY" },
+  { name: "Design/Decorative Art", value: "DESIGN_DECORATIVE_ART" },
+  { name: "Textile Arts", value: "TEXTILE_ARTS" },
+  { name: "Other", value: "OTHER" },
```
orta: These values are needed as keys to metaphysics also

src/lib/Components/Consignments/Screens/Overview.tsx
```diff
@@ -72,8 +78,12 @@ export default class Info extends React.Component<Props, ConsignmentSetup> {
     }
   }
 
-  submitFinalSubmission = () =>
-    this.props.navigator.push({ component: FinalSubmissionQuestions, passProps: { setup: this.state } })
+  submitFinalSubmission = async (setup: ConsignmentSetup) => {
+    this.setState(setup, async () => {
+      await this.updateMetaphysics()
+      // EXIT or something
```
orta: This is a TBD, likely requires some native work

src/lib/Components/Consignments/Screens/FinalSubmissionQuestions.tsx
```diff
@@ -50,6 +49,9 @@ export default class FinalSubmissionQuestions extends React.Component<Props, Con
   updateSigned = () => this.setState({ signed: !this.state.signed })
   updateCert = () => this.setState({ certificateOfAuth: !this.state.certificateOfAuth })
 
+  updateEditionSize = text => this.setState({ editionInfo: { ...this.state.editionInfo, size: text } })
```
sweir27: Maybe to be explicit (since convection expects an integer for this field) we can rename the argument to `size` or `value`? (#minor). Only because as I was reading this, `text` suggests it would be a `string` input 😄

orta: Ah, looks like this is wrong in metaphysics then, as it expects a string

sweir27: Oh shucks I thought we caught that 😞

orta: Shipped a PR to metaphysics for it: https://github.com/artsy/metaphysics/pull/767

orta:
Cool, fixed the minors 👍

Issue #625, opened 2017-06-19
[MSG] Network connectivity states in Composer


sarahscott:
UI + Relay handling of:
- [x] Successful message send
- [x] Unsuccessful send
- [x] Unsuccessful fetch
- [ ] ~~Local storage~~ We decided not to make this part of the MVP

![](https://cloud.githubusercontent.com/assets/6482926/26464297/46a042c0-4155-11e7-8447-9363b520ac44.png)

alloy:
For the MVP we settled on the following. On _any_ failure:

* remove the optimistic message
* restore the text into the textfield

alloy:
@sarahscott The changes in #757 don’t fully implement the requested design in this ticket. Maybe discuss the differences with Katarina and then decide whether to create new tickets for iterations or how else to proceed.

sarahscott:
Doesn't #757 implement the MVP comment above?

alloy:
It sure does! But not the original requirements, so before we lose track of that, only for it to be resurfaced by Katarina during QA, let’s make sure we’re all aware of it _now_.

sarahscott:
Sounds good. I guess for posterity we can say that maintaining failed messages would require some sort of local storage, and that's why we decided on the textfield solution for now.

tayapage6:
Note from standup - if there is one pending message, another cannot be sent (send button disabled)

alloy:
For the MVP we decided to still have the yellow ‘pending’ message.

PR #757, opened 2017-09-18
[msg] Retry state for failed messages

sarahscott:
If a message mutation fails, the message text will show up in the Composer again so the user can try again

Fixes: https://github.com/artsy/emission/issues/625

### DIFF
--- a/src/lib/Components/Inbox/Conversations/Composer.tsx
+++ b/src/lib/Components/Inbox/Conversations/Composer.tsx
@@ -1,6 +1,5 @@
 import * as React from "react"
 import { Keyboard, KeyboardAvoidingView, TextInput, TouchableWithoutFeedback } from "react-native"
-import { RelayRefetchProp } from "react-relay"
 
 import styled from "styled-components/native"
 import colors from "../../../../data/colors"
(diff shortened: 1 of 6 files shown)


src/lib/Containers/Conversation.tsx
```diff
@@ -127,9 +128,10 @@ export class Conversation extends React.Component<Props, State> {
 
     return (
       <Composer
-        disabled={this.state.sendingMessage}
+        disabled={this.state.sendingMessage || !this.state.isConnected}
+        value={this.state.submissionError}
```
alloy: I’m not a big fan of this name, I wonder if it could be something like `failedMessageText` ?

sarahscott:
oh good catch! i totally agree and will update.

mennenia:
@sarahscott looks awesome! Wanna merge the conflicts with master? Then we can 🚢  🎉 !! :D

src/lib/Components/Inbox/Conversations/Composer.tsx
```diff
@@ -74,6 +80,8 @@ export default class Composer extends React.Component<Props, State> {
       paddingRight: 10,
     }
 
+    const disableSendButton = !(this.state.text && this.state.text.length) || this.props.disabled
```
alloy: FYI I’d generally revert this to _first_ check the simple property and only if that doesn’t short-circuit the conditional do some computation.

PR #773, opened 2017-09-28
[react-tracking] Add typings for react-tracking and our schema.

alloy:
This adds the `react-racking` types which are augmented with our schema.

I flip-flopped a lot. In the first commit you can find a version that uses a noop function that has multiple overloads and in the second commit I replaced it with a single interface that the user can use to augment the track function themselves (which means no otherwise unnecessary function calls).

What do you think?

"Skip New Tests"


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

- `src/lib/__tests__/metaphysics-tests.ts`
- `src/lib/utils/__tests__/errors-tests.tsx`
- `src/lib/utils/__tests__/track-tests.ts`

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/utils/track.ts
```diff
@@ -0,0 +1,107 @@
...
+ *          id: string
+ *          slug: string
+ *        }
+ *      }
+ *
+ *      interface State {
+ *        following: boolean
+ *      }
+ *
+ *      const track: Track<Props, State, Schema.Entity> = _track
```
l2succes: Little observation: it looks like the order of the interfaces is different in this example from the following:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-tracking/test/react-tracking-with-types-tests.tsx#L20

Is that intended or I'm missing something?

alloy: Good spot. So that’s because with the generic version from DefinitelyTyped it’s more reasonable to expect schema typings, whereas with this version it already has default schema typings (`Schema.Global`) and so it seemed likely that there will be cases where you want to specify `Props` but don’t need to specify different schema typings.

alloy: What do you think?

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Missing Test Files:

- `src/lib/__tests__/metaphysics-tests.ts`
- `src/lib/utils/__tests__/errors-tests.tsx`
- `src/lib/utils/__tests__/track-tests.ts`

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/lib/utils/track.ts
orta: I think that makes sense ( though we'll want the same thing inside Reaction so that it's consistent between projects )

src/lib/Components/Artist/Header.tsx
```diff
@@ -16,17 +16,18 @@ import SerifText from "../Text/Serif"
...
-}
+// tslint:disable-next-line:no-empty-interface
+interface Props extends RelayProps {}
 
 interface State {
   following: boolean
   followersCount: number
 }
 
+const track: Track<Props, State, Schema.Entity> = _track
```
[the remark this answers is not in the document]
orta: did you consider an exposed HOC with those three generic params that passes it into a blank Component?

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Missing Test Files:

- `src/lib/__tests__/metaphysics-tests.ts`
- `src/lib/utils/__tests__/errors-tests.tsx`
- `src/lib/utils/__tests__/track-tests.ts`

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/lib/utils/track.ts
l2succes: I see, I get the reasoning now but I think it'd be better to keep the order the same globally. Otherwise, we might get confused in our own codebase if we look at the DefinitelyTyped definitions as a reference.

I personally I prefer the version with `Props` first as it's easier to remember and it follows `Component<Props, State>` but either is fine I think.

src/lib/Components/Artist/Header.tsx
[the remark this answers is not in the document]
alloy: Can you give a pseudo code example?

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Missing Test Files:

- `src/lib/__tests__/metaphysics-tests.ts`
- `src/lib/utils/__tests__/errors-tests.tsx`
- `src/lib/utils/__tests__/track-tests.ts`

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/lib/utils/track.ts
alloy: I agree that having the same signature would simplify the understanding a bit, but I also do think the different signatures make sense in their own contexts. I would suggest we give it a try and see if people actually run into issues first, unless others have a preference as well?

alloy: @orta Aye, I’ll port this back to Emission once a final decision has been taken here.

src/lib/utils/track.ts
```diff
@@ -0,0 +1,107 @@
...
+ *          entity_slug: props.artist.slug
+ *        }))
+ *        handleFollow() {
+ *          // ...
+ *        }
+ *      }
+ *
+ *      ```
+ */
+export interface Track<P = any, S = any, T extends Schema.Global = Schema.Global> extends _Track<T, P, S> {} // tslint:disable-line:no-empty-interface
```
mennenia: I'm confused here (not sure if this is what Luc was referring to above as well)...

So `Track` takes `P, S, T`, but `_Track` takes `T, P, S`?

alloy: lol, yup, you’re correct and that’s what @l2succes was asking about.

src/lib/utils/track.ts
alloy: To summarise:

* The generic react-tracking types in DefinitelyTyped take types in the following order, because without any default types it seems more likely that you’ll specify a schema over props and state:
  1. `Schema`, default `any`
  2. `Props`, default `any`
  3. `State`, default `any`
* In our specific types the order is different, because we have a default schema and it seems more likely that you’ll want to specify props or state _without_ having to specify the default schema types:
  1. `Props`, default `any`
  2. `State`, default `any`
  3. `Schema`, default `Schema.Global`

alloy: I would add that type parameters aren’t really different than other API parameters, and we change order and provide more specific defaults all the time. I know that this all still requires getting used to reading typings, but once comfortable with it it shouldn’t really read differently than any other API would do. (I mean, ideally :D)

src/lib/Components/Artist/Header.tsx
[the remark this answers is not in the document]
mennenia: Any further discussion here or shall we merge? @orta @alloy

src/lib/utils/track.ts
mennenia: 👍 I think it makes sense. Just had to take a sec to make sure haha 😄

mennenia:
🙌

src/lib/Components/Artist/Header.tsx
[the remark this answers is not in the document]
alloy: In case it wasn’t obvious, the first commit used a function call to offer various type overloads and cast the result, whereas the idea of this second commit was to get rid of noop function calls completely.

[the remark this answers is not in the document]
orta: Let's merge, it'd only be a refactor anyway

PR #778, opened 2017-10-02
[Dev] Updates RN to 0.48.4

orta:
Mainly because I couldn't use node 8.x with 0.48 due https://github.com/facebook/jest/issues/4433

I made a compare link: https://github.com/facebook/react-native/compare/v0.48.0...0.48-stable

and it contained something interesting: https://github.com/facebook/react-native/commit/9c90ba6b0bcd01bc19efc68c65ac7637c479ff57


orta:
Hrm, this should be considered blocked due to https://github.com/facebook/react-native/issues/15791#issuecomment-327004744

alloy:
Gah. Maybe add a `sed` script to disable that validation in `postinstall`?

alloy:
This shouldn’t be a blocker for a release, right? As there won’t be JS over the network.

orta:
Ah yeah, I can fix that with a sed

orta:
I commented it out in mine and it works fine, I also did send off a nebula release

alloy:
Are you going to add the sed script in this PR?

orta:
Yes, it was trickier than I thought, pushing now

orta:
OK, that should be it 👍

orta:
I've also ensured that storybooks JS file goes to the right place, add added back the bottom VCs

package.json
```diff
@@ -34,7 +34,7 @@
     "sync-schema": "cd externals/metaphysics && git fetch && git checkout origin/master && yarn install && npm run dump-schema -- ../../data && rm -rf node_modules",
     "version": "yarn install && yarn bundle && pushd Example && bundle install && bundle exec pod install && popd && git add Pod/Assets Example/Podfile.lock",
     "postversion": "git push --follow-tags && env EMISSION_ROOT=\"$(pwd)\" pod repo push artsy --allow-warnings --use-json --skip-import-validation",
-    "postinstall": "sed -i '' 's/#import <RCTAnimation\\/RCTValueAnimatedNode.h>/#import \"RCTValueAnimatedNode.h\"/' ./node_modules/react-native/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.h || true",
+    "postinstall": "sed -i '' 's/#import <RCTAnimation\\/RCTValueAnimatedNode.h>/#import \"RCTValueAnimatedNode.h\"/' ./node_modules/react-native/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.h || ./scripts/post_installs.sh || true",
```
alloy: Won’t this only run if the first part (before `||`) fails?

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.ios.js",
   "scripts": {
     "bundle": "haul bundle --platform=ios --dev=false --entry-file=index.ios.js --bundle-output Pod/Assets/Emission.js --sourcemap-output Pod/Assets/Emission.js.map --assets-dest Pod/Assets && rm -rf Pod/Assets/assets/node_modules",
-    "bundle-with-storybooks": "haul bundle --platform=ios --dev=true --entry-file=Example/Emission/index.ios.js --bundle-output Emission.js --assets-dest Pod/Assets",
+    "bundle-with-storybooks": "haul bundle --platform=ios --dev=true --entry-file=Example/Emission/index.ios.js --bundle-output  Pod/Assets/Emission.js --assets-dest Pod/Assets",
```
alloy: Ah I see. The problem was that Xcode wasn’t including the js bundle inside the app bundle 👍

scripts/post_installs.sh
```diff
@@ -0,0 +1,4 @@
+# Comes from https://github.com/artsy/emission/pull/778
+touch tmp.file
+awk '/returned javascript/ {f=15} f && f-- {$0="//"$0}1' node_modules/react-native/React/Base/RCTJavaScriptLoader.mm > tmp.file
+mv tmp.file node_modules/react-native/React/Base/RCTJavaScriptLoader.mm
```
alloy: Nice 🕺

package.json
orta: oh yeah!

orta:
The blocker here is dependency CI which is fine, going to self-merge

PR #779, opened 2017-10-03
Store Submission State 

orta:
Blocked by #756

Uses the react-native [AsyncStorage](https://facebook.github.io/react-native/docs/asyncstorage.html) key value store to store and restore a user's consignment session.

Touches on https://github.com/artsy/consignments/issues/70 but fixes https://github.com/artsy/consignments/issues/84


.vscode/settings.json
```diff
@@ -1,78 +1,67 @@
...
+  "files.exclude": {
+    "**/.git": true,
+    "**/.svn": true,
+    "**/.hg": true,
+    "**/.DS_Store": true,
+    ".jest/": true,
+    "build": true,
+    ".awcache": true,
+    ".happypack": true,
+    "Example/build": true,
```
orta:
```diff
+    ".awcache": true,
+    ".happypack": true,
```

is the real change here, but I removed the comments so prettier can handle the JSON formatting

src/lib/Components/Consignments/Screens/Overview.tsx
```diff
@@ -123,7 +139,7 @@ export default class Info extends React.Component<Props, ConsignmentSetup> {
 
             <Row style={{ justifyContent: "center" }}>
               <View style={{ height: 43, width: 320, marginTop: 20, opacity: canSubmit ? 1 : 0.3 }}>
-                <Button text="NEXT" onPress={canSubmit && this.goToFinalSubmission} style={{ flex: 1 }} />
+                <Button text="NEXT" onPress={canSubmit ? this.goToFinalSubmission : emptyFunc} style={{ flex: 1 }} />
```
orta: This was causing runtime warnings

src/lib/Components/Consignments/__stories__/Consignments.story.tsx
```diff
@@ -31,7 +31,8 @@ const blankMetadata = {
 storiesOf("Consignments/_Screens")
   .add("Consignments", () => <Nav />)
   .add("Welcome Page", () => <Welcome navigator={nav} route={route} />)
-  .add("Overview Page", () => <Overview navigator={nav} route={route} setup={{}} />)
+  .add("Overview Page (Fresh)", () => <Overview navigator={nav} route={route} setup={{}} />)
+  .add("Overview Page", () => <Overview navigator={nav} route={route} setup={null} />)
```
orta: This one will always be blank, and the other will always try restore the last version

src/lib/Components/Consignments/Screens/Overview.tsx
```diff
@@ -67,9 +79,11 @@ export default class Info extends React.Component<Props, ConsignmentSetup> {
   updateLocation = (city: string, state: string, country: string) =>
     this.updateStateAndMetaphysics({ location: { city, state, country } })
 
-  updateStateAndMetaphysics = (state: any) => this.setState(state, this.updateMetaphysics)
+  updateStateAndMetaphysics = (state: any) => this.setState(state, this.updateLocalStateAndMetaphysics)
+
+  updateLocalStateAndMetaphysics = async () => {
+    this.saveStateToLocalStorage()
```
xtina-starr: The method says it updates State and Metaphysics, but it seems it just set the item on the `AsyncStorage` (update the local store). I'm mainly curious if it should also be updating metaphysics (or if it does some where and I missed that). If it is indeed just updating the local store do it does it need to be called in the `submitFinalSubmission` on like 97?

orta: The lines underneath ( 73-77 ) update Metaphysics (`updateSubmission` and `createSubmission`) - they're still around 👍

xtina-starr: gotcha!

xtina-starr:
👍

PR #781, opened 2017-10-04
[Dev] Adds a lock for the node/yarn/npm versions which match metaphysics

orta:
screen shot 2017-10-04 at 11 22 25 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,11 @@
   "name": "emission",
   "version": "1.4.0-beta.4",
   "description": "Artsy React(Native) components.",
+  "engines": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -2,6 +2,11 @@
   "name": "emission",
   "version": "1.4.0-beta.4",
   "description": "Artsy React(Native) components.",
+  "engines": {
+    "node": "8.4.x",
+    "npm": "5.4.x",
+    "yarn": "1.x"
+  },
```
mennenia: omg YAS 🙌

PR #775, opened 2017-10-01
[packager] Fix release builds by switching from Metro to Haul.

alloy:
* https://github.com/callstack/haul
* Looks like we’ll loose a little packaging performance, but not too much and probably does not outweigh the amount of time we spent during upgrades to get our setup to work http://blog.nparashuram.com/2017/04/react-native-packager-vs-haul.html#results
* TODO We need to make automatically launching a custom debugger (React Native Debugger.app) work, someone should PR that to the Haul repo.

Skip New Tests

PR #785, opened 2017-10-06
[Dev] switch back to RNP

### LINKED ISSUE #776: TODOs related to switch to Haul
- [ ] Make Haul open a custom debugger (React Native Debugger.app in our case)
- [ ] Make Haul’s Babel usage compatible with Babel 7 by making this `ignore` value an array: https://github.com/callstack/haul/blob/9bca526d050de98423b4a56a087ee1cc3d7666ea/bin/cli.js#L4 (Need to check if Babel 6 also accepts an array as the value for compatibility.)
- [ ] Try the [relay-compiler-webpack](https://github.com/danielholmes/relay-compiler-webpack-plugin) plugin, rather than using a standalone tool.

orta:
Hand reverts #777  / #775  and fixes #776 

I spent 2 days using this build, and found development speed to be a bit too slow to be worth the value we'd get from the consolidation on WebPack ( performance was on the same order of magnitude as seen in Reaction https://github.com/artsy/reaction/issues/211 ) vs the defaults inside RNP/Metro

A live-reload would take 11s on average, and a HMR would take 5s. So, was worth the experiment, but switching back for the moment. Would recommend opening #775 to review.


alloy:
> Would recommend opening #775 to review.

What do you mean by this?

orta:
ah, mainly so that you can compare the before/after

PR #798, opened 2017-10-24
[dev] v1.3.12 release for scrollview bug

sarahscott:
Reverts scroll hack commits from a few months ago - related to https://github.com/artsy/eigen/issues/2439


alloy:
@sarahscott These changes also need to be ported to `master`, otherwise we’d re-introduce them in the next major release.

Issue #801, opened 2017-10-26
Convert betas to be permanently up-to-date


orta:
Instead use the s3 technique for PRs, but instead post a "master.js" and include some metadata (date / last merged PR )

orta:
This is in.

PR #803, opened 2017-10-26
Revise deployment and JS strategy 

orta:
Re: https://github.com/artsy/emission/issues/801

This makes the default experience always master on Emission. We're rarely making changes to the native bridge nowadays, and I think it's a pretty safe option. 

This means that when you open the app, it shows a spinner and tells you that it's downloading a new build (and says what the last PR merged was) then you're in the app like normal. 

It removes apphub, no-one cares about older commits TBH, and they are all being stored on S3 anyway (as PRs) so it's trivial to add a generic load closed PR interface at some point.


Example/Emission/LoadingSpinner.m
```diff
@@ -8,13 +10,37 @@ @interface LoadingSpinner()
...
+    if (subtitle) {
+      UILabel *subtitleLabel = [[UILabel alloc] init];
+      subtitleLabel.text = subtitle;
+      subtitleLabel.font = [UIFont serifFontWithSize:18];
+      subtitleLabel.textAlignment = NSTextAlignmentCenter;
+      [spinnerView addSubview:subtitleLabel];
+      [subtitleLabel alignLeading:@"0" trailing:@"0" toView:spinnerView];
+      [subtitleLabel alignCenterYWithView:spinnerView predicate:@"-100"];
+    }
+
```
orta: Looks like: 
screen shot 2017-10-27 at 11 15 31 am

PR #805, opened 2017-10-27
Just add the deploy master metadata

orta:
Re: #803 

It's  _real_ hard to work without some real metadata, so this is just the S3 uploading part of #803 - so that it can go through the full lifecycle ( check it out in travis on build 8.4.0 on the master build once this is merged)


orta:
https://travis-ci.org/artsy/emission/jobs/293740948 was the CI run

PR #813, opened 2017-11-01
[Crash] guard against conversations_existence_check being null

xtina-starr:
**The error message caused by `this.props.me.conversations_existence_check` being `null`:**
simulator screen shot - iphone 6 - 2017-11-01 at 17 11 31

**After the null check:**
simulator screen shot - iphone 6 - 2017-11-01 at 17 10 14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@
 
 ###### Messaging
 
+-   Fix crash by guarding against null `conversations_existence_check` - christina
 -   Fix crash of empty node in conversation list - maxim
 -   Hyperlinks in messages are now artsy purple & underlined - sarah
 -   Aligns message header with existing back button - sarah
--- a/src/lib/Containers/Inbox.tsx
+++ b/src/lib/Containers/Inbox.tsx
@@ -46,7 +46,8 @@ export class Inbox extends React.Component<Props, State> {
 
   render() {
     const hasBids = this.props.me.lot_standings.length > 0
-    const hasConversations = this.props.me.conversations_existence_check.edges.length > 0
+    const hasConversations =
+      this.props.me.conversations_existence_check && this.props.me.conversations_existence_check.edges.length > 0
     return hasBids || hasConversations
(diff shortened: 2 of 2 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

orta:
Awesome work - CHANGELOG please 👍

mzikherman:
Ah, sorry we discussed offline:

This is a temporary issue due to a switch in some authentication thats out on staging currently. While there's nothing wrong with being extra defensive (IMO), we should understand and reason about the data we get back and only be defensive on _actual_ scenarios, so we'll close for now.

Although...I suppose being defensive, even in a total red herring scenario thats _not_ expected, is probably better than an outright crash, no? In that case maybe we should merge this in!

Reopening!

mennenia:
So that's what I was wondering... I think it's always worth guarding against a crash, though @xtina-starr rightly mentioned... maybe it _should_ actually crash if something is truly amiss?

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

mzikherman:
Let's CHANGELOG and merge!

@mennenia I'd think in that case, we'd want to log/be notified of it, but I think gracefully handling it for the end user is 👌

orta:
ideally the metaphysics schema should tell us if this can be nil ( but we don't [have strict checks on](https://ariya.io/2016/10/typescript-2-and-strict-null-checking) ) so more checking is better (to an extent )

orta:
👍 works for me

PR #825, opened 2017-11-16
[QA] Inquiry Screen

mennenia:
Fix #[CE635](https://github.com/artsy/collector-experience/issues/635)

Phone:
simulator screen shot - iphone 6 - 9 1 - 2017-11-16 at 10 13 29

Pad:
simulator screen shot - ipad pro 12 9 inch - 2017-11-16 at 10 13 34

### DIFF
--- a/src/lib/Containers/Inquiry.tsx
+++ b/src/lib/Containers/Inquiry.tsx
@@ -4,6 +4,7 @@ import { createFragmentContainer, graphql } from "react-relay"
 import { MetadataText, SmallHeadline } from "../Components/Inbox/Typography"
 
 import {
+  Dimensions,
   FlatList,
   ImageURISource,
   KeyboardAvoidingView,
@@ -24,6 +25,8 @@ import ArtworkPreview from "../Components/Inbox/Conversations/Preview/ArtworkPre
 import ARSwitchBoard from "../NativeModules/SwitchBoard"
 import { gravityURL } from "../relay/config"
 import { NetworkError } from "../utils/errors"
+
+const isPad = Dimensions.get("window").width > 700
 
 const Container = styled.View`
   flex: 1;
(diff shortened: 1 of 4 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Containers/Inquiry.tsx
```diff
@@ -55,6 +61,8 @@ const CancelButton = styled.TouchableOpacity`padding-left: 20;`
 const Content = styled.View`
   margin-left: 20;
   margin-right: 20;
+  alignSelf: ${isPad ? "center" : "stretch"};
+  ${isPad ? "width: 472;" : ""};
```
alloy: It’s unfortunate that this is laid out at 480px on iPone, otherwise we could have just gone with always setting `center` and `max-width: 472` :/

mennenia: I know :( it's just too wide on the phone. 

When I center and add a max width, it just stays on that width, no longer shrinking to fit the screen. I tried it with a `min-width` also, but to no avail.

alloy: Do you mean that it would be ok as per the design to use the same max width value for both iPhone and iPad?

mennenia: No, I had hoped I could put in a min and a max, such that it would stretch to fill appropriately on the devices, within those two bounds. However, you can't then also `center` the view. (the entire inquiry screen will be left aligned on iPad) 

Basically, all I meant to say is that it is indeed a bit ugly and unfortunate 😂 I tried a bunch of things, and this was the only solution that gave the desired result.

alloy: That’s fair 👍

PR #827, opened 2017-11-16
[MSG] Update zero state inbox for ipad

mennenia:
Fixes #[CE715](https://github.com/artsy/collector-experience/issues/715)

Phone

simulator screen shot - iphone 6 - 9 1 - 2017-11-29 at 13 48 02

Pad P

simulator screen shot - ipad pro 12 9 inch - 2017-11-29 at 12 55 37

Pad L

simulator screen shot - ipad pro 12 9 inch - 2017-11-29 at 12 55 39


src/lib/Components/Inbox/Conversations/ZeroStateInbox.tsx
```diff
@@ -53,21 +60,23 @@ export default () =>
...
-      </HorizontalLayout>
-      <HorizontalLayout>
-        <Icon source={require("../../../../../images/message.png")} />
-        <SmallHeadline>Find your ongoing conversations and bidding activity here</SmallHeadline>
-      </HorizontalLayout>
-      <HorizontalLayout>
-        <Icon source={require("../../../../../images/pay.png")} />
-        <SmallHeadline>Easily process payment through our secure platform</SmallHeadline>
-      </HorizontalLayout>
+      <List>
```
mennenia: Anyone know if I can make this element conditional? if I remove the `<List>` and `</List>` further down, iPhone works perfectly, but iPad breaks. And vice versa. Even if I empty the `List` styled component, iPhone still breaks. It just doesn't like the extra wrapper view there. 

cc @alloy @orta

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/Inbox/Conversations/ZeroStateInbox.tsx
alloy: I don’t really know the answer to that (maybe track down the original author of that code and ask them), but I do think it _should_ be possible to have a wrapper view that doesn’t really do anything.

mennenia: That person is me 😄 

And FYI, I just added the `List`, in order to support iPad (it wasn't there before). 

I agree that it shouldn't make a difference, but if I remove all styling (not even commenting out), it does change. Would love to debug, might take a few mins of our 1-1 for that.

alloy: I guess it might have to do with default styling that it has? i.e. maybe you can’t leave it empty but do have to override some styling

katarinabatina:
@orta did your header adjustments that we did on iphone make it into ipad?

src/lib/Components/Inbox/Conversations/ZeroStateInbox.tsx
```diff
@@ -48,25 +55,59 @@ const Icon = styled(Image)`
...
-  </CenteredView>
+export default class ZeroStateInbox extends React.Component<null, null> {
+  render() {
+    /** TODO: Maxim see if there's any other way  */
+    if (isPad) {
+      return (
+        <CenteredView>
+          <VerticalLayout>
+            <Title>BUYING ART ON ARTSY IS SIMPLE</Title>
+            <List>
```
orta: This is the only difference? you could do something like: `const Wrapper = isPad ? List : View` then use that inside this render function

mennenia: the thing is that `View` causes iPhone to be rendered differently. It should be `List` or _nothing_

mennenia: Ok just had another idea to try which worked. BRB

mennenia:
@orta done!

orta:
👍

PR #835, opened 2017-11-17
Adds a bunch of native VCs for our exposed components

orta:
This means that the Eigen router can work with these VCs


Pod/Classes/ViewControllers/ARComponentViewController.m
```diff
@@ -82,4 +82,9 @@ - (UIInterfaceOrientationMask)supportedInterfaceOrientations;
   return self.shouldAutorotate ? UIInterfaceOrientationMaskAll : UIInterfaceOrientationMaskPortrait;
 }
 
+- (UIStatusBarStyle)preferredStatusBarStyle
+{
+    return UIStatusBarStyleLightContent;
+}
+
```
orta: As ever VC is a subclass of this, it applies it everywhere

src/lib/relay/QueryRenderers.tsx
```diff
@@ -84,23 +81,6 @@ export const ConversationRenderer: React.SFC<ConversationRendererProps> = ({ ren
...
-            ...Home_home
-          }
-        }
-      `}
-      variables={{}}
-      render={render}
-    />
-  )
-}
-
```
orta: 👋

PR #842, opened 2017-11-20
[Conversation] Don’t let scrollview capture taps.

alloy:
Fixes https://github.com/artsy/collector-experience/issues/651


sarahscott:
nice

PR #853, opened 2017-11-29
More crash related fixes

alloy:
* I’ve removed cases where `setState` is being used in a promise and we subsequently catch all errors that can come out of the promise, which can apparently [leave React in a bad state](https://github.com/facebook/react/issues/8579#issuecomment-267312367). Going forward we should either just `console.error(…)` them so they show up in Sentry or be selective about which errors we catch and rethrow all others.
* The Sentry version we were using [has a bug](https://github.com/getsentry/sentry-cocoa/pull/192). To be able to update Sentry in Eigen we need to also update the RN client.
* The Gene container was crashing when it was initialised without filter defaults, which would happen when Eigen’s switchboard would route a Gene URL that did not contain any of the required filter parameters.
  - This would crash: `https://www.artsy.net/gene/graffiti-and-street-art`
  - This would work: `https://www.artsy.net/gene/graffiti-and-street-art?medium=*&price_range=*-*`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 -   QA for composer/thread view: ipad and fixed iphone size attachments - maxim
 -   QA for message attachments - maxim
 -   QA for home header font sizes and margins - maxim
+-   Don’t rely on user of the Gene VC to specify filter defaults, which could lead to a crash - alloy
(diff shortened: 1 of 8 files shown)


orta:
👍

PR #857, opened 2017-11-30
[inbox] fix for inbox pull-to-refresh

sarahscott:
- [x] Refresh conversations
- [x] Refresh active bids

Before, when the `Inbox` tried to force fetch its child `Conversations` container, we got an error related to pagination because `Conversations` is a `PaginationContainer` and [maintains its own cursor](https://github.com/artsy/emission/blob/master/src/lib/Components/Inbox/Conversations/index.tsx#L125).

I got around this by making `refreshConversations` and `refreshActiveBids` methods that the `Inbox` can call upon pull-to-refresh. The `ActiveBids` container is now a `RefetchContainer` and can be updated with a normal force fetch.

refresh

Skip New Tests

### DIFF
--- a/src/lib/Components/Inbox/ActiveBids/index.tsx
+++ b/src/lib/Components/Inbox/ActiveBids/index.tsx
@@ -1,5 +1,5 @@
 import React from "react"
-import { createFragmentContainer, graphql } from "react-relay"
+import { createRefetchContainer, graphql, RelayRefetchProp } from "react-relay"
 import styled from "styled-components/native"
 
 import { View } from "react-native"
@@ -8,7 +8,25 @@ import ActiveBid from "./ActiveBid"
 
 const Container = styled.View`margin: 20px 0 40px;`
 
-class ActiveBids extends React.Component<RelayProps, null> {
+interface Props extends RelayProps {
+  relay?: RelayRefetchProp
+}
+
+interface State {
+  fetchingData: boolean
+}
+
+class ActiveBids extends React.Component<Props, State> {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      fetchingData: false,
+    }
+
+    this.refreshActiveBids = this.refreshActiveBids.bind(this)
+  }
+
   hasContent() {
     return this.props.me.lot_standings.length > 0
   }
@@ -18,6 +36,27 @@ class ActiveBids extends React.Component<RelayProps, null> {
       return <ActiveBid key={bidData.most_recent_bid.__id} bid={bidData} />
     })
     return bids
+  }
+
+  refreshActiveBids(callback) {
+    if (this.state.fetchingData) {
+      return
+    }
+
+    this.setState({ fetchingData: true })
+
+    this.props.relay.refetch(
+      {},
+      {},
+      () => {
+        this.setState({ fetchingData: false })
+
+        if (callback) {
+          callback()
+        }
+      },
+      { force: true }
+    )
   }
 
   render() {
@@ -30,15 +69,24 @@ class ActiveBids extends React.Component<RelayProps, null> {
   }
 }
 
-export default createFragmentContainer(
+export default createRefetchContainer(
   ActiveBids,
-  graphql`
-    fragment ActiveBids_me on Me {
-      lot_standings(live: true) {
-        most_recent_bid {
-          __id
+  {
+    me: graphql`
+      fragment ActiveBids_me on Me {
+        lot_standings(live: true) {
+          most_recent_bid {
+            __id
+          }
+          ...ActiveBid_bid
         }
-        ...ActiveBid_bid
+      }
+    `,
+  },
+  graphql`
+    query ActiveBidsRefetchQuery {
+      me {
+        ...ActiveBids_me
       }
     }
   `
--- a/src/lib/Components/Inbox/Conversations/index.tsx
+++ b/src/lib/Components/Inbox/Conversations/index.tsx
@@ -66,7 +66,10 @@ export class Conversations extends React.Component<Props, State> {
     this.setState({ fetchingNextPage: true })
     this.props.relay.refetchConnection(10, () => {
       this.setState({ fetchingNextPage: false })
-      callback()
+
+      if (callback) {
+        callback()
+      }
     })
   }
 
--- a/src/lib/Components/Inbox/Conversations/index.tsx
+++ b/src/lib/Components/Inbox/Conversations/index.tsx
@@ -40,6 +40,8 @@ export class Conversations extends React.Component<Props, State> {
       dataSource,
       fetchingNextPage: false,
     }
+
+    this.refreshConversations = this.refreshConversations.bind(this)
   }
 
   componentDidMount() {
@@ -53,6 +55,18 @@ export class Conversations extends React.Component<Props, State> {
 
     this.setState({
       dataSource: this.state.dataSource.cloneWithRows(conversations),
+    })
+  }
+
+  refreshConversations(callback) {
+    if (this.state.fetchingNextPage) {
+      return
+    }
+
+    this.setState({ fetchingNextPage: true })
+    this.props.relay.refetchConnection(10, () => {
+      this.setState({ fetchingNextPage: false })
+      callback()
     })
   }
 
--- a/src/lib/Containers/Inbox.tsx
+++ b/src/lib/Containers/Inbox.tsx
@@ -19,6 +19,7 @@ const Container = styled.ScrollView`flex: 1;`
 
 export class Inbox extends React.Component<Props, State> {
   conversations: any
+  activeBids: any
 
   constructor(props) {
     super(props)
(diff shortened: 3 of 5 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Containers/Inbox.tsx
```diff
@@ -34,14 +36,9 @@ export class Inbox extends React.Component<Props, State> {
...
     this.setState({ fetchingData: true })
-    this.props.relay.refetch(
-      {},
-      null,
-      () => {
-        this.setState({ fetchingData: false })
-      },
-      { force: true }
-    )
+    this.conversations._refetchConnection(10, () => {
```
alloy: This appears to be a private method. Is this not usable in a public way?

src/lib/Containers/Inbox.tsx
```diff
@@ -34,14 +36,14 @@ export class Inbox extends React.Component<Props, State> {
...
-        this.setState({ fetchingData: false })
-      },
-      { force: true }
-    )
+
+    // Force-fetch self; this will update Active Bids but not Conversations (there will be a warning)
+    this.props.relay.refetch({}, {}, null, { force: true })
+
+    // Allow Conversations to properly force-fetch itself
+    this.conversations.refs.component.refreshConversations(() => {
```
alloy: Where does the `refs.component` part come from?

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/Inbox/Conversations/index.tsx
```diff
@@ -56,6 +58,18 @@ export class Conversations extends React.Component<Props, State> {
     })
   }
 
+  refreshConversations(callback) {
+    if (this.state.fetchingNextPage) {
+      return
+    }
+
+    this.setState({ fetchingNextPage: true })
+    this.props.relay.refetchConnection(10, () => {
```
alloy: What does this effectively do? Does it reset the connection to only load the initial 10 (after previously have paginated in more pages)?

src/lib/Containers/Inbox.tsx
sarahscott: It seems to be the only way to get to the actual `Conversations` component because it's wrapped in the `PaginationContainer`. This was kind of difficult to figure out because we used to be able to access child methods from a ref directly.

sarahscott: This is what `this.conversations` looks like:
screen shot 2017-12-04 at 2 35 17 pm

src/lib/Components/Inbox/Conversations/index.tsx
sarahscott: What I can gather from the [docs](https://facebook.github.io/relay/docs/en/pagination-container.html) is that it refetches the 10 newest items in the connection, and that it's the [recommended way to force-fetch `PaginationContainer`s](https://facebook.github.io/relay/docs/en/api-cheatsheet.html#to-force-fetch-a-component) (as opposed to `loadMore`).

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/Inbox/Conversations/index.tsx
alloy: Ok.

src/lib/Containers/Inbox.tsx
alloy: Ah right, the first component ref is the Relay container and then from there we get our actual component. Can you add a comment that explains that?

src/lib/Containers/Inbox.tsx
```diff
@@ -34,14 +36,15 @@ export class Inbox extends React.Component<Props, State> {
...
-        this.setState({ fetchingData: false })
-      },
-      { force: true }
-    )
+
+    // Force-fetch self; this will update Active Bids but not Conversations (there will be a warning)
+    this.props.relay.refetch({}, {}, null, { force: true })
+
+    // Allow Conversations to properly force-fetch itself
+    console.log(this.conversations)
```
l2succes: `console.log`

src/lib/Containers/Inbox.tsx
```diff
@@ -34,14 +36,15 @@ export class Inbox extends React.Component<Props, State> {
...
-      {},
-      null,
-      () => {
-        this.setState({ fetchingData: false })
-      },
-      { force: true }
-    )
+
+    // Force-fetch self; this will update Active Bids but not Conversations (there will be a warning)
+    this.props.relay.refetch({}, {}, null, { force: true })
```
l2succes: Have you considered calling `refetch` on the `<ActiveBids />` component directly? You would just need to change the `createFragmentContainer` to `createRefreshContainer` in the line below. I reckon that would fix the warning as well

https://github.com/artsy/emission/blob/16677639790fdd61afa3cb14f48d19915e7741af/src/lib/Components/Inbox/ActiveBids/index.tsx#L33

sarahscott: @l2succes that would get rid of the warning (it's caused by Conversations when Inbox does the force fetch). Is there a reason why the Active Bids were built as a fragment container? I don't mind changing it to a RefetchContainer but I'm curious as to whether that would have any other implications.

sarahscott: I was also trying to avoid some complexity because the `_refetch` relay methods are private, and they can be called by the parent but as @alloy mentioned, it's not ideal. So I would have to make a public `refetch` method for the active bids as well. Not a ton of work, but not as simple 🤷🏻‍♀️.

src/lib/Containers/Inbox.tsx
sarahscott: good catch 😄

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Missing Test Files:

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/lib/Containers/Inbox.tsx
l2succes: @sarahscott Right, originally the Inbox container was refreshing the data for both active bids and conversations so the `<ActiveBids />` didn't need to be a refresh container. 
The active bids component isn't being used anywhere else but in Inbox so I don't think it'll have any repercussions if we change it.

l2succes: as far as the private relay methods goes, as you said it's a bit of a dance but you can add a wrapper method on the active bids component class and call it from here.

src/lib/Containers/Inbox.tsx
l2succes: 😉

src/lib/Containers/Inbox.tsx
damassi: Curious -- what's the reason behind `_refetch` being private? Is there a public representation thats tied into some kind of data / list / container condition as opposed to being given a function to manually trigger?

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Missing Test Files:

- `src/lib/Components/Inbox/ActiveBids/__tests__/index-tests.tsx`

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

sarahscott:
Ready for review!

src/lib/Containers/Inbox.tsx
sarahscott: @damassi my theory is that relay is set up for child containers to update themselves as much as possible given prop or state changes - maybe it isn't ideal for parent containers to be calling child relay methods directly.

src/lib/Components/Inbox/ActiveBids/index.tsx
```diff
@@ -20,6 +38,27 @@ class ActiveBids extends React.Component<RelayProps, null> {
     return bids
   }
 
+  refreshActiveBids(callback) {
+    if (this.state.fetchingData) {
+      return
+    }
+
+    this.setState({ fetchingData: true })
```
alloy: I don’t believe any of this type of state is still necessary, you can probably just check `this.props.relay.isLoading()`

PR #876, opened 2017-12-06
[Inbox] reload data on re-entering tab

l2succes:
Fixes: https://github.com/artsy/collector-experience/issues/794

Depends on: https://github.com/artsy/eigen/pull/2479


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Containers/Inbox.tsx
```diff
@@ -31,6 +32,12 @@ export class Inbox extends React.Component<Props, State> {
     this.fetchData = this.fetchData.bind(this)
   }
 
+  componentWillReceiveProps(newProps) {
```
mennenia: So just checking I understand it right: 

Whenever you navigate to the messaging inbox, the property `isVisible` is set in the props, which means the props are updated, and `componentWillReceiveProps` is triggered, upon which any new data can be fetched? And then _just_ the new messages will be fetched and rendered/added as opposed to the whole VC loaded fresh every time (which we do now and is what is generally being fixed here).

mennenia:
Ahhh 😍😍😍  so coooool

src/lib/Containers/Inbox.tsx
l2succes: Yes, that's precisely what happens 😊.

mennenia: 🎉

PR #880, opened 2017-12-11
Adds nil protection to a check for the length of images to upload to s3

orta:
Resolves: https://sentry.io/artsynet/eigen-staging/issues/413385206/

This was due to our nil checking not being enforced in Emission as `toUpload` could be an falsey value.

#trivial


mennenia:
:+1: woop

PR #862, opened 2017-12-01
[Home/Auctions] Lots By Artists You Follow > II 

damassi:
Follow up to https://github.com/artsy/emission/pull/851. 

(Comments inline)

Skip New Tests

PR #882, opened 2017-12-11
[Home] Adds some null protection to the home / auctions pages

orta:
In looking at https://sentry.io/artsynet/eigen-staging/issues/406930867/ we had wondered if it was related to an expired access token. TBH, I don't think it does right now, (and our access tokens are 25 years by default, you'd need to change your password) but this at least exposes a bunch of easy to grab nil crashes bugs.

#trivial

### DIFF
--- a/src/lib/Containers/WorksForYou.tsx
+++ b/src/lib/Containers/WorksForYou.tsx
@@ -42,7 +42,8 @@ export class WorksForYou extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props)
 
-    const notifications = this.props.viewer.me.notifications.edges.map(edge => edge.node)
+    const notifications =
+      (this.props.viewer.me.notifications && this.props.viewer.me.notifications.edges.map(edge => edge.node)) || []
     if (this.props.viewer.selectedArtist) {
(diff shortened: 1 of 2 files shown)


damassi:
👍 Yeah I had removed this guard [from here](https://github.com/artsy/emission/pull/862/files#diff-de60aba6af766c37ea0f39eaca08d720R30) in a previous PR assuming that metaphysics would, no matter what, return an empty array from `edges`. Was I wrong?

orta:
You're not _wrong, wrong_ - it's more that the related _me_ request could fail to resolve if you have no access token - so then everything is a bit broke :D

It's a trade-off on whether these should come in, but it does mean that all root screens won't crash if you  happen to change your password. Might assign to @alloy to have final say

alloy:
I’m not fully following what issues you were able to reproduce and how, could you elaborate on that? Without that info it’s kinda hard to judge the code.

alloy:
Seeing as the MP schema changed and the merge conflict needs to be resolved, I’d also like us to take some time to re-test that it all works as expected after doing that. As such, I’ll leave this one for 4.0.1

PR #886, opened 2017-12-12
[home] use an enum instead of ints #trivial

sarahscott:
Forgot to do this in my last PR but it's a bit nicer.

Skip New Tests.


sarahscott:
merging

PR #891, opened 2017-12-13
[Gene] Pass defaults for medium/price to native refine hook

orta:
I think a React Native update meant that a value being `undefined` instead of being sent as `null` does not get passed to the native code at all. This means that checks in Eigen that the data is correct would fail. So now if we have undefined, we send the default values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ### Master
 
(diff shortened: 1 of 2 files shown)


orta:
Re: https://github.com/artsy/collector-experience/issues/744

Issue #869, opened 2017-12-03
[RFC] Change "Skip New Tests" to #skip_new_tests


orta:
Mainly so that it is the same as #trivial

PR #894, opened 2017-12-14
Adds support for a native code versioning system

orta:
Adds a native code versioning system, so that we can know when we've broken native compatibility with the JS. Right now this is passed through both the Podspec and via the [master JSON file](https://s3.amazonaws.com/artsy-emission-js/master-metadata.json) 

Also fixes https://github.com/artsy/emission/issues/869 - given that no-one responded I figured it was time to migrate them to be consistent.

### DIFF
--- a/scripts/deploy_master.sh
+++ b/scripts/deploy_master.sh
@@ -22,8 +22,11 @@ PR_NUM=`git log --format=%B -n 1 $SHA | grep -Eo '#[0-9]+' | tail -n 1 | cut -d 
 # format hardcoded because https://stackoverflow.com/questions/7216358/date-command-on-os-x-doesnt-have-iso-8601-i-option
 DATE=`date -u +"%Y-%m-%dT%H:%M:%SZ"`
 
+# Pulls the native code version from the package.json
+NATIVE_VERSION=`ruby -e "require'json';puts(JSON.parse(File.read('package.json'))['native-code-version'])"`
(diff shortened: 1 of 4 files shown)


scripts/deploy_master.sh
```diff
@@ -22,8 +22,11 @@ PR_NUM=`git log --format=%B -n 1 $SHA | grep -Eo '#[0-9]+' | tail -n 1 | cut -d
 # format hardcoded because https://stackoverflow.com/questions/7216358/date-command-on-os-x-doesnt-have-iso-8601-i-option
 DATE=`date -u +"%Y-%m-%dT%H:%M:%SZ"`
 
+# Pulls the native code version from the package.json
+NATIVE_VERSION=`python -c "import json;print json.loads(open('package.json', 'r').read())['native-code-version']"`
```
alloy: 🐍????

We already use Ruby in this repo, can we limit to that?

orta: alright, alright

orta: its now ruby 🍡

orta:
This adds an extra `,` to the JSON

alloy:
Is that a good or a bad thing?

orta:
Bad, it won’t parse - will fix in AM

alloy:
Ah yes, I see

PR #900, opened 2017-12-15
[Messaging] Don't crash when you're on the zerostate

orta:
fixes https://github.com/artsy/collector-experience/issues/818

Checks to see if the refs exist, and if not, will call a relay refetch. TBH I'm like 90% on the fact that this works as I think it will, so will poke @alloy but should be fine regardless.

#trivial

### DIFF
--- a/src/lib/Containers/Inbox.tsx
+++ b/src/lib/Containers/Inbox.tsx
@@ -47,12 +47,16 @@ export class Inbox extends React.Component<Props, State> {
 
     this.setState({ fetchingData: true })
 
-    // Allow Conversations & Active Bids to properly force-fetch themselves.
-    // The stored refs are the Relay containers; the components themselves are nested under as refs.
-    this.activeBids.refs.component.refreshActiveBids()
-    this.conversations.refs.component.refreshConversations(() => {
-      this.setState({ fetchingData: false })
-    })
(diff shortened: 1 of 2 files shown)


src/lib/Containers/Inbox.tsx
```diff
@@ -47,12 +47,16 @@ export class Inbox extends React.Component<Props, State> {
...
-    // The stored refs are the Relay containers; the components themselves are nested under as refs.
-    this.activeBids.refs.component.refreshActiveBids()
-    this.conversations.refs.component.refreshConversations(() => {
-      this.setState({ fetchingData: false })
-    })
+    if (this.activeBids && this.conversations) {
+      // Allow Conversations & Active Bids to properly force-fetch themselves.
+      // The stored refs are the Relay containers; the components themselves are nested under as refs.
+      this.activeBids.refs.component.refreshActiveBids()
+      this.conversations.refs.component.refreshConversations(() => {
```
damassi: Can you add an error handler here in a similar vein that we've started placing elsewhere? https://github.com/artsy/emission/blob/master/src/lib/Scenes/Home/Components/ForYou/Components/ArtworkCarousel.tsx#L243

damassi: Oh actually, thats already handled in the method! Ignore

PR #901, opened 2017-12-18
Improve network error reporting in Emission

orta:
Adds a `console.error` to a failing Relay query. 

This will show a red screen in dev mode when a Relay query totally fails ( e.g. not status 200 ) which happens when you make a bad request e.g.  `{"errors":[{"message":"Cannot query field \"__id\" on type \"FilterArtworks\"."}]}`

You can't rely on [this code](https://github.com/artsy/emission/blob/master/src/lib/metaphysics.ts#L20-L30) because by this point Relay has caught the error and has asked you to re-render.

#trivial #skip_new_tests

### DIFF
--- a/src/lib/utils/renderWithLoadProgress.tsx
+++ b/src/lib/utils/renderWithLoadProgress.tsx
@@ -8,6 +8,12 @@ export default (Component: React.ReactType, initialProps: object = {}): RenderCa
   let retrying = false
   return ({ error, props, retry }) => {
     if (error) {
+      const networkError = error as any
+      if (networkError.response && networkError.response._bodyInit) {
+        const data = networkError.response._bodyInit || {}
+        console.error(`Metaphysics Error: ${error.message}\n`, JSON.parse(data))
+      }
+
       if (retrying) {
         retrying = false
         // TODO: Even though this code path is reached, the retry button keeps spinning. iirc it _should_ disappear when
--- a/src/lib/utils/errors.tsx
+++ b/src/lib/utils/errors.tsx
@@ -1,3 +1,6 @@
+/**
+ * An extension of the Error with a Response object
+ */
 export class NetworkError extends Error {
-  response: any
+  response?: Response
 }


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

- `src/lib/utils/__tests__/errors-tests.tsx`
- `src/lib/utils/__tests__/renderWithLoadProgress-tests.tsx`

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/utils/renderWithLoadProgress.tsx
```diff
@@ -3,11 +3,18 @@ import React from "react"
...
+import { NetworkError } from "lib/utils/errors"
 
 export default (Component: React.ReactType, initialProps: object = {}): RenderCallback => {
   let retrying = false
   return ({ error, props, retry }) => {
     if (error) {
+      const networkError: NetworkError = error as any
+      if (networkError.response) {
+        const data = (networkError.response as any)._bodyInit
+        console.error(`Metaphysics Error: ${error.message}\n`, JSON.parse(data))
```
alloy: Do we really need to parse the data instead of just including it as-is? My paranoia is telling me that I wouldn’t want us to get into a situation where there are exceptions during error reporting.

orta: I guess it depends, it makes it much easier to debug at dev time (because it's an object you an introspect ) otherwise it's a bunch of escaped JSON  - I can add some null checking around it, which I think is reasonable

alloy: Fair enough 👍

PR #909, opened 2017-12-21
Bring back the artist rails

alloy:
Fixes https://github.com/artsy/collector-experience/issues/820.

Also re-enables removal of off-screen subviews on the home view.

#skip_new_tests

### DIFF
--- a/src/lib/Components/Home/ArtistRails/ArtistRail.tsx
+++ b/src/lib/Components/Home/ArtistRails/ArtistRail.tsx
@@ -177,30 +177,26 @@ export class ArtistRail extends Component<Props, State> {
   }
 
   refreshData = () => {
-    // TODO: Ensures rail has been mounted before a refresh occurs, circumventing setState errors.
-    // See https://stackoverflow.com/a/40969739/1038901 for more info.
-    if (this.refs.rail) {
-      if (this.inflightRequest) {
-        this.inflightRequest.dispose()
-      }
-
-      return new Promise((resolve, reject) => {
-        this.inflightRequest = this.props.relay.refetch({ ...this.props.rail, fetchContent: true }, null, error => {
-          if (error) {
-            console.error("ArtistRail.jsx", error.message)
-
-            this.setState({
-              loadFailed: true,
-            })
-
-            reject(error)
-          } else {
-            this.inflightRequest = null
-            resolve()
-          }
-        })
-      })
+    if (this.inflightRequest) {
+      this.inflightRequest.dispose()
     }
+
+    return new Promise((resolve, reject) => {
+      this.inflightRequest = this.props.relay.refetch({ ...this.props.rail, fetchContent: true }, null, error => {
+        if (error) {
+          console.error("ArtistRail.jsx", error.message)
+
+          this.setState({
+            loadFailed: true,
+          })
+
+          reject(error)
+        } else {
+          this.inflightRequest = null
+          resolve()
+        }
+      })
+    })
   }
 
   render() {
@@ -209,7 +205,7 @@ export class ArtistRail extends Component<Props, State> {
     }
 
     return (
-      <View ref="rail">
+      <View>
         <View style={styles.title}>
           <SectionTitle>{this.title()}</SectionTitle>
           <SectionTitle style={styles.subtitle}>{this.subtitle()}</SectionTitle>
--- a/src/lib/Components/Home/ArtistRails/ArtistRail.tsx
+++ b/src/lib/Components/Home/ArtistRails/ArtistRail.tsx
@@ -21,7 +21,7 @@ const Animation = {
 }
 
 interface Props extends ViewProperties, RelayProps {
-  registerRailModule?: (rail: ArtistRail) => void
+  registerRailModule?: (rail: ArtistRail | null) => void
   relay?: RelayRefetchProp
 }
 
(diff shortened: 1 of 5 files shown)


src/lib/Components/Home/ArtistRails/ArtistRail.tsx
```diff
@@ -177,30 +177,26 @@ export class ArtistRail extends Component<Props, State> {
   }
 
   refreshData = () => {
-    // TODO: Ensures rail has been mounted before a refresh occurs, circumventing setState errors.
-    // See https://stackoverflow.com/a/40969739/1038901 for more info.
-    if (this.refs.rail) {
```
alloy: This guard would never pass, because `refreshData` is called from `componentDidMount` and that’s apparently called before the `render` that would add the `ref`.

alloy: @damassi I’d like to know steps to reproduce the warning you were trying to fix, I haven’t seen it occur yet.

alloy: (But it was definitely only a warning, so this won’t lead to any runtime crashes.)

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

- `src/lib/Components/Home/ArtistRails/__tests__/ArtistRail-tests.tsx`
- `src/lib/Scenes/Home/Components/ForYou/__tests__/index-tests.tsx`

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
damassi: @alloy - if you comment out the ref check the issue should happen -- And it _appears_ as though its a warning and things dont break but its actually a more serious error in that from this point on no more rail refreshing occurs. (Meaning, if 10 out of 14 rails have been refreshed and the error occurs on the 11th iteration, 12, 13, and 14 will contain stale data.)

> This guard would never pass, because refreshData is called from componentDidMount and that’s apparently called before the render that would add the ref.

This is incorrect -- `componentDidMount` is the last thing that executes, post render. And since a ref is effectively a "unit" of DOM it _will_ execute since the dom is there and mounted. See https://reactjs.org/docs/react-component.html#componentdidmount. 

I took care to make sure that everything was working like it should so I'm wondering what the issue was exactly or if perhaps there was some kind of regression or change? Because it is super cool if we can remove the `ref` hack, but note that the warning _does_ stop rail fetch execution.

alloy: I read the docs the same way, but think there may be a slight difference in React Native, as there is no DOM in the same way those docs refer to. That’s not to say that RN maybe should work closer to that, but it was not doing so in my testing.

damassi: Hmm 🤔 I just dropped in some logs to test and unless theres something odd about this `componentDidMount` is definitely firing last: 

screen shot 2017-12-21 at 11 54 30 am

orta: Weird that this even worked, `ref` in react native is a function: https://github.com/artsy/emission/blob/13535321a2fd30efda15fadf9a137ba2fbf4bf78/src/lib/Components/Consignments/Screens/Metadata.tsx#L151

damassi: @orta - They can be strings or functions or references. 

For bug, see https://github.com/artsy/emission/pull/910

src/lib/Scenes/Home/Components/ForYou/index.tsx
```diff
@@ -86,8 +82,11 @@ export class ForYou extends React.Component<ViewProperties & RelayProps, State>
     }
   }
 
-  registerRailModule = (railModule: RailModule) => {
-    this.railModules.push(railModule)
+  /**
+   * Can be used to register a mounting module or de-register an unmounting module by setting the index to `null`.
+   */
+  registerRailModule(index: number, railModule: RailModule) {
+    this.railModules[index] = railModule
```
alloy: Allow the rail to be removed from the list of mounted rails by passing `null` as `railModule`.

src/lib/Scenes/Home/Components/ForYou/index.tsx
```diff
@@ -66,11 +66,7 @@ export class ForYou extends React.Component<ViewProperties & RelayProps, State>
     this.setState({ isRefreshing: true })
 
     try {
-      await Promise.all(
-        this.railModules.map(railModule => {
-          railModule.refreshData()
-        })
-      )
+      await Promise.all(this.railModules.map(railModule => railModule && railModule.refreshData()))
```
alloy: Only refresh mounted rails.

src/lib/Components/Home/ArtistRails/ArtistRail.tsx
alloy: I didn’t look further into the reason for the artist rail not having a `ref` on mounting, because I solved the actual problem in a different way.

src/lib/Scenes/Home/Components/ForYou/index.tsx
damassi: Ah, right on, nice and clear. Should we explicitly `Promise.reject()` here or is a falsy return ok for `Promise.all`? (I'm not totally certain)

alloy: `Promise.all` just returns an array with return values from the promises (which is `undefined` for mounted components and `null` for unmounted components`), it doesn’t care about those values itself, so we’re good :+1:

PR #917, opened 2018-01-03
[consignments] fixes cut-off text on small screens #trivial

sarahscott:
Fixes https://github.com/artsy/collector-experience/issues/832

The view looks unchanged on larger iPhone/iPad screens.

Also seems to solve some issues with rotation on iPad.

consignments_5


src/lib/Components/Consignments/Screens/Welcome.tsx
```diff
@@ -84,7 +82,7 @@ export default class Welcome extends React.Component<Props, null> {
   render() {
     return (
       <ConsignmentBG>
-        <ScrollView style={{ flex: 1 }} centerContent>
```
sarahscott: This `centerContent` prop in particular can cause problems.

mennenia:
Awesome! 👍

PR #918, opened 2018-01-03
[msg] fixes text clipping and rotation bug in zero state inbox ui #trivial

sarahscott:
Fixes https://github.com/artsy/collector-experience/issues/866 & https://github.com/artsy/collector-experience/issues/848

simulator screen shot jan 3 2018 2 19 55 pm

simulator screen shot jan 3 2018 2 18 03 pm

#skip_new_tests

### DIFF
--- a/src/lib/Components/Inbox/Conversations/ZeroStateInbox.tsx
+++ b/src/lib/Components/Inbox/Conversations/ZeroStateInbox.tsx
@@ -17,8 +17,7 @@ const CenteredView = styled.View`
 const VerticalLayout = styled.View`
   flex: 1;
   flex-direction: column;
-  min-height: 400;
-  max-height: 800;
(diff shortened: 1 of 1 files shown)


src/lib/Components/Inbox/Conversations/ZeroStateInbox.tsx
```diff
@@ -17,8 +17,7 @@ const CenteredView = styled.View`
 const VerticalLayout = styled.View`
   flex: 1;
   flex-direction: column;
-  min-height: 400;
-  max-height: 800;
+  ${isPad && "max-height: 600;"};
```
alloy: Why is this max-height still needed on iPad?

alloy:
Would love to understand the need for the max height on iPad, but looks innocuous enough 👍

PR #924, opened 2018-01-05
[Home/Fair] Don’t clip end of icons list.

alloy:
Fixes https://github.com/artsy/collector-experience/issues/869


alloy:
This is pretty much the same code as from the pervious hero units rail, so gonna self-merge.

PR #940, opened 2018-01-10
[Consignments] Sets the default option when you first select a category

orta:
Fixes https://github.com/artsy/collector-experience/issues/894

When you start selecting the categories, painting is now set in the state without you doing anything as the onselect trigger on happens when you change selection.


sweir27:
👍 awesome thanks!

PR #941, opened 2018-01-10
[Consignments] Bring required fields in line with force

mennenia:
Fix https://github.com/artsy/collector-experience/issues/898

Thanks for the pointer @orta!

Not enough time to make a gif, sorry 😞


orta:
Cool, I'll ship a commit or two adding tests

PR #947, opened 2018-01-12
[Home/Auctions] fixes loading indicator/section header not hiding 

l2succes:
Closes: https://github.com/artsy/collector-experience/issues/905

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ### Master
 
+* Fix loading indicator not hiding in Auctions tab of Home view - luc
+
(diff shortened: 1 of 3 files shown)


damassi:
@l2succes - i'm going to revert this because I believe it was a false positive related to the API change that was made. In a follow up PR i'll make sure to include the `loading: true` setState update. Sorry for the confusion; here's a link to a fuller explanation: https://artsy.slack.com/archives/C02BAQ5K7/p1515801541000196

alloy:
@damassi and keep logic to only render the section if there is any data at all

PR #948, opened 2018-01-12
[Analytics] Get rid of double fire upon entering background state

mennenia:
Fix https://github.com/artsy/collector-experience/issues/909

Issue was that when setting state, I was also calling analytics (and incorrectly setting the state of the tab). Actually, the `if` statement above takes care of it. If it goes to the Artists WFY page, then from that state change the analytics will fire anyway. No need to keep it in the `setState` below.


alloy:
Gotcha 👍

PR #951, opened 2018-01-13
[Home/Auctions] Disable Lots by Artists You Follow for this release.

alloy:
As mentioned on https://github.com/artsy/collector-experience/issues/905#issuecomment-357438308, this needs a bit more work so we’re pulling the feature now to be able to ship.


orta:
Cool cool

PR #955, opened 2018-01-22
Emulate how eigen handles status bars

orta:
Now the status bar is provided at the root level of the VC hierarchy, so individual screens shouldn't have to care about it at all.

screen shot 2018-01-22 at 4 42 00 pm

Previous to this PR, there wouldn't be that white bar.

PR #956, opened 2018-01-22
[iPhoneX] Fix the top margin for home

orta:
In #955 I removed the JS side of the equation, but I need to fix up the native side too.


alloy:
👍

PR #958, opened 2018-01-23
[MSG] Fix inbox header title when there are no messages

mennenia:
Fixes [this issue](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=2&projectKey=IN&selectedIssue=IN-7)

The zerostate inbox test is actually outdated also. I'll sort that when I come back. 

#skip_new_tests

### DIFF
--- a/src/lib/Components/Inbox/Conversations/__tests__/Conversations-tests.tsx
+++ b/src/lib/Components/Inbox/Conversations/__tests__/Conversations-tests.tsx
@@ -3,17 +3,27 @@ import React from "react"
 import "react-native"
 import * as renderer from "react-test-renderer"
 
+import { getTextTree } from "lib/utils/getTestWrapper"
+
 import { Conversations } from "../"
 
-it("looks correct when rendered", () => {
-  const tree = renderer.create(<Conversations me={meProps} relay={{ hasMore: jest.fn() } as any} />).toJSON()
-  expect(tree).toMatchSnapshot()
-})
 
-it("doesn't render the header when there are no messages", () => {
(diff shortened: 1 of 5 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/Inbox/Conversations/__tests__/Conversations-tests.tsx
```diff
@@ -10,6 +10,11 @@ it("looks correct when rendered", () => {
   expect(tree).toMatchSnapshot()
 })
 
+it("doesn't render the header when there are no messages", () => {
+  const tree = renderer.create(<Conversations me={mePropsEmpty} relay={{ hasMore: jest.fn() } as any} />).toJSON()
+  expect(tree).toMatchSnapshot()
+})
```
orta: TBH, this test doesn't _actually_ test that, it just verifies the JSON - would recommend adding a check for the lack of that specifically - you can see some examples here: https://github.com/artsy/emission/blob/master/src/lib/Components/Consignments/Screens/__tests__/Confirmation-tests.tsx#L87-L108

mennenia: ah ok, will push after dinner 👌

mennenia:
@orta that what you had in mind?

orta:
Yep 👍

Issue #848, opened 2017-11-27
Move to CocoaPods Keys for the auth token


orta:
As trusted artsy services  can only talk to other trusted services, we don't really want our OSS tokens being classed as trusted ( they are right now ) - what we should do is move the tokens to be the same as Eigen's via CP Keys

PR #960, opened 2018-01-25
Move to CocoaPods keys for Emission work

orta:
At the end of last year we started adding the ability to declare an OAuth client as having the role `"artsy"` which meant that we trusted the client id/secret as coming internally. 

This initially hit us during the Eigen 4.0 crunch, in which I made this issue https://github.com/artsy/emission/issues/848 to note we had to move. In that time, the OSS keys we use in this repo were classed as internal artsy apps. We've switched it back now, and the OSS keys have moved back to untrusted - this means we need to make Emission use "real" keys.

Thus: the move back to CocoaPods Keys so that we can have unique dev keys vs the OSS ones. It's the same flow in Eigen/Energy/Emergence/Eidolon.


alloy:
So for OSS builds nothing really changes? They can still fully utilise the example app? I guess I don’t really understand what changes in practice otherwise.

orta:
Nothing changes for OSS builds, but we can auth with Eigen's gravity creds so you can receive messages or submit consignments to prod/staging

alloy:
Ok, so some access for OSS builds does change. Not that I really think it matters much in practice, just trying to see the full picture.

PR #963, opened 2018-01-26
Reduces the complexity of submitting a consignment

orta:
Re: [IN-16](https://artsyproduct.atlassian.net/browse/IN-16)

There was a few extra layers of indirection here, which could end up with `this.state` maybe not being what you expect. 

Old behavior:

* Set local state to include `{ state: submitted }`
* Use callback of that to set state again (I assume for tests)
* Use callback of that to use the changed state to submit based on the current state, by reusing the shared update mechanism

New behavior:

* Use the current state, and inject  `{ state: submitted }`
* Use the function to send directly


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

PR #964, opened 2018-01-26
Impersonation and build failure

alloy:
The commits tell the story.


orta:
You could quickly edit the dangerfile to look for `static NSString *UserID = nil;` in the AppDelegate? Failing if it's not there

alloy:
Good idea 👍

PR #972, opened 2018-01-30
Synch with latest MP schema

ashkan18:
# Problem
Need to update local schema to match latest changes with https://github.com/artsy/metaphysics/pull/918

# Depends on https://github.com/artsy/metaphysics/pull/918


orta:
That's merged now 👍

orta:
This should fail danger (PR from a fork) which reminds me to try move some of the dangerfile here to peril. So I'l take it if it's red.

PR #979, opened 2018-02-02
[WIP][Dev] Adds detox

orta:
https://github.com/wix/detox - https://hackernoon.com/detox-gray-box-end-to-end-testing-framework-for-mobile-apps-196ccd9564ce

You should ideally have a look at both of ^ before digging deeper.

This PR adds support for running our RN components end to end as gray box tests with Emission. 

I think this could probably be useful if we have a known user that we can verify that loading a Gene page would open a gene, with a known description etc etc

- [ ] Authenticate
- [ ] Open a root view
- [ ] Validate data that must have come through a full relay query


orta:
Closing for now, as I've not looked at this in a year

PR #980, opened 2018-02-02
[WIP][Dev] Adds loki

orta:
Another angle of attack vs #979 

loki2018-02-02 17_17_20

^ maybe 4x or 8x faster than reality

Adds Loki, which does visual regression tests for React Native components. It does this by connecting to a running copy of storybooks ( the same way my vscode extension does) and then looping through all the stories and generates an image for them.

This is slow, like too slow for me to watch slow, I wonder if there are ways to improve, but there's _something_ of value here.


orta:
Closing for now, as I've not looked at this in a year

PR #981, opened 2018-02-05
Update README.md

damassi:
I'm setting up a new computer and the current automated copy / paste command fails in zsh. Figured `&&` would be safer in more environments.


orta:
Yeah, fine by me

PR #982, opened 2018-02-06
[Relay] Uses latest version and our patches

alloy:
* Uses [our custom Node ID patch](https://github.com/facebook/relay/pull/2249)
* Uses [our language plugin patch](https://github.com/facebook/relay/pull/2293)
* Uses [the TS language plugin](https://github.com/kastermester/relay-compiler-language-typescript) I’ve been working on with @kastermester 🙌 

I decided to start checking in the artefacts, because now that we have and use TS typings it will be useful to be able to see those when reviewing PRs.

⚠️ If you already had Emission checked out before, be sure to delete all previous artefacts and reset the transformer cache:

```
$ find . -name __generated__ | xargs rm -rf
$ yarn relay
$ yarn start-packager --reset-cache
```

Afterwards you can use the regular script again:

```
$ yarn start
```

#skip_new_tests


alloy:
Seems to fail on CI due to https://github.com/ds300/patch-package/issues/36. Will look into that tomorrow; for those interested, you _should_ be able to pull and run this locally.

src/lib/Scenes/Home/Components/ForYou/Components/ArtworkCarouselHeader.tsx
```diff
@@ -174,31 +181,31 @@ export default createFragmentContainer(
           }
         }
       }
+      # FIXME: There is a bug in the Relay transformer used before generating Flow types, and thus also our TS type
+      #        generator, that leads to a union selection _with_ a __typename selection being normalized incorrectly.
+      #        What ends up happening is that _only_ the common selection is being omitted from the second fragment,
+      #        i.e. in this case the artist selection is not present in the second fragment.
+      #
+      #        This can be seen much more clear when adding __typename to the context part in ArtworkRail.tsx.
+      #
```
alloy: Working around a bug here ^

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

- `src/lib/Components/Artist/Articles/__tests__/index-tests.tsx`
- `src/lib/Components/Artist/Artworks/__tests__/index-tests.tsx`
- `src/lib/Components/Artist/Shows/__tests__/index-tests.tsx`
- `src/lib/Components/ArtworkGrids/__tests__/GenericGrid-tests.tsx`
- `src/lib/Components/ArtworkGrids/RelayConnections/__tests__/ArtistForSaleArtworksGrid-tests.tsx`
- `src/lib/Components/ArtworkGrids/RelayConnections/__tests__/ArtistNotForSaleArtworksGrid-tests.tsx`
- `src/lib/Components/Home/ArtistRails/__tests__/ArtistRail-tests.tsx`
- `src/lib/Components/Home/ArtworkRails/__tests__/ArtworkRailHeader-tests.tsx`
- `src/lib/Components/Home/__tests__/HeroUnits-tests.tsx`
- `src/lib/Components/Inbox/ActiveBids/__tests__/index-tests.tsx`
- `src/lib/Components/Inbox/Conversations/Preview/Attachment/__tests__/AttachmentPreview-tests.tsx`
- `src/lib/Components/RelatedArtists/__tests__/RelatedArtist-tests.tsx`
- `src/lib/Components/RelatedArtists/__tests__/index-tests.tsx`
- `src/lib/Scenes/Favorites/Components/Artists/__tests__/index-tests.tsx`
- `src/lib/Scenes/Favorites/Components/Artworks/Relay/__tests__/ArtworksRenderer-tests.tsx`
- `src/lib/Scenes/Favorites/Components/Artworks/__tests__/index-tests.tsx`
- `src/lib/Scenes/Favorites/Components/Categories/__tests__/index-tests.tsx`
- `src/lib/Scenes/Home/Components/ForYou/Components/__tests__/ArtworkCarousel-tests.tsx`
- `src/lib/Scenes/Home/Components/ForYou/Components/__tests__/ArtworkCarouselHeader-tests.tsx`
- `src/lib/Scenes/Home/Components/ForYou/__tests__/index-tests.tsx`
- `src/lib/Scenes/Home/Components/Sales/Relay/__tests__/SalesRenderer-tests.tsx`

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

kastermester:
I hope this works out well for you guys - and I look forward to continue working with @alloy to make sure TS and Relay feels like tools that are made to work together! :)

alloy:
Alright, this is all green now 👍

.gitattributes
```diff
@@ -1 +1,2 @@
 CHANGELOG.md merge=union
+src/__generated__/*.graphql.ts linguist-generated
```
orta: Good stuff, thanks @pchaigno 🥇

.prettierignore
```diff
@@ -0,0 +1,2 @@
+src/__generated__/*.ts
+node_modules
```
orta: also, a great move 👍

src/__generated__/About_artist.graphql.ts
```diff
@@ -0,0 +1,106 @@
+/* tslint:disable */
```
orta: Ah, shame, guess the gitattributes has to be merged before we get to use it in PRs

src/__generated__/About_artist.graphql.ts
```diff
@@ -0,0 +1,106 @@
...
+
+import { ConcreteFragment } from "relay-runtime";
+export type About_artist = {
+    readonly has_metadata: boolean | null;
+    readonly is_display_auction_link: boolean | null;
```
orta: In [our schema](https://github.com/artsy/emission/blob/5126e093809c015f6ff2d5de867807505703a2cf/data/schema.graphql#L356-L357) this looks like:

```graphql
type ArtistItem implements Node {
  [...]
  # Only specific Artists should show a link to auction results.
  is_display_auction_link: Boolean
}
```

It would be nice to bring these comments into the interface in v1.1 of the TS generator

src/lib/Containers/Artist.tsx
```diff
@@ -93,11 +96,11 @@ export class Artist extends Component<Props & ViewProperties, State> {
...
     switch (this.selectedTabTitle()) {
       case TABS.ABOUT:
-        return <About artist={this.props.artist} />
+        return <About artist={this.props.artist as any} />
       case TABS.WORKS:
-        return <Artworks artist={this.props.artist} />
+        return <Artworks artist={this.props.artist as any} />
       case TABS.SHOWS:
-        return <Shows artist={this.props.artist} />
+        return <Shows artist={this.props.artist as any} />
```
orta: I assume these additional `as any` are due to the relay data masking, I think @kastermester covers [that here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23460#pullrequestreview-94403153)

src/lib/Containers/Inbox.tsx
```diff
@@ -49,9 +46,8 @@ export class Inbox extends React.Component<Props, State> {
 
     if (this.activeBids && this.conversations) {
       // Allow Conversations & Active Bids to properly force-fetch themselves.
-      // The stored refs are the Relay containers; the components themselves are nested under as refs.
-      this.activeBids.refs.component.refreshActiveBids()
-      this.conversations.refs.component.refreshConversations(() => {
+      this.activeBids.refreshActiveBids()
+      this.conversations.refreshConversations(() => {
```
orta: nice, simpler

src/lib/Containers/__tests__/Conversation-tests.tsx
```diff
@@ -78,14 +78,17 @@ it("sends message when composer is submitted", async () => {
...
     from: {
       name: "Anita Garibaldi",
       email: "anita@garibaldi.br",
       initials: "AG",
     },
     to: { name: "Kimberly Klark", initials: "KK" },
+    unread: false,
     initial_message: "Adoro! Por favor envie-me mais informações",
+    last_message_id: "222",
     messages: {
```
orta: did the relay runtime warn you about these missing fields?

src/lib/Containers/Inbox.tsx
alloy: And using public API.

src/lib/Containers/Artist.tsx
alloy: Aye 👍

alloy: Btw, the only reason that in some cases we didn’t need to specify `as any` previously was that those components were overriding the constructor without re-typing the props, which meant that the component basically accepted anything by default.

src/__generated__/About_artist.graphql.ts
alloy: What do you mean? Is it not working for you?

src/__generated__/About_artist.graphql.ts
alloy: Great idea!

kastermester: Great idea, I have some experience with this in some internal tooling we use where I work. Can you create an issue in https://github.com/relay-tools/relay-compiler-language-typescript ?

alloy: https://github.com/relay-tools/relay-compiler-language-typescript/issues/35

src/lib/Containers/Artist.tsx
kastermester: The future plans tie in with the way the artifacts are generated - it means that you cannot pass data to a component, without having actually specified the fragment as part of your fragment text (ie. you need to spread the child component's fragment into your own at the appropriate place to be able to render the child component with the given data).

As mentioned, we're waiting for conditional types coming in TS 2.8 (and conditional infered types as well to be able to type up the `componentRef` prop).

.prettierignore
alloy: Turns out that the graphql-js patch wasn’t applying because the whole file had changed due to prettier on save :)

src/lib/Containers/__tests__/Conversation-tests.tsx
alloy: Nope, the expanded type definitions did that.

src/__generated__/About_artist.graphql.ts
orta: They're flattened, yep, but I can't tell if that's because the PR is too long because all of the in app source code changes are also flattened after it.

alloy: Ah I see. I’m pretty sure that some of these artefacts were expanded by default yesterday.

alloy:
Alright, going for the merge 🚀

src/__generated__/About_artist.graphql.ts
orta: Ah great, I threw a few minutes at understanding the code to add it- happy to leave that to you 🍡

src/lib/Containers/Inbox.tsx
kastermester: Just nitpicking for you guys here: Stuff coming from `ref` callbacks in `React` can be set to null (by the ref callback) ie. when the component is unmounting. I'd advice setting the type to `T | null` and then initialize it to `null`. Also newer TS versions have definite assignment checks that will complain about this sort of code.

You can then define a method such as this:

```ts
function assertNonNull<T>(obj: T | null): T {
  if (obj == null) {
    throw new Error('Object is null');
  }
  return obj;
}
```

And use that when you assume the object is non null. You may be aware of this and simply having chosen to save the hassle, but I thought I'd let you know :)

src/__generated__/About_artist.graphql.ts
alloy: Unless @kastermester has time immediately, I think that somebody else –such as yourself– contributing it would be great. At least you’ll know who to ask questions and perhaps @kastermester knows of some things to keep in mind from the top of his head.

orta: Ergh, that alloy, always pushing for more contributors 📟 

Will continue my digging

kastermester: The biggest concern here comes from interface types and how they interact with the type generation (without remembering exactly how that works in the TS plugin here mind you). If the type ends up being something like:

```ts
type MyType = {
  __typename: 'MyType1';
  name: string;
} | {
  __typename: 'MyType2';
  name: string;
} | {
  // Some comment here telling that this is to cover future types being added
  __typename: ' %other';
}
```

The question now becomes - which comment should be added to the `name` property - assuming there's one for the interface, and one for each of the types. Previously I have added the comment from the concrete types, such that the interface comment never shows up, but simply the concrete type comments. This has the effect on editors (at least VS Code but AFAIk all editors get their info from the TS language server) that you get the combined comment when simply accessing `name`.

I would love to find a better way to handle this sort of thing - but disregarding that, it should be pretty straight forward. If you have the time to make a pull request that'd be great.

src/lib/Containers/Inbox.tsx
kastermester: I'm not sure you can type a function to never return if some condition is not met, sadly. I believe it is baked into flow that the invariant function behaves like that.

> Regarding the ref callback also calling for unmounting, that’s just a thing that never happens for these two components.

Yeah I know, I'm just personally a bit obsessed with semantics and there's something in me that itches when I see stuff typed in an incorrect way, as I know the next person editing the code is going to rely on the types and make assumptions based on that. I understand how doing the code properly can be a bit of a hinderance and annoyance though.

alloy: Ah yes, good point. Looks like it is being discussed https://github.com/Microsoft/TypeScript/issues/10421 & https://github.com/Microsoft/TypeScript/issues/8655.

I hear you on worrying about the next person, but I try to strike a balance for my coworkers between untyped and typed code that doesn’t become too tiresome on them :)

.gitattributes
damassi: @orta - can you explain whats going on here?

src/lib/Containers/Inbox.tsx
damassi: The beauty of TypeScript 😄

.gitattributes
orta: I can - here's the historical context: https://twitter.com/pchaigno/status/961134607113637889

damassi: tyty, makes sense 👍

src/__generated__/About_artist.graphql.ts
alloy: @orta FYI https://github.com/graphql/graphql-js/pull/1165

PR #954, opened 2018-01-19
[Dev] Update to React Native 0.54

orta:
[@alloy here] This is working now.

0.51 [gives us SafeAreaView](https://github.com/facebook/react-native/tree/0.51-stable/Libraries/Components/SafeAreaView) - which will be needed for iPhone X support

Plus we can do function only components and get HMR 👍

PR #983, opened 2018-02-14
[msg] adds a scrollToLastMessage function

sarahscott:
- [x] scrollToLastMessage function in Messages
- [x] resolve TS errors

scroll_to_msg

fixes https://artsyproduct.atlassian.net/browse/IN-10


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Containers/Conversation.tsx
```diff
@@ -56,6 +56,7 @@ const track: Track<Props, State> = _track
 
 @track()
 export class Conversation extends React.Component<Props, State> {
+  messages?: any
```
sarahscott: Because `Messages` is a container, TypeScript doesn't recognize it as a valid type, so it has to be an `any`. We don't have refs like this to relay containers elsewhere in emission; is there a workaround that I'm missing?

src/lib/Components/Inbox/Conversations/Messages.tsx
```diff
@@ -129,13 +136,14 @@ export class Messages extends React.Component<Props, State> {
           width: 708,
           alignSelf: "center",
         }
-      : {}
+      : ({} as any)
```
alloy: I wonder if it would work without needing to cast to `any` if this were `null` (or `undefined`)?

src/lib/Containers/Conversation.tsx
```diff
@@ -156,6 +158,7 @@ export class Conversation extends React.Component<Props, State> {
           </Header>
           {!this.state.isConnected && <ConnectivityBanner />}
           <Messages
+            ref={messages => (this.messages = messages)}
```
alloy: Use `componentRef` instead so you don’t need to use `this.messages.refs.component` above

src/lib/Components/Inbox/Conversations/Messages.tsx
```diff
@@ -16,8 +16,9 @@ const isPad = Dimensions.get("window").width > 700
 
 interface Props {
   conversation: Messages_conversation
-  relay: RelayPaginationProp
+  relay?: RelayPaginationProp
```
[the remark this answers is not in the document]
sarahscott: Nope, it should be fine without that; was an experiment.

src/lib/Containers/Conversation.tsx
alloy: Ah yes, what you could do is also export the _actual_ component class and import that just for this purpose.

src/lib/Components/Inbox/Conversations/Messages.tsx
sarahscott: It throws this TS error for `null`, `undefined`, or `{}`:

```
Type '{ inverted: boolean; data: { __id: string; impulse_id: string; is_from_user: boolean; body: strin...' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<FlatListProperties<any>, ComponentState>...'.
  Type '{ inverted: boolean; data: { __id: string; impulse_id: string; is_from_user: boolean; body: strin...' is not assignable to type 'Readonly<FlatListProperties<any>>'.
    Types of property 'style' are incompatible.
      Type '{ width: number; alignSelf: string; }' is not assignable to type 'StyleProp<ScrollViewStyle>'.
        Type '{ width: number; alignSelf: string; }' is not assignable to type 'RecursiveArray<false | ScrollViewStyle | RegisteredStyle<ScrollViewStyle>>'.
          Property 'length' is missing in type '{ width: number; alignSelf: string; }'.
```

alloy: Interesting! One would think that you could omit `style` for a component. Might be a bug in the RN DT typings? 🤔

sarahscott: This is actually  a bug in my VSCode setup - I can run `type-check` with no problems 🤔. Will remove the `any`.

src/lib/Containers/Conversation.tsx
sarahscott: Would it make sense to rename it something like `MessagesComponent`? Won't it be a duplicate declaration otherwise?

alloy: Yeah indeed, you can do so with ES6 imports. E.g.

```ts
import { Messages as MessagesComponent } from "…"
```

src/lib/Components/Inbox/Conversations/Messages.tsx
```diff
@@ -31,6 +32,8 @@ const LoadingIndicator = styled.ActivityIndicator`
 `
 
 export class Messages extends React.Component<Props, State> {
+  flatList: FlatList<any> | any
```
sarahscott: This, however, is because our `@types/react-native` are slightly out-of-date:

types/react-native/index.d.ts:
```
/**
     * Scrolls to the item at a the specified index such that it is positioned in the viewable area
     * such that `viewPosition` 0 places it at the top, 1 at the bottom, and 0.5 centered in the middle.
     * May be janky without `getItemLayout` prop.
     */
    scrollToIndex: (params: {animated?: boolean, index: number, viewPosition?: number}) => void
```

[React Native docs](https://facebook.github.io/react-native/docs/flatlist.html#scrolltoindex)
```
Valid params keys are:

'animated' (boolean) - Whether the list should do an animation while scrolling. Defaults to true.
'index' (number) - The index to scroll to. Required.
'viewOffset' (number) - A fixed number of pixels to offset the final target position. Required.
'viewPosition' (number) - A value of 0 places the item specified by index at the top, 1 at the bottom, and 0.5 centered in the middle.
```

alloy: Is that because `viewOffset` is missing in the typings? If so, send a PR to update https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts

sarahscott: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23648

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

iRoachie:
@alloy good to merge over at DT

src/lib/Components/Inbox/Conversations/Messages.tsx
```diff
@@ -136,6 +142,7 @@ export class Messages extends React.Component<Props, State> {
         inverted={!this.state.shouldStickFirstMessageToTop}
         data={this.state.shouldStickFirstMessageToTop ? messages.reverse() : messages}
         renderItem={this.renderMessage.bind(this)}
+        ref={flatList => (this.flatList = flatList as any)}
```
sarahscott: See https://github.com/DefinitelyTyped/DefinitelyTyped/issues/16318#issuecomment-346970646

alloy:
@iRoachie Thanks!

sarahscott:
My type fix is for RN 0.52+ and it looks like [we're about to update RN anyway](https://github.com/artsy/emission/pull/954), so this can be merged once that's all set.

alloy:
@sarahscott The new RN types version is available.

sarahscott:
Right but aren't we using RN 0.48? I was thinking things could break using the 0.52+ types with RN 0.48.

sarahscott:
Should be fine to merge; we'll have to add a `viewOffset` to the `scrollToIndex` method when we upgrade RN.

alloy:
Makes sense 👍

PR #986, opened 2018-02-20
Adds a user-agent to emission queries

orta:
Adds a user agent to the metaphysics graphql requests - re https://github.com/artsy/force/issues/2191

We could expose some more information in the user agent (version of app, version of emission) if we write a bit of native code, opening so that the discussion on _what_ to send can happen,


alloy:
I agree with @joeyAghion.

Not sure what you were thinking in terms of native code, but I think it can be limited to an extra configuration that the `AREmission` object takes when initialising.

orta:
OK, this will force someone to add the user agent in Eigen when it updates 👍

alloy:
Perfect 👌

PR #987, opened 2018-02-22
[dev] Make sure we set a user agent #trivial

sarahscott:
This makes sure Emission's AppDelegate set its user agent.

#skip_new_tests


alloy:
Thanks 👍

PR #988, opened 2018-02-22
[Auctions] Fix for dynamic updating of numColumns in auctions FlatList #trivial

sarahscott:
We can't dynamically update `numColumns` in our `FlatList`, but that's ok because we only need to check if we're on an iPad once. 

Note for the curious or for future implementations that would require dynamically changing `numColumns`: it is technically possible to update the number of columns but one would also have to [update keys as well](https://stackoverflow.com/questions/44291781/dynamically-changing-number-of-columns-in-react-native-flat-list). 

Fixes https://artsyproduct.atlassian.net/browse/IOS-95


mennenia:
:+1: awesome :)

We might want to change the column count for landscape ipad pro someday (there's a bit of an empty space on the side unfortunately), so the bit about keys is excellent info 🎉

PR #989, opened 2018-02-22
[Auctions] Use is_closed field in artwork grid

mennenia:
https://artsyproduct.atlassian.net/browse/IOS-90

The `is_open` field is like an internal Artsy auction field, essentially saying that the work is open to bid on Artsy. This works for live auctions, as you can already put bids in before it's started.

So what happened in the issue is that timed auctions can be published before the auction opens, the works aren't "biddible" yet, yet the auction also hasn't closed either. 

So the fix here is using the `is_closed` field, which is only true when the auction has properly closed, and as such we can then and only then display "Auction Closed". 

What the data looks like in metaphysics for the type of work that caused the bug:

screen shot 2018-02-22 at 13 41 35


src/lib/Components/ArtworkGrids/__tests__/__snapshots__/Artwork-tests.tsx.snap
```diff
@@ -1,5 +1,176 @@
...
+            "color": "#666666",
+            "fontSize": 12,
+          },
+          Object {
+            "fontFamily": "AGaramondPro-Regular",
+          },
+        ]
+      }
+    >
+      $200 (1 bid)
```
mennenia: monies!

sweir27:
This logic looks okay to me! Thanks for the fix 😄

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

mennenia:
@sweir27 Thank you for having a look and getting a staging auction up! 🙏

mennenia:
@orta could you have a look over as well and merge if you think it's ok? :)

orta:
Yep - this makes sense

PR #993, opened 2018-03-13
remove showing `fromEmail` on conversation snippet

ashkan18:
# Problem
We currently show `fromEmail` in the signature section when showing messages of a conversation. We don't want to expose emails.
https://artsyproduct.atlassian.net/browse/PLATFORM-95
# Solution

Removed `fromEmail` from conversation snippet and just showing `fromName`.


mzikherman:
I think this is only ❌ since its open from a fork (instead of a branch).

src/lib/Components/Inbox/Conversations/Message.tsx
```diff
@@ -186,7 +183,7 @@ export class Message extends React.Component<Props, any> {
 
             {this.renderBody()}
 
-            {!message.is_from_user && <FromSignature>{fromSignature}</FromSignature>}
+            {!message.is_from_user && <FromSignature>{fromName}</FromSignature>}
```
mzikherman: Nice! I guess this can be null, in which case I this will render fine (div with height of 0), but maybe worth a check?

Overall 👍

ashkan18: good point! ended up adding a new story for message and testing it there, noticed that not having from name was still adding space between messages, so ended up adding extra check for `fromName &&`. Now it looks like this:

screen shot 2018-03-14 at 12 39 27 pm

and for a message with fromName:
screen shot 2018-03-14 at 12 39 34 pm

src/lib/Components/Inbox/Conversations/Message.tsx
```diff
@@ -186,7 +182,7 @@ export class Message extends React.Component<Props, any> {
 
             {this.renderBody()}
 
-            {!message.is_from_user && <FromSignature>{fromSignature}</FromSignature>}
+            {!message.is_from_user && fromName && <FromSignature>{fromName}</FromSignature>}
```
mzikherman: nice

PR #995, opened 2018-03-14
[Relay] Add query response caching

alloy:
This caches responses for 2 minutes, which is the average session time of users in recent app versions. This makes sure that as somebody is actively browsing and go back and forth between e.g. an artist and home the artist is only fetched once.

While that is a nice optimisation on its own, this is also the foundation for where Emission will look for preloaded data that we will fetch natively in a next PR.


mennenia:
Looks great!

PR #996, opened 2018-03-16
Adds a user agent to the inquiry POST

orta:
@anipetrov asked me about an unknown user agent in analytics, turns out in 4.0 we ended up with one being generated at runtime instead of being consistent with our old Eigen one.


alloy:
Oh yeah, good one 👍

PR #997, opened 2018-03-18
[Dev] Updates the makefile for shipping new RN deps

orta:
Adds the new deps - https://github.com/artsy/Specs/commit/9d5378347300bade19a4b30ec39ec91832ab3eee

#trivial

### DIFF
--- a/Emission.podspec
+++ b/Emission.podspec
@@ -30,7 +30,7 @@ podspec = Pod::Spec.new do |s|
   s.dependency 'Artsy+UIFonts', '>= 3.0.0'
   s.dependency 'Extraction', '>= 1.2.1'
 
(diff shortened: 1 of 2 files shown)


orta:
if there's any more this should probably get turned into a ruby/js script, it's a bit unmanageable

orta:
screen shot 2018-03-18 at 2 37 11 pm

They aren't needed

orta:
Spotted this too:

https://github.com/facebook/react-native/blob/1c534719115fc4e99da9a414cff8638a6059a948/ReactCommon/yoga/yoga.podspec#L42

alloy:
We should just take some time to move the specs in the RN repo into an actual Spec repo and never deal with this stuff again. I’ll do so during this sprint, because this is just a time hog.

PR #1001, opened 2018-04-09
use ruby api to retrieve spec details

koenpunt:
this is much faster than reaching out to the command line util, which has to spawn a new ruby process for each podspec.

Is has to be verified that this actually works as-is, because in my implementation I had to prefix the podspec files with the root variable (here from `EMISSION_ROOT`), because otherwise it couldn't find the them. But since the ipc method hadn't that either, I kept it without the root prefix for now.


orta:
Agree 👍

orta:
We'll be deploying pretty soon, so this will get a real battle test one of these days

orta:
I'm ok with the red

PR #1002, opened 2018-04-11
Adds <SelectMaxBid> component

yuki24:
addresses https://artsyproduct.atlassian.net/browse/PURCHASE-42 and is possibly related to https://artsyproduct.atlassian.net/browse/PURCHASE-34

The goal of this PR is to set up a foundation where all the components for the new bidding flow lives, by actually building a component. I picked up the pattern used for the `Consignments` directory to get started, but wanted to make sure that this is the pattern we all feel comfortable with before I move forward too deeply.

In this PR there's only `<SelectMaxBid>`, but I'd imagine that the `Screens/` will grow as we implement more pages and add a `Components/` directory to organize smaller components. One interesting thing we are going to do is that a subclass of `UIViewController` may have to live in the Emission as well. so we'll have to think about how we are going about this.

This PR is basically just a conversation starter. I'm not really tied to any specific pattern at this point. We don't even have to follow the existing pattern if there's something that feels more organized. Any feedback is welcome!

### Directory structure:

```
src/lib/Components/Bidding/
├── Screens
│   └── SelectMaxBid.tsx
├── __stories__
│   └── Bidding.story.tsx
└── __tests__
```

### The `<SelectMaxBid>` component:

screen shot 2018-04-11 at 2 29 03 pm

### DIFF
--- a/src/lib/Components/Bidding/Components/Button.tsx
+++ b/src/lib/Components/Bidding/Components/Button.tsx
@@ -1,0 +1,22 @@
+import React from "react"
+import { StyleSheet } from "react-native"
+
+import { ButtonProps, InvertedButton } from "../../Buttons"
+
+export class Button extends React.Component<ButtonProps> {
+  render() {
+    const { style, ...props } = this.props
+
+    return <InvertedButton style={[styles.default, style]} {...props} />
+  }
+}
+
+const styles = StyleSheet.create({
+  default: {
+    // fontSize: 14,
+    // fontWeight: 'normal',
+    // fontStyle: 'normal',
+    // letterSpacing: 1,
+    // color: '#ffffff'
+  },
+})
--- a/src/lib/Components/Bidding/Components/Title.tsx
+++ b/src/lib/Components/Bidding/Components/Title.tsx
@@ -1,0 +1,19 @@
+import React from "react"
+import { TextProperties } from "react-native"
+import styled from "styled-components/native"
+
+import { Fonts } from "../../../data/fonts"
+
+export class Title extends React.Component<TextProperties> {
+  render() {
+    return <StyledText {...this.props}>{this.props.children}</StyledText>
+  }
+}
+
+const StyledText = styled.Text`
+  font-family: "${Fonts.GaramondRegular}";
+  margin: 10px;
+  font-weight: bold;
+  font-size: 20px;
+  text-align: center;
+`
--- a/src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
+++ b/src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
@@ -1,0 +1,55 @@
+import React from "react"
+import styled from "styled-components/native"
+
+import { Button } from "../Components/Button"
+import { MaxBidPicker } from "../Components/MaxBidPicker"
+import { Title } from "../Components/Title"
+
+interface SelectMaxBidProps {
+  saleArtworkID: string
+}
+
+// we are using hard-coded values for now.
+const Bids = [
+  {
+    label: "$35,000 USD",
+    value: 3500000,
+  },
+  {
+    label: "$40,000 USD",
+    value: 4000000,
+  },
+  {
+    label: "$45,000 USD",
+    value: 4500000,
+  },
+  {
+    label: "$50,000 USD",
+    value: 5000000,
+  },
+  {
+    label: "$55,000 USD",
+    value: 5500000,
+  },
+]
+
+export class SelectMaxBid extends React.Component<SelectMaxBidProps> {
+  render() {
+    return (
+      <Container>
+        <Title>Your max bid</Title>
+
+        <MaxBidPicker selectedValue={4500000} bids={Bids} />
+
+        <Button text="NEXT" />
+      </Container>
+    )
+  }
+}
+
+const Container = styled.View`
+  flex: 1;
+  flex-direction: column;
+  justify-content: space-between;
+  margin: 20px;
+`
--- a/src/lib/Components/Bidding/__stories__/BidFlow.story.tsx
+++ b/src/lib/Components/Bidding/__stories__/BidFlow.story.tsx
@@ -1,7 +1,13 @@
 import { storiesOf } from "@storybook/react-native"
 import React from "react"
-import BidFlow from "../Screens/BidFlow"
 
-storiesOf("Bidding").add("Show bid flow", () => {
-  return <BidFlow saleArtworkID="5aada729139b216c0bf18103" />
-})
+import BidFlow from "../Screens/BidFlow"
+import { SelectMaxBid } from "../Screens/SelectMaxBid"
+
+storiesOf("Bidding")
+  .add("Show bid flow", () => {
+    return <BidFlow saleArtworkID="5aada729139b216c0bf18103" />
+  })
+  .add("Select Max Bid", () => {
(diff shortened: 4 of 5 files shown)


src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
```diff
@@ -0,0 +1,38 @@
...
+import styled from "styled-components/native"
+
+import { InvertedButton } from "../../Buttons"
+import Serif from "../../Text/Serif"
+
+export class SelectMaxBid extends React.Component<SelectMaxBidProps> {
+  render() {
+    return (
+      <Container>
+        <Serif style={{ fontSize: 20, textAlign: "center", fontWeight: "bold", margin: 18 }}>Your max bid</Serif>
```
ashfurrow: What is the convention around static components like this that don't really depend on props and have no children? Should we use styled components here?

orta: You probably will want the same formatting in some other screen too, so that makes sense

yuki24: Yeah I had the same question. @orta @mennenia do you have any wisdom on this?

orta: I think the ideal state is that the render method doesn't contain any styling, I know I've not succeeded in that in a bunch of places

On the other hand, I think it's fine to make a re-usable title component where it needs a custom bottom margin to just have that margin inlined, so you can understand where that change has come in from

mennenia: I would indeed make a `Title` styled component :). 

In Emission now we do try to steer away from inline style, unless, like Orta just said, it's a small margin indent. Though I would then still use a styled component that is based off something and adds that one property.

But you will very likely still see inline style throughout our codebase, from before we used Styled Comps / while we were still figuring things out.

mennenia:
@yuki24 The `UIViewController` will live in the pod (= the library that we export to Eigen). 

If you look in `emission/Pod/Classes/ViewControllers` you will see that we already have some corresponding native view controllers, and how they work. For example, the [`ARArtistComponentViewController`](https://github.com/artsy/emission/blob/master/Pod/Classes/ViewControllers/ARArtistComponentViewController.m). Similarly, we would probably get a `ARNewBidComponentViewController` or something 😄. (edit = this is what @ashfurrow's [PR](https://github.com/artsy/emission/pull/1003/files) does).

The [`AppRegistry.tsx`](https://github.com/artsy/emission/blob/master/src/lib/AppRegistry.tsx) is where components then get registered with the corresponding containers. The [module name](https://github.com/artsy/emission/blob/master/Pod/Classes/ViewControllers/ARArtistComponentViewController.m#L22-L30) coming from the native view controller will be matched up to the right high level RN container / component.

The reason all native (objective C) view controllers start with `AR` by the way, is to make them unique. This is a bit of a legacy iOS thing, and so it's common practice to pick 2-3 letters that denotes your organisation or project. In our case, AR = Artsy.

src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
sepans: That header 👆 is the same on some other screens according to the [designs](https://app.zeplin.io/project/5ab94282c124db8d58365b47/screen/5acbdf943000a9041d8e9abe) so it could be reused.

src/lib/Components/Bidding/Components/Button.tsx
```diff
@@ -0,0 +1,22 @@
...
+  }
+}
+
+const styles = StyleSheet.create({
+  default: {
+    // fontSize: 14,
+    // fontWeight: 'normal',
+    // fontStyle: 'normal',
+    // letterSpacing: 1,
+    // color: '#ffffff'
```
yuki24: Right now it doesn't seem like the `< InvertedButton>` component is customizable (see the warning below). I'll send a separate PR to make it more flexible.

screen shot 2018-04-13 at 1 18 10 pm

src/lib/Components/Bidding/__stories__/BidFlow.story.tsx
```diff
@@ -1,7 +1,13 @@
...
 
-storiesOf("Bidding").add("Show bid flow", () => {
-  return <BidFlow saleArtworkID="5aada729139b216c0bf18103" />
-})
+storiesOf("Bidding")
+  .add("Show bid flow", () => {
+    return <BidFlow saleArtworkID="5aada729139b216c0bf18103" />
+  })
+  .add("Select Max Bid", () => {
+    return <SelectMaxBid saleArtworkID="5aada729139b216c0bf18103" />
```
yuki24: Passing in `saleArtworkID` for now but this is subject to change.

src/lib/Components/Bidding/Components/Title.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+    return (
+      <Serif style={[styles.default, style, styles.required]} {...props}>
+        {children}
+      </Serif>
+    )
+  }
+}
+
+const styles = StyleSheet.create({
```
mennenia: Sorry if this wasn't clear on the other comment!

We're trying to move to Styled Components rather than using StyleSheets. Here's the [Inquiry screen](https://github.com/artsy/emission/blob/master/src/lib/Containers/Inquiry.tsx) as an example. It's a bit of a big file, but you can see towards the top of the file there are these different view components defined (e.g. [`TitleView`](https://github.com/artsy/emission/blob/master/src/lib/Containers/Inquiry.tsx#L40), [`Header`](https://github.com/artsy/emission/blob/master/src/lib/Containers/Inquiry.tsx#L28), and [`Content`](https://github.com/artsy/emission/blob/master/src/lib/Containers/Inquiry.tsx#L56)). These are then used in [the render method](https://github.com/artsy/emission/blob/master/src/lib/Containers/Inquiry.tsx#L200-L240). 

You can also see that there is a custom Style array being used, because there was an exception with the keyboard avoiding button view, that needed the styles passed through. But generally, that's left for special cases.

mennenia:
Also, I'm not sure what's up with CI. It's requesting a Danger API key? @orta, soz, would you know anything about this?

src/lib/Components/Bidding/Components/Title.tsx
sepans: Came across things like this in the codebase. Not sure what the benefits are vs using `Serif`:
```
import styled from "styled-components/native"
import { Fonts } from "lib/data/fonts"

const Title = styled.Text`
  font-family: "${Fonts.GaramondRegular}";
  flex: 1;
  font-weight: bold;
  margin: 18px;
  text-align: center;
  font-size: 20;
`
```

orta:
It's a cross-fork PR, which we don't normally do on mobile repos ^

mennenia:
Ohhh of course. I didn't realise that would cause this Danger error. Thanks @orta!

src/lib/Components/Bidding/Components/Title.tsx
sepans: Also noticed that the font for the header is `AGaramondPro-Semibold` vs serif that has `AGaramondPro-Regular`. I guess we have to add that font because the `GaramondRegular` bold is thicker than semibold

yuki24:
ah I see. I'll close this PR and open an new, in-repo PR.

PR #1012, opened 2018-04-17
Update Emission to 1.50, using React Native 0.54.4

mennenia:
Matches @orta's PR on [Eigen](https://github.com/artsy/eigen/pull/2586). As soon as we have the Eigen release out containing AR and Buy Now fixes, we can point Eigen to this cut of Emission. 

There are some yoga changes here (due to `Yoga` being renamed to `yoga`). We can't have both in the same spec repo, so we have a backwards compatible `Yoga` podspec we're keeping around. 

🎉


Makefile
```diff
@@ -58,38 +58,39 @@ oss:
...
+
 	@echo "Updating Artsy specs repo";
 	pod repo update artsy
 
 	@echo "Creating folder in artsy specs repo";
 	mkdir ~/.cocoapods/repos/artsy/React/$(RNVERSION)
-	mkdir ~/.cocoapods/repos/artsy/yoga/$(YOGAVERSION)
-	mkdir ~/.cocoapods/repos/artsy/Folly/$(FOLLYVERSION)
-	mkdir ~/.cocoapods/repos/artsy/DoubleConversion/$(DOUBLECONVERSIONVERSION)
-	mkdir ~/.cocoapods/repos/artsy/glog/$(GLOGVERSION)
```
orta: This code was wrong, we didn't need to ship them to our specs repo because they existed inside the public specs repo

Makefile
```diff
@@ -58,38 +58,39 @@ oss:
...
 	cd ~/.cocoapods/repos/artsy && git add .
 	cd ~/.cocoapods/repos/artsy && git commit -m "Shipping a new version of the React deps: v$(RNVERSION) for Emission deploys"
 	cd ~/.cocoapods/repos/artsy && git push
+
+# Use this PR for yoga podspec
+# https://github.com/facebook/react-native/pull/18492
+ship_yoga:
+	# curl $(YOGA_MASTER_PODSPEC_URL) > ./externals/yoga/yoga.podspec
+	INSTALL_YOGA_WITHOUT_PATH_OPTION="true" pod ipc spec ./externals/yoga/yoga.podspec > ./externals/yoga/yoga.podspec.json
+	# rm -f ./externals/yoga/yoga.podspec
```
orta: The 0.54.2 version of RN does not include https://github.com/facebook/react-native/pull/18492 - so, we use the master version and run that through the pod IPC command to generate JSON to put in this repo

orta:
Ok, then this is good to go.

alloy:
👍

PR #1014, opened 2018-04-18
[WIP] Load Bid prices dynamically

sepans:
Loads bid increments from graphql for SelectMaxBid
https://artsyproduct.atlassian.net/browse/PURCHASE-39


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
```diff
@@ -1,45 +1,27 @@
...
-  },
-  {
-    label: "$55,000 USD",
-    value: 5500000,
-  },
-]
-
 export class SelectMaxBid extends React.Component<SelectMaxBidProps> {
   render() {
+    // TODO metaphysics should return formatted values
```
sweir27: We consume this endpoint from many places so if we make an API change we'd have to be cognizant of backwards-compatibility

ashfurrow: Yeah agreed. [The code](https://github.com/artsy/metaphysics/blob/fa05b4c5194168d7b8ddfd8f6e1ac6216e621305/src/schema/sale_artwork.js#L42) specifies bid increments as a list of floats, which was [added here, two years ago](https://github.com/artsy/metaphysics/commit/12c378ca5a11394c72334a748f7e59eeae2fdf6a#diff-ce3847826391a6fd4d78314bb78f4dd2R218). It would be better, I think, to use [`money`](https://github.com/artsy/metaphysics/blob/fa05b4c5194168d7b8ddfd8f6e1ac6216e621305/src/schema/fields/money.js) to avoid having Emission format the data in any way. Maybe we could add a new field (`bid_increment_moneys` or similar?) and deprecate `bid_increments`? Any thoughts?

src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
```diff
@@ -1,45 +1,27 @@
...
+    const bids = this.props.sale_artwork.bid_increments.map(d => ({ label: d.toString(), value: d }))
     return (
       <Container>
         <Title>Your max bid</Title>
 
-        <MaxBidPicker selectedValue={4500000} bids={Bids} />
+        <MaxBidPicker selectedValue={4500000} bids={bids} />
```
sweir27: If you want to add slightly to the scope of this PR, the initial `selectedValue` should be the first increment in the list, and subsequent selections should be based on what the user picks

ashfurrow: We can probably change the `selectedValue` to be the first of the bids.

ashfurrow: Ha, Sarah beat me to it!

yuki24: I added `selectedValue` just to make the initial screen look like the mockup (sorry for the noisy and confusing interface). I'd be happy to drop it and change `<MaxBidPicker>` to default to the first increment.

ashfurrow: Makes sense!

src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
sweir27: Formatted values sound 👌! Naming is hard, I might lean towards something like `bid_increments_formatted`? But we can debate on the corresponding metaphysics PR.

ashfurrow: Okay!

src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
sepans: Actually it defaults to the first value when `selectedValue` is not passed. So I guess I will just remove that.

sepans:
Getting this warning. Am I overlooking something? 🤔 

screen shot 2018-04-18 at 4 37 32 pm

PR #1015, opened 2018-04-18
Implements navigator for bid flow.

ashfurrow:
This pull request is for https://artsyproduct.atlassian.net/browse/PURCHASE-41 . It replaces the "hello, world" component for the bid flow with a [navigator component](https://facebook.github.io/react-native/docs/navigatorios.html) that will give us the push/pop navigation stack that iOS developers are used to. The way it works is that each component rendered with it will get passed props to manipulate the navigation stack ([here is an example](https://github.com/artsy/emission/blob/master/src/lib/Components/Consignments/Screens/Welcome.tsx#L27-L30)). 

This PR also changes `BidFlow.tsx` to not use default exports, which we're trying to move away from.


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

- `src/lib/Components/Bidding/Screens/__tests__/BidFlow-tests.tsx`

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

ashfurrow:
Oh hey whoops, now that that the component isn't hello world it should have tests.

src/lib/Components/Bidding/Screens/__tests__/BidFlow-tests.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react"
+import "react-native"
+import * as renderer from "react-test-renderer"
+
+import { BidFlow } from "../BidFlow"
+
+it("renders properly", () => {
+  const bg = renderer.create(<BidFlow saleArtworkID="5aada729139b216c0bf18103" />).toJSON()
+  expect(bg).toMatchSnapshot()
```
ashfurrow: Okay! So now that the `BidFlow` component is going to _use_ the `SelectMaxBid` component, this poses a challenge with #1014. We'll need to modify the tests here to have an initial GraphQL state once that PR is merged. I'm happy to take care of that, or pair. 

/cc @sepans

ashfurrow:
I've WIP'd this until #1014 gets merged, since it'll have implications on this PR (that I mention above).

src/lib/Components/Bidding/Screens/BidFlow.tsx
```diff
@@ -1,12 +1,26 @@
...
-export default class BidFlow extends React.Component<BidFlowProps> {
+export class BidFlow extends React.Component<BidFlowProps> {
   render() {
-    return <SerifText>This will be the home of the new bidding flow.</SerifText>
+    return (
+      <NavigatorIOS
+        navigationBarHidden={true}
+        initialRoute={{
+          component: SelectMaxBid,
+          title: "Your max bid", // title is required, though we won't likely use/display it.
```
yuki24: I wonder if there's any way to pass this `title` in to each component so we don't have to hard-code them (though I think there's some trede-off here)

src/lib/Containers/BidFlow.tsx
```diff
@@ -0,0 +1,39 @@
...
+    )
+  }
+}
+
+export default createFragmentContainer(
+  BidFlow,
+  graphql`
+    fragment BidFlow_sale_artwork on SaleArtwork @relay(mask: false) {
+      bid_increments
+      ...SelectMaxBid_sale_artwork
```
ashfurrow: Okay so the two lines above are a problem. `...SelectMaxBid_sale_artwork` should spread on that fragment, which includes `bid_increments`. But if we remove `bid_increments` from the line above, they aren't fetched. It is as though the `SelectMaxBid_sale_artwork` fragment isn't being spread correctly.

I'll keep looking into this, but I haven't found any promising leads. @sarahscott @orta any ideas here?

ashfurrow:
Okay so after looking with @l2succes, our hypothesis is that we've hit an edge case between Relay and GraphQL for lists of non-object scalars. `bid_increments` is `GraphQLList(GraphQLFloat)`, which is odd, and we were planning to move away from that anyway, so I'm going to put this PR on hold while I make that change to Metaphysics.

ashfurrow:
Okay, this is ready for review @yuki24. Unfortunately I couldn't get the Relay fragments to work as intended, but made a workaround suggested by @l2succes here: https://github.com/artsy/emission/pull/1015/commits/8cdc09b460c35d4419bec563c858b5345c432c5e Once this is merged, we can look at doing things the "right way", but this PR is blocking a lot of work, so I'd prefer to get it merged and improve it later. 

Here's an update screenshot of what's going on:

screen shot 2018-04-20 at 12 11 04 pm

sweir27:
Coool! The workaround seems like a 👍plan to unblock other work, thanks!

ashfurrow:
Okay, going to self-merge this.

PR #1019, opened 2018-04-23
Fixes bid picker not staying on selected index

ashfurrow:
I noticed that after selecting a new bid increment, that the picker would always snap back. This PR adds state to the `SelectMaxBid` component. I think that it makes sense to put the state here, since we'll need to access it to forward on to the subsequent confirmation screen. But open to suggestions!

This work was done in the course of https://artsyproduct.atlassian.net/browse/PURCHASE-41

### DIFF
--- a/src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
+++ b/src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
@@ -14,7 +14,14 @@ interface SelectMaxBidProps extends ViewProperties {
   sale_artwork: SelectMaxBid_sale_artwork
 }
 
-export class SelectMaxBid extends React.Component<SelectMaxBidProps> {
+interface SelectMaxBidState {
+  selectedBidIndex: number
+}
+
+export class SelectMaxBid extends React.Component<SelectMaxBidProps, SelectMaxBidState> {
+  state = {
+    selectedBidIndex: 0,
+  }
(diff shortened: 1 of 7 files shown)


src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
```diff
@@ -26,7 +33,11 @@ export class SelectMaxBid extends React.Component<SelectMaxBidProps> {
       <Container>
         <Title style={Margins.m1}>Your max bid</Title>
 
-        <MaxBidPicker bids={bids} />
+        <MaxBidPicker
+          bids={bids}
+          selectedBidIndex={this.state.selectedBidIndex}
+          onSelectNewBidIndex={index => this.setState({ selectedBidIndex: index })}
```
yuki24: Is there any reason we can't just use `onValueChange` or `selectedValue ` directly, rather than adding a new interface?

```jsx
<MaxBidPicker 
  bids={bids}
  onValueChange={(_, index) => this.setState({ selectedBidIndex: index })}
  selectedValue={this.state.selectedBidIndex}
/>
```

ashfurrow: Mmmm that would be because I didn't know about them. Thanks!

yuki24:
LGTM! :+1:

PR #1022, opened 2018-04-24
Confirmation screen transition

### LINKED ISSUE #1018: Relay fragment not spreading
As discussed in #1015, we're seeing odd behaviour with our new bid flow components. We have  a [`BidFlow` component](https://github.com/artsy/emission/blob/33c60f237d16e76dcb6a54c51a30d9ec1306c904/src/lib/Containers/BidFlow.tsx) and it contains a [`SelectMaxBid` component](https://github.com/artsy/emission/blob/33c60f237d16e76dcb6a54c51a30d9ec1306c904/src/lib/Components/Bidding/Screens/SelectMaxBid.tsx), with two respective fragments:

```tsx
fragment BidFlow_sale_artwork on SaleArtwork {
  ...SelectMaxBid_sale_artwork
}
```

```tsx
fragment SelectMaxBid_sale_artwork on SaleArtwork {
  increments {
    display
    cents
  }
}
```

The problem is, when using the `BidFlow_sale_artwork` fragment in our [bid flow query renderer](https://github.com/artsy/emission/blob/33c60f237d16e76dcb6a54c51a30d9ec1306c904/src/lib/relay/QueryRenderers.tsx#L65-L82), the `increments` are not marshalled into the component tree. After debugging with @l2succes, we concluded that the request _was_ getting made correctly to metaphysics, but Relay wasn't marshalling the data as we expected.

The workaround is to use `@relay(mask: false)` when spreading over the component: 

https://github.com/artsy/emission/blob/33c60f237d16e76dcb6a54c51a30d9ec1306c904/src/lib/Containers/BidFlow.tsx#L35

We merged with the workaround because that PR was blocking lots of other work on this feature.

This approach isn't going to scale well and we should prioritize getting to the root of this before shipping this feature. I am tracking this ticket in our sprint work here: https://artsyproduct.atlassian.net/browse/PURCHASE-64 Any suggestions or advice would be welcome.

ashfurrow:
Okay so apologies for the large PR. This work represents the actual transition from the bid-picking screen to the bid-confirmation screen, but also cleans up our use of query renderers as our understanding of Relay has grown. 

The Relay issue I was experiencing was solved rather simply. I was trying to pass the Relay context on to `SelectMaxBid` instead of `MaxBidScreen` (which wraps `SelectMaxBid` in a Relay fragment container). So it was a component type mismatch. Fixes #1018. 

The bid confirmation screen now loads actual data over Relay from metaphysics, instead of hardcoded data. Note that in order for Relay to work, the `SelectMaxBid_sale_artwork` fragment **must** also spread over `ConfirmBid_sale_artwork`. I initially thought that spreading over `ConfirmBid_sale_artwork` in the higher-level `BidFlow_sale_artwork` fragment would be sufficient, but the lesson is: spread over the fragment when directly using a child component, rather than at a higher level.

This PR also moves the `SelectMaxBidRenderer` out of `QueryRenderers.tsx` and into the story-based flow, as it shouldn't be used by anything but the stores. This is consistent with how other parts of emission structure query renderers, but as the number of components grows, it might make sense to move them into their own file.

The work is tracked in Jira at https://artsyproduct.atlassian.net/browse/PURCHASE-41 and https://artsyproduct.atlassian.net/browse/PURCHASE-58


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Containers/BidFlow.tsx
```diff
@@ -32,7 +32,7 @@ export default createFragmentContainer(
   BidFlow,
   graphql`
     fragment BidFlow_sale_artwork on SaleArtwork {
-      ...SelectMaxBid_sale_artwork @relay(mask: false)
```
sepans: 👍

sweir27:
@ashfurrow this is awesome! Great to see all the pieces fitting together 🎉

src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
```diff
@@ -39,7 +51,7 @@ export class SelectMaxBid extends React.Component<SelectMaxBidProps, SelectMaxBi
           selectedValue={this.state.selectedBidIndex}
         />
 
-        <Button text="NEXT" onPress={() => null} />
+        <Button text="NEXT" onPress={this.onPressNext} />
```
orta: careful with using this, I think it may need `this` re-binding, I know we have that in many other places

PR #1023, opened 2018-04-24
Adds note about emission-nebula.

ashfurrow:
We should point this repo to emission-nebula for the next time someone goes looking for an answer to "how is Emission deployed to TestFlight?"


sarahscott:
👍

PR #1024, opened 2018-04-24
Updates logging in native code to use NSNumber instead of platform-specific casts

ashfurrow:
`NSLog` uses C-style format strings, which works well except for primitives whose size is platform-dependent. This was causing an issue with our automated Emission beta deploys (updated Xcode version, which was erroring out on these lines). My solution is to bypass casting altogether and box the primitive values in an `NSNumber` by wrapping them in an object literal expression (the `@` sign) and then letting `toString` figure out how to print them.


ashfurrow:
Looks like prettier stripped a bunch of whitespace too, nice.

PR #1025, opened 2018-04-25
Bidder position mutation

sepans:
https://artsyproduct.atlassian.net/browse/PURCHASE-59

When bid is high enough the flow is working.

screen shot 2018-04-25 at 10 14 09 am

~~But when it is not by some reason I cannot catch the error and forward to the right screen. I don't know if it is a front end problem or [metaphysics mutation](https://github.com/artsy/metaphysics/blob/master/src/schema/me/bidder_position_mutation.js) doesn't handle the error correctly.~~

 **(fixed)**

screen shot 2018-04-25 at 10 14 09 am


src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -15,10 +16,61 @@ interface ConfirmBidProps extends ViewProperties {
     display: string
     cents: number
   }
+  relay?: RelayPaginationProp
+  navigator?: NavigatorIOS
 }
 
 export class ConfirmBid extends React.Component<ConfirmBidProps> {
+  placeBid() {
```
sweir27: We may want to replicate some of force's logic here, which is a little more complex than just placing a mutation given that we have to wait for gravity's bids to be processed before we can show you the results screen.

(for reference, the logic is [here](https://github.com/artsy/force/blob/master/src/desktop/apps/auction_support/client/bid_form.coffee#L44-L84))

- First, it does a [client-side check](https://github.com/artsy/force/blob/master/src/desktop/apps/auction_support/client/bid_form.coffee#L49) to make sure that your bid isn't below what we're minimally accepting (I think this case is fairly rare, but if you'd been hanging out on the bid picker screen while bidding advanced much farther, it's possible).
- Then it sends the request to gravity and [shows an error](https://github.com/artsy/force/blob/master/src/desktop/apps/auction_support/client/bid_form.coffee#L61) if there was an error placing your bid (I don't believe being "outbid" counts as an error here-- it's more if the sale is closed to bidding or there is an internal error)
- Then we poll for the bid to be processed. After that, we can [tell you that you have been outbid](https://github.com/artsy/force/blob/master/src/desktop/apps/auction_support/client/bid_form.coffee#L77)

In this scenario, the mutation will likely not return an error for an outbid case (since the bidder position _was_ successfully created), but when you fetch the bidder position during polling from the `me/bidder_positions` endpoint you should be able to retrieve the status.

ashfurrow: This is a good point; I think we should move as much of this logic into metaphysics as possible, though. If I understand the GraphQL approach to error-handling, the mutation should never produce an error, but always a result.

sweir27:
Makes sense about trying to move the bid status logic to MP @ashfurrow based on the values that are returned from gravity! It always seemed weird to me that force overrode them with custom values (there may be some underlying reason for this, but seems like we could re-think). 

Where does eidolon implement logic for waiting for the bid to be processed?

sweir27:
Oh I think I found it [here](https://github.com/artsy/eidolon/blob/master/Kiosk/Bid%20Fulfillment/BidCheckingNetworkModel.swift).

The only tricky thing is that _creating_ the bidder position the first time won't necessarily return an error from gravity in the normal outbid case. It's subsequent fetches to the `/me/bidder_position/:id` endpoint that will return the correct status.

ashfurrow:
@sweir27 Agreed, the polling is tricky. I don't think it belongs in metaphysics; but whatever we _can_ put there, I'd like to.

sepans:
Now polls MP until `processed_at` flag is on as @sweir27 pointed above. Still cannot catch the MP error to avoid the red error screen. This [line](https://github.com/artsy/emission/pull/1025/files#diff-8a3a841add9adeac68c88de41a452cc4R72) should handle it but it doesn't. 🤔 .

Also the bid increments don't get updated after creating a bidder position after refreshing the page in emission but it works fine when directly querying MP so it seems to be a emission cache.

joeyAghion:
Note that reserve status is orthogonal to the bidder's own status w.r.t. the lot. (I.e., anyone can view that a lot has met reserve, has not met reserve, or has an unknown reserve. And any of these can be true while the current bidder is winning _or_ outbid.)

sepans:
Added handling error messages from MP and rebased. 

The red error message above is fixed.

**TODO:** parse markdown and display links. (will do it in another PR since this one already has become too long)

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -71,9 +183,12 @@ export class ConfirmBid extends React.Component<ConfirmBidProps> {
               You agree to <LinkText onPress={this.onPressConditionsOfSale}>Conditions of Sale</LinkText>.
             </Serif14>
 
-            <Flex m={4}>
-              <Button text="Place Bid" onPress={() => null} />
-            </Flex>
+            <Button
+              text="Place Bid"
+              onPress={() => {
```
ashfurrow: I'm curious, since I'm still getting oriented around JS, could this be rewritten as `onPress={this.placeBid}`? Or is there some JS binding problem?

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -30,12 +33,121 @@ interface ConfirmBidProps extends ViewProperties {
...
+            bidder_position(id: "${positionId}") {
+              processed_at
+              is_active
+            }
+          }
+        }
+      `
+      const interval = setInterval(() => {
+        metaphysics({ query }).then(this.checkBidPosition.bind(this))
+      }, 2000)
```
ashfurrow: Can we double-check the desired interval? [Force uses 1000](https://github.com/artsy/force/blob/316894dc4a2ae6aa4723c730fc9635d897f82884/src/desktop/models/sale_artwork.coffee#L10) and so does Eidolon, but both of those frontends _are_ typically accessed over wifi and not 3G. In any case, let's describe why we picked that number in a comment.

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -30,12 +33,121 @@ interface ConfirmBidProps extends ViewProperties {
...
+    this.props.navigator.push({
+      component: BidResult,
+      title: "",
+      passProps: {
+        message_header: messageHeader,
+        message_description_md: messageDescriptionMd,
+        winning,
+      },
+    })
+  }
```
ashfurrow: Should we move the bid polling logic into its own file, so we can extensively unit test it? Or is it common in React to have components do networking stuff like this?

In either case, we should add some tests for the polling code. Maybe just a `#TODO` for now?

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
sepans: I think `onPress={() => this.placeBid()}` is similar to `onPress={this.placeBid.bind(this)}` (to bind `this` back to the react component object instead of event handler's `this`) but there has been ongoing debate about what is the best way to do that. I haven't seen any objection to `onPress={() => this.placeBid()}` recently but doing `bind` on each render is not great 😆

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
sepans: Thats a good idea. I will put a TODO note and do that in a follow up PR.

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
sepans: Thanks for looking it up. 👍

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -30,12 +33,122 @@ interface ConfirmBidProps extends ViewProperties {
...
+        // wining
+        this.showBidResult(true)
+      } else {
+        // outbid
+        this.showBidResult(false)
+      }
+    } else {
+      // poll again
+      if (this.state.pollCount > MAX_POLL_ATTEMPTS) {
+        clearInterval(this.state.intervalToken)
```
sweir27: Can we add a TODO here to insert the "Bid result not available" screen if we've passed the max number of attempts? I'm not sure if that's implied by this.

src/lib/Components/Bidding/__stories__/BidFlow.story.tsx
```diff
@@ -56,24 +56,28 @@ const ConfirmBidScreenRenderer: React.SFC<any> = ({ render, saleArtworkID }) =>
...
   .add("Select Max Bid", () => (
-    <SelectMaxBidRenderer render={renderWithLoadProgress(MaxBidScreen)} saleArtworkID="5aada729139b216c0bf18103" />
+    <SelectMaxBidRenderer render={renderWithLoadProgress(MaxBidScreen)} saleArtworkID="5ae72c86ae68a93562b5706f" />
   ))
   .add("Confirm Bid", () => {
     return (
       <ConfirmBidScreenRenderer
         render={renderWithLoadProgress(ConfirmBidScreen, { bid: { display: "$20,000", cents: 2000000 } })}
-        saleArtworkID="5aada729139b216c0bf18103"
+        saleArtworkID="5ae72c86ae68a93562b5706f"
```
sweir27: #minor (and not related to this PR per se), but since we re-set staging each week, this ID is going to become outdated 😞 (I'm assuming that this pulls real staging data into the storybooks, right?)

sweir27:
One #minor comment, but I think this looks good! Once the conflicts are resolved I think we should merge 🎉 .

PR #1028, opened 2018-04-26
Corrects Changelog header format

ashfurrow:
The changelog has historically used an H2 for master and H3's for releases. But this convention was broken [in this release](https://github.com/artsy/emission/commit/eaebfb8e9d8066d80993478d487d848e02c1bca2), which our automated deploy script relies on:

https://github.com/artsy/emission-nebula/blob/f8bd8465cd6e2e13038020099119dfaef22c983d/fastlane/Fastfile#L61

This PR does two things:

- Corrects the 1.4.6 header.
- Changes master to use H3 too, so this won't be likely to happen again.

If merged, I'll make a corresponding change in emission-nebula to make sure this still works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 -->
 
-## Master
+### Master
 
 * Adds Recently Viewed Works rail to Home - sarah
 * Upgrade to React Native 0.54.4 - alloy/orta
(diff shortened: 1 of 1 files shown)


yuki24:
sorry for the delay, but merging!

ashfurrow:
Okay, made that change to emission-nebula: https://github.com/artsy/emission-nebula/commit/a0d68054b4a9a66cff0812fc6a783e1b905e66c8 I didn't use a PR to avoid triggering a CI build, and it's a trivial change. Thanks Yuki!

PR #1032, opened 2018-05-01
Adds "Your billing address" screen

yuki24:
This PR adds a "Your bidding address" screen.

 * [x] ~Add a transition to the input border color~ will be addressed as a separate ticket
 * [x] ~Add a `<Dropdown>` component~ we decided to remove country from the form

## Text input

screen shot 2018-05-01 at 2 11 23 pm

## Your bidding address screen

screen shot 2018-05-03 at 11 36 40 am

#skip_new_tests


src/lib/Components/Bidding/Screens/BillingAddress.tsx
```diff
@@ -0,0 +1,48 @@
...
+            <Input placeholder="Enter your street address" mb={4} />
+
+            <Serif16 mb={2}>Address line 2 (optional)</Serif16>
+            <Input placeholder="Enter your apt, floor, suite, etc." mb={4} />
+
+            <Serif16 mb={2}>City</Serif16>
+            <Input placeholder="Enter city" mb={4} />
+
+            <Serif16 mb={2}>State, Provice, or Region</Serif16>
+            <Input placeholder="Enter state, provice, or region" mb={4} />
```
ashfurrow: "provice" -> "province"

yuki24:
This PR is ready for review, but we may have to backfill some tests before merging.

yuki24:
This probably needs a CHANGELOG entry, I'll update the PR later.

yuki24:
Alright, added a CHANGELOG entry and a test. Now this is really ready for a review!

src/lib/Components/Bidding/Elements/Input.tsx
```diff
@@ -0,0 +1,37 @@
...
+  extends BorderColorProps,
+    BorderProps,
+    ColorProps,
+    FlexboxProps,
+    FontSizeProps,
+    HeightProps,
+    SpaceProps,
+    WidthProps {}
+
+export const Input = styled.TextInput.attrs<InputProps>({})`
```
ashfurrow: Should we add a comment above this component, and the other `Input`, describing their purpose? I'm not clear on the difference now, between a component and an element.

src/lib/Components/Bidding/Components/Input.tsx
```diff
@@ -0,0 +1,46 @@
...
+        fontSize={3}
+        onBlur={() => this.onBlur()}
+        onFocus={() => this.onFocus()}
+        p={3}
+        pb={2}
+        {...this.props}
+      />
+    )
+  }
+}
```
ashfurrow: With this new component, could we add a few tests? It's got state and behaviour that we should have covered. I know the "best way" for Emission to be tested is still undecided, but even snapshot tests would be better than nothing.

src/lib/Components/Bidding/Components/Input.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { Component } from "react"
+import { TextInputProperties } from "react-native"
+import { Input as StyledInput, InputProps as StyledInputProps } from "../Elements/Input"
+
+interface InputProps extends StyledInputProps, TextInputProperties {
+  error?: boolean
+}
+
+class InputStete {
```
sweir27: `InputState`?

yuki24: 😅

src/lib/Components/Bidding/Elements/Input.tsx
yuki24: The responsibility of the components in the `elements` dir is to expose `styled-system`'s interface while the components in `components` dir can be opinionated about how we use them in the bidding flow. Right now, the `Grid` component is the only one that goes against this philosophy. We could move it to the `components` if we want to make the boundary more clear. We could also add a prefix to make them more self-documenting.

yuki24:
fixed the typos and added tests!

ashfurrow:
@yuki24 I'm taking a final look now. For PRs, would you mind waiting until we want to merge before squashing commits? I like clean git histories too, but squashing early makes it impossible to see what's changed since the last time I reviewed the PR.

src/lib/Components/Bidding/Components/Input.tsx
ashfurrow: Following up on this comment, in `Input-tests.tsx` I think we should add tests for different border colours and the presence of an error and for focusing.

yuki24: It always sounds good to say "Add more tests", but sometimes it's a trap. I don't find snapshot testing very useful, and tests that are not helpful are liabilities we should have less, so I added more explicit tests. Even these tests could be liabilities because they depend on implementation details (`onBlur` and `onFocus`).

src/lib/Components/Bidding/Screens/BillingAddress.tsx
```diff
@@ -0,0 +1,46 @@
...
+            <Serif16 mb={2}>Address line 1</Serif16>
+            <Input placeholder="Enter your street address" mb={4} />
+
+            <Serif16 mb={2}>Address line 2 (optional)</Serif16>
+            <Input placeholder="Enter your apt, floor, suite, etc." mb={4} />
+
+            <Serif16 mb={2}>City</Serif16>
+            <Input placeholder="Enter city" mb={4} />
+
+            <Serif16 mb={2}>State, Provice, or Region</Serif16>
```
sweir27: Eep didn't notice this before! But `Provice` --> `Province`

src/lib/Components/Bidding/Components/Input.tsx
ashfurrow: Okay, thanks. I'll take a look at the tests. I agree that snapshot tests aren't _that_ useful, but I have to stress that Emission's test coverage is pretty poor right now. The Purchase team has the opportunity to lead by example, to figure out what a meaningful test looks like for this type of component, etc. This is part of the work I've done gathering examples here: https://github.com/artsy/emission/pull/1034/files#diff-b5e6d39e5fb762a421d5bc45ebba4a0eR37

yuki24:
> waiting until we want to merge before squashing commits?

I've added a separate commit for a review. I totally hear you, but this would make the feedback loop unnecessarily larger since it may make a single merge button click a process that requires asking to squash, squashing and pushing, and finally merging...

src/lib/Components/Bidding/Screens/BillingAddress.tsx
yuki24: 😩

yuki24:
I'm closing this PR to send a PR against `develop-1.5.x`.

src/lib/Components/Bidding/Screens/BillingAddress.tsx
orta: I use an amazing spellchecker! Maybe there's a port for webstorm? https://github.com/Jason-Rev/vscode-spell-checker

yuki24: IntelliJ does check spelling! I should just pay more attention to its suggestions...

PR #1033, opened 2018-05-01
Show conditions of sale

ashfurrow:
This is one of two required pull requests for conditions of sale. The other is here: https://github.com/artsy/eigen/pull/2599 This is a WIP because we still need to scope the `/conditions-of-sale` route to _only_ condition of sale links from Emission components (we'll add a query string to avoid other links to `/conditions-of-sale` being presenting modally).

Work tracked against [PURCHASE-60](https://artsyproduct.atlassian.net/browse/PURCHASE-60).

### DIFF
--- a/Example/Emission/AppDelegate.m
+++ b/Example/Emission/AppDelegate.m
@@ -181,6 +181,7 @@ randomBOOL(void)
     }
     UIViewController *viewController = [self viewControllerForRoute:route];
     UINavigationController *navigationController = [[EigenLikeNavigationController alloc] initWithRootViewController:viewController];
+    navigationController.navigationBarHidden = NO;
     viewController.navigationItem.leftBarButtonItem = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemDone
                                                                                                     target:self
                                                                                                     action:@selector(dismissModalViewController)];
@@ -251,7 +252,7 @@ randomBOOL(void)
 - (void)dismissModalViewController;
 {
   UINavigationController *navigationController = (UINavigationController *)self.window.rootViewController;
-  [navigationController.visibleViewController.navigationController dismissViewControllerAnimated:YES completion:nil];
+  [navigationController dismissViewControllerAnimated:YES completion:nil];
 }
 
 - (void)downloadPRBuildAndLoginWithUserID:(NSString *)userID
--- a/Example/Emission/UnroutedViewController.m
+++ b/Example/Emission/UnroutedViewController.m
@@ -37,7 +37,7 @@
 
   BOOL useStaging = [[NSUserDefaults standardUserDefaults] boolForKey:ARUseStagingDefault];
   NSURL *url = [NSURL URLWithString: useStaging ? @"https://staging.artsy.net" : @"https://artsy.net"];
-  return [url URLByAppendingPathComponent:route];
+  return [NSURL URLWithString:route relativeToURL:url];
 }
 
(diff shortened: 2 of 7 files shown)


Example/Emission/AppDelegate.m
```diff
@@ -181,6 +181,7 @@ - (void)setupEmissionWithUserID:(NSString *)userID accessToken:(NSString *)acces
     }
     UIViewController *viewController = [self viewControllerForRoute:route];
     UINavigationController *navigationController = [[EigenLikeNavigationController alloc] initWithRootViewController:viewController];
+    navigationController.navigationBarHidden = NO;
```
ashfurrow: Since the nav bar was always hidden, the left bar button item set on the next line was never visible.

Example/Emission/AppDelegate.m
```diff
@@ -251,7 +252,7 @@ - (UIViewController *)viewControllerForRoute:(NSString *)route;
 - (void)dismissModalViewController;
 {
   UINavigationController *navigationController = (UINavigationController *)self.window.rootViewController;
-  [navigationController.visibleViewController.navigationController dismissViewControllerAnimated:YES completion:nil];
+  [navigationController dismissViewControllerAnimated:YES completion:nil];
```
ashfurrow: The above code was causing a crash. It looks like it was never/seldom called, so it's likely been a but in the Emission test app. Generally speaking, UIKit handles walking up the view controller hierarchy to dismiss view controllers for us automatically.

Example/Emission/UnroutedViewController.m
```diff
@@ -37,7 +37,7 @@ - (NSURL *)urlForRoute:(NSString *)route
 
   BOOL useStaging = [[NSUserDefaults standardUserDefaults] boolForKey:ARUseStagingDefault];
   NSURL *url = [NSURL URLWithString: useStaging ? @"https://staging.artsy.net" : @"https://artsy.net"];
-  return [url URLByAppendingPathComponent:route];
+  return [NSURL URLWithString:route relativeToURL:url];
```
ashfurrow: Since the new route contains a query string, it isn't strictly a path component. This moves to `NSURL`'s recommended way to make a relative URL absolute.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

Example/Emission/AppDelegate.m
orta: Pretty sure this is to support having storybooks with navigation elements inside them

ashfurrow: Okay, I'll take a look at see if this breaks things.

ashfurrow: So this code was [added two years ago](https://github.com/artsy/emission/commit/ce2a6f376ff4cbb13ebd5fcd100fcc3554fd0179) strictly as a mechanism for handling `presentModalViewController` from JS. I've looked at other cases where `presentModalViewController` is used and this still works. I think we're 👍 on this change.

This is also only in the Emission Example app, not the pod, so anything we inadvertently break is only visible to people testing the Emission beta app.

mennenia: :+1:

mennenia:
Looks great!

PR #1035, opened 2018-05-04
[Home/ForYou] Fix crash when following artist from ‘related artist’ rail

alloy:
EV-117 #merged

Due to some messy parts of our schema around the home page artwork modules and their context, we now have some context types that have the `__id` field for no good reason. However, our current Relay version will query for that regardless, if it exists, and this leads to `followedArtistContext` always being an object, albeit sometimes with nothing more than `{ __id: null }`.

There is no good way to fix this in MP without breaking schema compatibility, this will at least fix the issue going forward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 * Adds a new color `red100` (`#F7625A`) to the theme - yuki24
 * Adds a `<CheckBox>` component - yuki24
 * Adds support for viewing conditions of sale - ash
+* Fixes crash when following an artist from the ‘related artist’ Home/ForYou rail - alloy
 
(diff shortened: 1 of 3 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  Missing Test Files:

- `src/lib/Scenes/Home/Components/ForYou/Components/__tests__/ArtworkCarouselHeader-tests.tsx`

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

orta:
👍

PR #1036, opened 2018-05-04
[git] Hide query maps from GitHub diffs.

alloy:
#trivial

Noticed these files showing diffs in #1025, so omitting those now too.


orta:
nice one

PR #1037, opened 2018-05-04
@ashfurrow => Merge Emission crasher into a branch for 1.4.7

orta:
Fixin that changelog now

Contains: #1035

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,33 +14,10 @@
 
 -->
 
-## 1.4.6
 ### Master
 
-* Adds Recently Viewed Works rail to Home - sarah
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,7 +14,13 @@
 
 -->
 
-## 1.4.6
+### Master
+
+## 1.4.7
```
ashfurrow: This needs a triple `#` for emission-nebula. We changed this recently, the current code is here: https://github.com/artsy/emission-nebula/blob/904321ac7a8becb79a91b240a09233c416dbd959/fastlane/Fastfile#L63

I will send a Danger rule requiring this.

PR #1039, opened 2018-05-07
Simple native markdown renderer

sepans:
Parses and renders markdown to handle markdown messages from [metaphysics](https://github.com/artsy/metaphysics/pull/1031).

Currently only works for **text (paragraphs)** and **links** to avoid over generalization.

screen shot 2018-05-07 at 2 06 59 pm

Markdown is parsed using [Simple Markdown](https://github.com/Khan/simple-markdown) which allows overriding markdown rendering rules to customize it for emission native components. Also unlike other markdown parsers it can return react object instead of string which is very useful here since react native doesn't have something like `dangerouslySetInnerHTML`.

### DIFF
--- a/src/lib/Components/Bidding/Components/MarkdownRenderer.tsx
+++ b/src/lib/Components/Bidding/Components/MarkdownRenderer.tsx
@@ -1,0 +1,71 @@
+import React from "react"
+import { Text, View } from "react-native"
+import SimpleMarkdown from "simple-markdown"
+import styled from "styled-components/native"
+
+import { Row } from "../Elements/Grid"
+
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+
+// Rules for rendering parsed markdown. Currently only handles links and text. Add rules similar to
+// https://github.com/CharlesMangwa/react-native-simple-markdown/blob/next/src/rules.js for new functionalities.
+const rules = {
+  ...SimpleMarkdown.defaultRules,
+  link: {
+    ...SimpleMarkdown.defaultRules.link,
+    react: (node, output, state) => {
+      state.withinText = true
+      let element
+      const openUrl = url => {
+        SwitchBoard.presentModalViewController(element, url)
+      }
+      return (
+        <LinkText key={state.key} onPress={() => openUrl(node.target)} ref={el => (element = el)}>
+          {output(node.content, state)}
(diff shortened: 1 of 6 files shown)


src/lib/Components/Bidding/Components/MarkdownRenderer.tsx
```diff
@@ -0,0 +1,71 @@
...
+  return rawBuiltParser(blockSource, { inline: false })
+}
+const reactOutput = SimpleMarkdown.reactFor(SimpleMarkdown.ruleOutput(rules, "react"))
+
+export const toReact = md => {
+  const syntaxTree = parser(md)
+  return reactOutput(syntaxTree)
+}
+
+interface NariveMarkdownProps {
```
orta: `NariveMarkdownProps` ->  `NativeMarkdownProps`

src/lib/Components/Bidding/Components/__tests__/MarkdownRenderer-tests.tsx
```diff
@@ -0,0 +1,12 @@
...
+import "react-native"
+import * as renderer from "react-test-renderer"
+import { MarkdownRenderer } from "../MarkdownRenderer"
+
+describe("Markdown parser", () => {
+  it("parses", () => {
+    const md = "text with some [link](http://example.com) in it"
+    const bg = renderer.create(<MarkdownRenderer md={md} />).toJSON()
+    expect(bg).toMatchSnapshot()
+  })
```
orta: I wonder if there could be a better test for this, maybe a check that `example.com` isn't included in the text because that would indicate that the MD was parsed?

sepans: 👍👍

l2succes: you could also use `enzyme` to actually look check for the link and its props. 
In a test:

```javascript
   const wrapper = mount(<MarkdownRenderer md={md} />)
   expect(wrapper.find(TextLink)).toHaveLength(1)
   ...
```

sepans: Thanks @l2succes. much better now!

PR #1042, opened 2018-05-09
Replace smart quotes with normal quotes

yuki24:
Because we don't want smart quotes to be too smart and take over the entire earth 😉

Also, `pod install` was suggesting removing it:

screen shot 2018-05-09 at 10 36 21

#trivial


orta:
Yep, thanks!

PR #1047, opened 2018-05-14
Show checkbox if you are unregistered (and verify copy)

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-114

This PR adds a checkbox to the "Confirm your bid" screen.

 * Right now it doesn't persist the flag so the checkbox will always show up, let me know if there's already a mutation endpoint we could use to do so.
 * The button is disabled by default and enabled once the user checks it off, not sure if the animation is appropriate here though. cc: @briansw
 * This also includes minor style tweaks

## Sceenshot

screen_shot_2018-05-14_at_4_08_39_pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 ### Upcoming
 
+* Adds a checkbox to the confirm your bid screen - yuki24
+
 ### 1.5.1
 
 * Adds Recently Viewed Works rail to Home - sarah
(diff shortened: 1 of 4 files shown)


sweir27:
👍 cool. From what I understand, we don't need to be saving whether or not someone checks the conditions of sale box (since it corresponds 1-1 with sale registration), but am confirming this with Katarina and Julie. She said we can move forward for now and if we end up needing to save it separately we can follow up.

yuki24:
added a changelog and updated the test and snapshot for the `<ConfirmBid>` component. Now this is ready for review.

PR #1049, opened 2018-05-15
Updates release with new script to merge query maps.

ashfurrow:
Nothing fancy, since this will be replaced by a longer-term solution on Metaphysics like a database. This is the first time the release script was run through prettier lately (ever?) so there are some superfluous changes.


alloy:
Nice 👍

PR #1006, opened 2018-04-13
Add the ability to specifically style the button headline

yuki24:
Right now, the `<Button>` component accepts the `style` prop that allows for styling the button container. But since it is [passed into the `<AnimatedView>` component](https://github.com/artsy/emission/blob/7cfcbd3c6b550925bc59bce75a66d6f2d7c3e76b/src/lib/Components/Buttons/Button.tsx#L154), it actually doesn't allow for styling [the headline in the button](https://github.com/artsy/emission/blob/7cfcbd3c6b550925bc59bce75a66d6f2d7c3e76b/src/lib/Components/Buttons/Button.tsx#L150) (or the styles on the headline are taking the precedence so we cannot override them).

In the new bidding flow, we are going to have to [set the font size of the button to `14`](https://app.zeplin.io/project/5ab94282c124db8d58365b47/screen/5ac66d5c15dd161833080c10) (which is currently [set to `12`](https://github.com/artsy/emission/blob/7cfcbd3c6b550925bc59bce75a66d6f2d7c3e76b/src/lib/Components/Text/Headline.tsx#L16) anywhere). This PR adds the ability to specifically style the button headline so we will be able to re-use it in the new bidding flow.

screen_shot_2018-04-13_at_4_39_54_pm

(in the screenshot, I deliberately used `fontSize: 18` to make the visible change clear, but in the actual flow we will use `14`)

#skip_new_tests

PR #1053, opened 2018-05-16
Add the ability to style the text in the inverted button with spinner

yuki24:
This is similar to https://github.com/artsy/emission/pull/1006, but there seems to be two different `<InvertedButton>`s, one with the ability to show a spinner and the other without. This PR adds the ability to style the text to the one with the spinner.

screen_shot_2018-04-13_at_4_39_54_pm


orta:
Ace - yep, looks good to me 👍

PR #1056, opened 2018-05-21
[Analytics] Update home analytics for tab changes and inital tab props

mennenia:
Making sure that we: 
- are dispatching analytics at the right time (when tabs are changed, and when `viewDidAppear` happens after the VC wasn't visible)
- set the right `PageName` prop by adding a separate method used by `screenTrack` and `fireHomeScreenAnalytics`

### DIFF
--- a/src/lib/Scenes/Home/index.tsx
+++ b/src/lib/Scenes/Home/index.tsx
@@ -24,8 +24,9 @@ const TabBarContainer = styled.View``
 interface Props {
   selectedArtist?: string
   selectedTab?: number
-  initialTab?: number
+  initialTab: number
   tracking: any
+  isVisible: boolean
 }
 
 interface State {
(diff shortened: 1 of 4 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

src/lib/Scenes/Home/index.tsx
```diff
@@ -37,13 +38,27 @@ const ArtistsWorksForYouTab = 0
...
+  return screenType
+}
+
 // This kills two birds with one stone:
 // It's necessary to wrap all tracks nested in this component, so they dispatch properly
 // Also, it'll only fire when the home screen is mounted, the only event we would otherwise miss with our own callbacks
-@screenTrack({
-  context_screen: Schema.PageNames.HomeArtistsWorksForYou,
+@screenTrack((props: Props) => ({
+  context_screen: screenSchemaForCurrentTab(props.initialTab),
```
ashfurrow: Wait, should this be `initialTab` or `selectedTab`?

ashfurrow:
@mennenia I've updated the changelog for this PR, you'll have to pull on your local machine before pushing any changes.

src/lib/Scenes/Home/index.tsx
mennenia: Great Q!

It should indeed `initialTab` because this event only gets fired once (and hence should reflect the first page it starts on). 

Any other changes are taken care of by the fireAnalytics method, also when changing tabs programmatically (and updating the `selectedTab` var).

PR #1057, opened 2018-05-21
Adds prop-setting

ashfurrow:
This adds a new method to `ARComponentViewController` that sets the props of the `rootView` if the view is loaded, _or_ sets the `initialProps` for the `rootView` if it hasn't been loaded yet. I don't see any Objective-C tests for Emission so I didn't add any, but something to think about longer term.

This is useful for when we want to switch tabs on the home component controller when the app is launching (for example, from a push notification).


orta:
Yeah, cool, this makes sense to me 👍

PR #1058, opened 2018-05-21
Use <ConformBid> component that does not depend on relay in storybook #trivial

yuki24:
Right now, we use the `<ConfirmBidScreen>` component that depends on relay. This has made visual testing a bit more difficult than it should since it could fail to load when the sale artwork that the hard-coded `saleArtworkId` points to does not exist. This PR will make the book more stable and easier to test by making it no longer dependent on external resources.

We may want to replace the other relay-based components with plain React components, except the `<BidFlowRenderer>` that's used in the actual flow, but let's see how this goes before we go too far.

#skip_new_tests


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

sweir27:
🙌  awesome, thanks!

PR #1059, opened 2018-05-22
[Home] Update home tab switching method 

mennenia:
Based off [@ashfurrow's comment](https://github.com/artsy/eigen/pull/2612#issuecomment-390775648), using the `Page` prop is indeed cautioned against. 

Originally we went for this, as changing tab through just the `initialPage` prop wasn't working, however we are now switching to the recommended `goToPage` function. (thanks @ashfurrow!) 

🎉


orta:
Yeah, OK, this makes sense 👍

mennenia:
I can't wait for "merge on green" functionality 💕

ashfurrow:
It's about to happen: https://github.com/artsy/peril-settings/pull/59

PR #1060, opened 2018-05-22
The <Input> component should update the border color when the parent component updates the error prop

yuki24:
This PR addresses an edge case where the parent component of an `<Input>` component updates the `error` prop but the border color wasn't updated properly.

I tried to use [the new `getDerivedStateFromProps` callback](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) but seems like we aren't on the latest version of React Native, so I just used the `componentWillReceiveProps()`. The new test does not depend on the internals of the component, so it should not be difficult to migrate to `getDerivedStateFromProps` after react upgrade.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 * Adds a checkbox to the confirm your bid screen - yuki24
 * Adds the ability to style the text in the inverted button with a spinner - yuki24
 * Now the `<ConfirmBid>` component shows a spinner while loading - yuki24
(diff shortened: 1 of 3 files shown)


alloy:
Yeah we’ll switch all of the code at once, with RN you never know what FB may decide to do and seeing as they have decided to silence the deprecation warnings even in newer versions we should take that as a sign to wait a bit more.

l2succes:
👍that makes sense

PR #1063, opened 2018-05-23
WIP: [Bidding flow] Add the ability for first-time bidders to add billing address

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-153

 * [ ] Add more test converage

## Screenshots

billing-address

PR #1064, opened 2018-05-25
PoC: Form without formik or yup

yuki24:
This PR addresses the same ticket as #1063 but without adding dependencies. This is just a prrof-of-concept, but it works exactly like #1063. Any feedback is welcome!

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,6 @@
     "relay-runtime": "https://github.com/alloy/relay/releases/download/v1.5.0-artsy.5/relay-runtime-1.5.0-artsy.5.tgz"
   },
   "dependencies": {
-    "formik": "^0.11.11",
     "lodash": "^4.17.4",
     "moment": "^2.19.4",
     "prop-types": "^15.5.10",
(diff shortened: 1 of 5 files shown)


orta:
Nice, that's like basically the same code 👍

sweir27:
This seems like a good approach! It looks like the other ticket (https://github.com/artsy/emission/pull/1063) also has dependency-free validations, but in a slightly different form (and with more granular validations). It's hard to compare them because this only includes a validation for `presence` so I'm curious how this approach scales.

Also side question-- do you feel like the form is going to become more complex and would benefit from a library in the long run? For basic stuff it makes sense to avoid the dependency, but libraries can have the added benefit of robust testing/community support/etc. 😄.

yuki24:
> this only includes a validation for presence so I'm curious how this approach scales.

The `validatePresence` function just returns `null` or a string, so you could just add more following calls to it and let the language handle the fallback:

```ts
postalCode: validatePresence(postalCode) || validateLength(postalCode, { min: 3 }) || ...,
```

We could also make these calls asynchronous (which is essentially what @erikdstock built), but I didn't do that to keep it simple. We always have the option to opt in, but it'd take a significant amount of time to remove it from things that wouldn't have had to depend on it. I think in this particular case we can maintain our own, small validation logic.

We have [a handful of tests](https://github.com/artsy/emission/blob/e9dd514b1c18ac1388b1070b47cd8de1a16718bd/src/lib/Components/Bidding/Screens/__tests__/BillingAddress-tests.tsx) that all pass on all of the three spikes we worked on. These tests don't depend on implementation details so we can easily swap out/in dependencies and our own implementations without changing tests.

yuki24:
closing in favor of https://github.com/artsy/emission/pull/1074

PR #1066, opened 2018-05-25
[Input] Fixes an edge case where input border color does not get updated

yuki24:
This commit changes the <Input> component to always update the border color when the error prop gets updated by the parent. Without this change, the border color would stay `red100` when the props transitions from `{ error: true }` to `{ error: false }`.

I'll post a screenshot later!


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

PR #1067, opened 2018-05-28
remove dangerfile.ts from TS include

koenpunt:
because exclude has precedence, there's no point in having it in include too

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -17,7 +17,7 @@
     "types": ["jest", "react", "react-native"],
     "noEmit": true
   },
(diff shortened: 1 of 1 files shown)


orta:
I'm pretty sure this dance was needed to make the dev tools work Danger, but I cloned this version and everything seemed to work fine

merge on green

koenpunt:
It looks like CI failed on missing credentials..

orta:
Yeah, that's fine 👍 - this repo uses same-repo branches

PR #1068, opened 2018-05-28
[WIP] Enable graphql-crunch

zephraph:
This enables graphql-crunch which results in smaller over-the-wire playloads. This change is dependent on https://github.com/artsy/metaphysics/pull/1061.

### DIFF
--- a/src/lib/metaphysics.ts
+++ b/src/lib/metaphysics.ts
@@ -1,11 +1,11 @@
-import { NativeModules } from "react-native"
 import { uncrunch } from "graphql-crunch"
+import { NativeModules } from "react-native"
 const Emission = NativeModules.Emission || {}
 
 import { metaphysicsURL } from "./relay/config"
 import { NetworkError } from "./utils/errors"
 
-const enableCrunch = Emission.crunch || "true"
+const enableCrunch = Emission.crunch != null ? Emission.crunch : "true"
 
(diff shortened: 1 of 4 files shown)


zephraph:
@alloy we'll need https://github.com/artsy/metaphysics/pull/1061 first, but after that we should be able to do some benchmarks.

src/lib/metaphysics.ts
```diff
@@ -1,9 +1,12 @@
+import { uncrunch } from "graphql-crunch"
 import { NativeModules } from "react-native"
 const Emission = NativeModules.Emission || {}
 
 import { metaphysicsURL } from "./relay/config"
 import { NetworkError } from "./utils/errors"
 
+const enableCrunch = Emission.crunch != null ? Emission.crunch : "true"
```
alloy: Is there a specific reason we need this to be an option at all?

alloy:
Exciting 👏  Looking forward to the numbers!

src/lib/metaphysics.ts
zephraph: Likely not. I noticed that some values below were being pulled from NativeModules so I added it as a potential configuration point. In hindsight that might be better served as an env variable. 

Would you like me to remove it?

alloy:
👍

PR #1071, opened 2018-05-29
[Input] Fixes an edge case where input border color does not get updated

yuki24:
This commit updates the `<Input>` component to always update the border color when the error prop gets updated by the parent component. Without this change, the border color would stay `red100` when the props transitions from `{ error: true }` to `{ error: false }`.

## Before

input-error-to-valid-border-color-before

## After

input-error-to-valid-border-color-after

merge on green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@
 * Adds a billing address screen - yuki24
 * Reflect style updates for the `<Checkbox>` component - yuki24
 * Load Real values into `BidResult` screen + reimplement `<MarkdownRenderer>` - sepans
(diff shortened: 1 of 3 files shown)


yuki24:
I'll look into the test failure and merge once it's solved.

orta:
Interesting, looks like the bot cannot merge into a protected  branch:

>  { HttpError: {"message":"You're not authorized to push to this branch. Visit https://help.github.com/articles/about-protected-branches/ for more information.","documentation_url":...]

orta:
Yep, known, and a long discussion here: https://platform.github.community/t/repositories-which-have-protected-branches-with-push-restrictions-have-no-ability-to-grant-push-rights-to-integrations/1376/47

PR #1076, opened 2018-05-31
Display "Credit card" and "Billing address" to users without credit cards

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-163

This PR wires up the `<ConfirmFirstTimeBid>` component so users without credit cards will be asked to register a credit card and billing address. I also added tests that don't depend on specific components, so if we were to merge `<ConfirmBid>` and `<ConfirmFirstTimeBid>` components into one these tests would act like a safety net.


mennenia:
Looks great 👍

Issue #1077, opened 2018-05-31
Deployments don't work


ashfurrow:
As documented, we _should_ be able to run `npm version patch` and get a new release made. We've had [other issues](https://github.com/artsy/emission/issues/971) around this in the past, but this issue is specifically about improving the deploy process and updating the documentation in the readme. 

Questions:

- Is `make update_specs_repos` still required? Does it work as expected? What about `make ship_yoga`?
- It seems like we've got our foot half in the door of automated npm magic, and half in the door of a home-rolled process, maybe we should go all-in on doing things ourselves? When something fails, it's not clear _where_ it failed and where to pick up from.

These are the steps I currently take to deploy Emission (not including any updates to React Native, Yoga, etc):

- Follow the instructions in the readme up the `npm version` command.
- `npm version patch` will fail when running this command in the `postversion` script: `env EMISSION_ROOT=\"$(pwd)\" pod repo push artsy --allow-warnings --use-json --skip-import-validation`. 
- To get around that failure, we need to push the podspec using a modified version of the [`release` script](https://github.com/artsy/emission/blob/master/scripts/release.js). Comment out all the lines at the bottom except for `publishPodspec("Emission.podspec")` and run the script with `node scripts/release.js`.
- In a local metaphysics folder adjacent to Emission, pull the latest from master and then run `node scripts/merge-graphql-query-map.js`. Then send a PR to Metaphysics with those changes.

alloy:
What we could do is move the podspec publishing to a script so we can make temporary adjustments there and the release tasks in `package.json` can remain stable.

orta:
linking these two issues, https://github.com/artsy/emission/issues/971

ashfurrow:
Right! I meant to lint to that, thanks Orta.

orta:
for anyone reading along, this error is because we don't have yoga (lowercase y) in our specs repo (and at version 0.54.4 ) https://github.com/artsy/Specs/tree/master/Yoga

orta:
As  long as it gets pushed to the tag correctly, which it look like it does then it should be fine

ashfurrow:
I'd be fine with that 👍

PR #1078, opened 2018-05-31
Adds note about deploys not working

ashfurrow:
I thought linking to an issue with the current workaround would be less brittle than documenting everything in the readme, so I opened #1077.


alloy:
👍

PR #1080, opened 2018-05-31
Use a plain variable when triggerring render() is not required #trivial

yuki24:
We can always use a plain variable when you do not have to trigger `render()` so the polling logic does not depend on the React lifecycle.

#skip_new_tests


ashfurrow:
Interesting, that makes sense. I'm curious: is there a risk of React re-rendering the component and loosing this state? Is this change something like a code style change, or a performance improvement? Just curious.

orta:
> is there a risk of React re-rendering the component and loosing this state?

That same issue would happened when it was in state too, so probably not an issue

PR #1083, opened 2018-06-04
Reafctor the `<MarkdownRenderer>` component #trivial

yuki24:
* Bring the simple-markdown dependency back in to replace the regex-based match
  * Use our styled-system-based components to make it more flexible
  * Update the tests to properly test the component
  * Add a storybook for easier visual testing

## New storybook

screen shot 2018-06-04 at 15 58 10


src/lib/Components/Bidding/Components/__tests__/Markdown-tests.tsx
```diff
@@ -0,0 +1,72 @@
...
+      </Markdown>
+    </BiddingThemeProvider>
+  )
+
+  expect(markdown.root.findAllByType(LinkText).length).toEqual(1)
+  expect(markdown.root.findAllByType(LinkText)[0].props.children[0]).toMatch("join the live auction")
+
+  markdown.root.findAllByType(LinkText)[0].props.onPress()
+
+  expect(SwitchBoardMock.presentModalViewController).toHaveBeenCalledWith(any(Object), "http://www.artsy.net")
```
yuki24: @ashfurrow Sorry I squashed my change by accident, but this line is basically what I added. I wasn't sure if sending an entire URL is enough to make the `SwitchBoard` work, but let me know if I'm missing something.

ashfurrow: Yeah this looks great! I think the convention in Emission is to use `expect.anything()` instead of `any(Object)` but I don't have a strong opinion about it.

sepans:
Looks great! Thanks for refactoring this. 🎉 🎉  I am wondering if you have tried it with the [actual messages](https://github.com/artsy/metaphysics/blob/master/src/schema/me/bidder_position_messages.js) coming from metaphysics. I am not sure the line break that MP returns ( space+space+line_break) works with 'simple-markdown' (if not we can update MP).

I also had difficulty with the alignment of [this](https://github.com/artsy/metaphysics/blob/master/src/schema/me/bidder_position_messages.js#L43) message where there is text, link and text again. Super curious to know if you had that problem and if yes how you solved it.

yuki24:
At least in the storybook it seems to be rendering the text + link + text sequence:

emission-markdown

I just assumed the message from MP contains a `\n` line break, but yes we should test it and update MP accordingly.

src/lib/Components/Bidding/Components/__tests__/Markdown-tests.tsx
```diff
@@ -0,0 +1,69 @@
...
+      </Markdown>
+    </BiddingThemeProvider>
+  )
+
+  expect(markdown.root.findAllByType(LinkText).length).toEqual(1)
+  expect(markdown.root.findAllByType(LinkText)[0].props.children[0]).toMatch("join the live auction")
+
+  markdown.root.findAllByType(LinkText)[0].props.onPress()
+
+  expect(SwitchBoardMock.presentModalViewController).toHaveBeenCalledWith(anything(), "http://www.artsy.net")
```
yuki24: Ok, I replaced the `any(Object)` with `anything()`. I think this is good to go?

ashfurrow: Yup! I'll leave merging up to Sepand.

PR #1085, opened 2018-06-06
Register flow infrastructure

ashfurrow:
This work is tracked in [PURCHASE-174](https://artsyproduct.atlassian.net/browse/PURCHASE-174).

This PR was originally going to add a new native view controller, but I decided that reusing the existing Objective-C class was a better choice, but does lead to a little more complexity in the query renderer. Here's the breakdown:

- Expand the existing `<BidFlow>` component to support a registration flow based on a new `intent` prop.
- That makes the `artworkID` a nullable prop, so the query renderer `<BidFlow>` needs to handle missing artwork IDs too. I made a `TODO`.
- There's a possible edge case where the intent is `"bid"` but the `artworkID` is null, which would indicate programmer error, but something I wanted to bring up.

I [verified with Brian](https://artsy.slack.com/archives/C9YNS4X32/p1528309649000818) that both the bidding and registration flows take place in modals. There's a question of what will happen when we move to a partially-visible component (one that doesn't hide the full screen behind it) when over the Auctions view controller, which is still in Swift, but we can deal with that when it gets here.

Marking as #native_no_changes since it's all backwards compatible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 * Adds going back to max bid when selecting edit bid from confirmation screen - maxim
 * Adds line-height to the type system - sepans
+* Adds `intent` to `ARBidFlowViewController` and `<BidFlow>` component - ash
 
 ### 1.5.3
 
--- a/Pod/Classes/ViewControllers/ARBidFlowViewController.h
+++ b/Pod/Classes/ViewControllers/ARBidFlowViewController.h
@@ -9,8 +9,8 @@ typedef enum : NSUInteger {
 
 @interface ARBidFlowViewController : ARComponentViewController
 
-- (instancetype)initWithArtworkID:(NSString *)artworkID saleID:(NSString *)saleID; /// Defaults to bid.
-- (instancetype)initWithArtworkID:(NSString *)artworkID saleID:(NSString *)saleID intent:(ARBidFlowViewControllerIntent)intent;
+- (instancetype)initWithArtworkID:(nullable NSString *)artworkID saleID:(NSString *)saleID; /// Defaults to bid.
+- (instancetype)initWithArtworkID:(nullable NSString *)artworkID saleID:(NSString *)saleID intent:(ARBidFlowViewControllerIntent)intent;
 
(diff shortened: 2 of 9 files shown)


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  This PR includes changes to the Emission Pod's native code but does not have a `package.json`
     change for the update to the `"native-code-version"`. If this is fine, add #native_no_changes to your PR message.
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

yuki24:
What's the reason why you didn't go with the separate view controller approach? I feel the registration flow is a different flow than the bidding flow and it would require us to add more logic to the `<ConfirmBid>` component that's already complex.

ashfurrow:
@yuki24 That's a good question. I started by adding a second view controller but realized that we're going to have to have a lot of code in Eigen to determine if the user should see the bidding vs registration views. I thought that by using a single component, we could isolate that complexity inside Emission where more developers are comfortable working.

I totally see what you're saying about the complexity of the confirmation component, and I want to be specific that I don't think we should use that component for both bidding and registration. My idea is to use a different `initialRoute` [in the BidFlow component](https://github.com/artsy/emission/blob/b85dc06e09f2b2a26d8a9817064568b02433d846/src/lib/Containers/BidFlow.tsx#L20-L24) so that it presents the user with either a registration UI or a bidding UI (the bidding UI might also include registration too, it's all kind of complicated). But yeah, my motivation here is really to keep as much logic in JavaScript as possible.

ashfurrow:
I guess what we need to balance is between:

- Adding complexity to Objective-C code to use one view controller or the other, preload one Relay query or another, etc.
- Adding complexity to JavaScript code to direct the user down one path or another, combining both paths' Relay fragments, etc.

ashfurrow:
> It sounds like we should think of the `<BidFlow>` component as a router

Yes exactly. Agreed about renaming, too.

ashfurrow:
@erikdstock Can you review+merge when ready?

PR #1087, opened 2018-06-07
Move dependency-free functions and constants out of ConfirmBid #trivial

yuki24:
As part of the effort to consolidate the `<ConfirmBid>` and `<ConfirmFirstTimeBid>` components, I started refactoring the `<ConfirmBid>` component. This PR extracts functions and constants that do not depend on `props` or `state` and makes it easier to understand what's specific to `<ConfirmBid>` and what is not.

Let me know if you have any feedback!

### DIFF
--- a/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
+++ b/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
@@ -57,6 +57,40 @@ const bidderPositionMutation = graphql`
     }
   }
 `
+
+const queryForBidPosition = (bidderPositionID: string) => {
+  return metaphysics({
+    query: `
+      {
+        me {
+          bidder_position(id: "${bidderPositionID}") {
+            status
+            message_header
+            message_description_md
+            position {
+              id
+              processed_at
+              is_active
+              suggested_next_bid {
+                cents
+                display
+              }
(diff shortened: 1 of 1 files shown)


src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -57,6 +57,37 @@ const bidderPositionMutation = graphql`
...
+            }
+          }
+        }
+      }
+    `,
+  })
+}
+
+const messageForPollingTimeout = `
+  We're receiving a high volume of traffic and your bid is still processing.\n
```
sweir27: I'm sure that @briansw will comment about this, but as we found in our mocktion the `'` here should be a curly "smart quote". I think the one on line 88 is correct.

yuki24: Ok I just made the quote smarter. I also added line breaks to match the styles in zeplin: zpl://screen?sid=5aecb372f73dd3091d585cad&pid=5ab94282c124db8d58365b47 (now that the `<markdown>` component interprets `\n` properly so we no longer have to add `\n`s or ` . \`)

sepans:
Looks great. merging

PR #1088, opened 2018-06-07
Refetch bid increments

ashfurrow:
This is initial work that @sepans and I did for [PURCHASE-168](https://artsyproduct.atlassian.net/browse/PURCHASE-168). It needs tests still, and will conflict a bit with @yuki24's work on consolidating the two confirm bid components, but basically we just pass along a callback to get executed whenever a bid has been processed (and, therefore, the increments need to get updated). We pass this function as a prop to the route pushed onto the navigator. Sepand and I will wrap this up tomorrow.


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  No CHANGELOG added.
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>




<p align="right">
</p>

ashfurrow:
Okay, wrapped this up. I added some tests for the refetching and cleaned up a few tests in `BidFlow-tests.tsx` that were using the Relay component directly instead of the component that it wrapped. This caused a problem trying to inject in our own faked Relay environment, which now works.

ashfurrow:
CI is failing but it's due to a hang. Travis is having a current service outage, looks unrelated so I've restarted the builds. We'll see!

ashfurrow:
Adding `--forceExit` to the Jest invocation should fix things, but I'd prefer to know _why_ this PR is making Jest hang. I'll take a look.

ashfurrow:
Re-marked the PR as a WIP to narrow down which tests are causing the hangs.

src/lib/Components/Bidding/Screens/__tests__/SelectMaxBid-tests.tsx
```diff
@@ -33,9 +48,63 @@ const SaleArtwork = {
...
+  component.root.findByType(Button).instance.props.onPress()
+  const nextScreen = fakeNavigator.nextStep()
+  fakeRelay.refetch.mockImplementationOnce((_params, _renderVars, callback) => {
+    callback()
+  })
+
+  nextScreen.root.instance.props.bidWasProcessed()
+
+  expect(component.root.findAllByType(Spinner).length).toEqual(0)
+})
```
damassi: v nice 👍

src/lib/Components/Bidding/Screens/__tests__/SelectMaxBid-tests.tsx
```diff
@@ -33,9 +48,63 @@ const SaleArtwork = {
...
     .toJSON()
   expect(component).toMatchSnapshot()
 })
+
+it("shows a spinner while fetching new bid increments", () => {
+  const component = renderer.create(
+    <SelectMaxBid me={Me} sale_artwork={SaleArtwork} navigator={fakeNavigator as any} relay={fakeRelay as any} />
+  )
+
+  component.root.instance.setState({ isRefreshingSaleArtwork: true })
```
sepans: I was thinking about mocking the `refresh` so it accepts the same signature, including the callback, and have the state set by the callback that is defined in the component instead of explicitly setting it here. But with the implementation I can think of the mock is going to be too complicated and the test mostly verifies the mock rather than the component so an overkill, unless there is a smarter way to do that 😄

sepans: Just noticed you have such e mock for the last test so all good 😄

PR #1090, opened 2018-06-07
Re-organize storybooks for the new bidding flow #trivial

yuki24:
The `BidFlow.story.tsx` has grown since we added it in the first sprint and I think it's time to extract some of the storybooks for smaller components.

Let me know if you have any feedback!


orta:
Yeah, this makes sense to me 👍

PR #1091, opened 2018-06-08
Refactor <ConfirmBid> #trivial

yuki24:
This PR simplifies most of the variable assignments in the `<ConfirmBid>` component using ES6's variable deconstruction.

Let me know if there's any feedback!

### DIFF
--- a/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
+++ b/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
@@ -135,13 +135,12 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConformBidState
   verifyBidPosition(results, errors) {
     // TODO: Need to handle if the results object is empty, for example if errors occurred and no request was made
     // TODO: add analytics for errors
-    const status = results.createBidderPosition.result.status
+    const { status, position, message_header, message_description_md } = results.createBidderPosition.result
+
     if (!errors && status === "SUCCESS") {
-      this.bidPlacedSuccessfully(results.createBidderPosition.result.position.id)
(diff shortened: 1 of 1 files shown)


src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -194,8 +179,7 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConformBidState
         status,
         message_header: messageHeader,
         message_description_md: messageDescriptionMd,
-        winning,
-        bid: this.props.bid,
+        winning: status === "WINNING",
```
yuki24: Ideally, we should pass the entire `BidderPositionResult` to the `<BidResult>` component and let it display an appropriate message, but this should be good for now.

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -135,13 +135,12 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConformBidState
   verifyBidPosition(results, errors) {
     // TODO: Need to handle if the results object is empty, for example if errors occurred and no request was made
     // TODO: add analytics for errors
-    const status = results.createBidderPosition.result.status
+    const { status, position, message_header, message_description_md } = results.createBidderPosition.result
```
ashfurrow: As long as we're refactoring this, the comment on line 136 about the `results` object being `null` could be looked at. No pressure though.

yuki24: I just created a new story to address the error case: https://artsyproduct.atlassian.net/browse/PURCHASE-187

PR #1092, opened 2018-06-12
Move the logic for showing bidding result messages to <BidResult> #trivial

yuki24:
This PR moves some logic for showing bidding result messages to the `<BidResult>` component. This is basically done but marking as WIP as the CI build is broken. I also added a few more storybooks for easier debugging.

 * [x] Fix the CI build

## Screenshots

screen shot 2018-06-12 at 7 18 57 pmscreen shot 2018-06-12 at 7 18 54 pm
screen shot 2018-06-12 at 7 18 51 pmscreen shot 2018-06-12 at 7 18 44 pm


src/lib/Components/Bidding/Screens/BidResult.tsx
```diff
@@ -42,54 +70,51 @@ export class BidResult extends React.Component<BidResultProps> {
...
+
+    return (
+      <BiddingThemeProvider>
+        <Container mt={6}>
+          <View>
+            <Flex alignItems="center">
+              <Icon20 source={Icons[status] || require("../../../../../images/circle-x-red.png")} />
+
+              <Title m={4}>
+                {status === "PENDING" ? "Bid Processing" : message_header || "You're the highest bidder"}
```
sweir27: :D we could knock out a QA ticket by renaming this to `Bid processing` (since they should all be sentence case). Also it looks like the apostrophe in `You're` should be a smart quote.

sweir27:
👏 👏 this is awesome, @yuki24 !!

src/lib/Components/Bidding/Screens/BidResult.tsx
yuki24: fixed :+1:

src/lib/Components/Bidding/Screens/BidResult.tsx
```diff
@@ -18,21 +18,49 @@ import { BidResult_sale_artwork } from "__generated__/BidResult_sale_artwork.gra
...
 }
 
 interface BidResultProps {
   sale_artwork: BidResult_sale_artwork
-  winning: boolean
-  status: string
-  message_header?: string
-  message_description_md?: string
-  suggested_next_bid?: Bid
+  bidderPositionResult: BidderPositionResult
```
erikdstock: #minor but if causality ever takes over bidding the name 'bidder position' will be inaccurate. In a slightly different situation in pulse we just call this a bid result.

yuki24: `BidderPositionResult` is a name that exists in MP, so we should change MP and then Emission if that's the case.

ashfurrow: Yeah I think we'd probably access Causality's API through a stitched version in Metaphysics anyway, so the remapping of `BidderPositionResult` could happen there instead of within clients – GraphQL is so cool! 😄

PR #1093, opened 2018-06-13
Decreases timer tests duration #trivial

ashfurrow:
I noticed that the five tests in `Timer-tests.tsx` were consistently taking longer than ten seconds to complete. I took a look and it appears that calls to [`advanceTimersByTime()`](https://github.com/facebook/jest/blob/3ce87434570d23be061648162488ca5599e40651/packages/jest-util/src/fake_timers.js#L260) does an amount of work that scales linearly with the amount of time to advance by (at least in our case, using `setInverval()`). When calling with `20 * MINUTES`, we're really calling `advanceTimersByTime(1200000)` 😅 

This PR reduces the amount of time the test takes to run to about 1s by reducing the number of minutes to advance time by to `1`, along with some adjustments to test data (to make sure time advances still cascade from seconds to minutes to hours). Let me know what I can clarify.

### DIFF
--- a/src/lib/Components/Bidding/Components/__tests__/Timer-tests.tsx
+++ b/src/lib/Components/Bidding/Components/__tests__/Timer-tests.tsx
@@ -59,17 +59,17 @@ it("shows 'Live' when the liveStartsAt prop is given", () => {
 })
 
 it("counts down to zero", () => {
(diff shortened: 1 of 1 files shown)


yuki24:
Wow thank you! I've been wondering why this was taking longer but didn't have time to look into it. Thanks again for fixing it!

Issue #905, opened 2017-12-19
[RFC] Add husky and lint-staged


orta:
Brings this repo inline with Reaction, Metaphysics et al. 

Lowers the barrier to entry because you don't have to keep telling newcomers to the codebase to run two arbitrary commands from our makefile to have it work like the other modern JS repos:

https://github.com/artsy/emission/blob/68debb7e08e1f9cffc3821ee2e55cb62f820b79c/Makefile#L34-L36

orta:
Ideally issues like this wouldn't crop up: https://github.com/artsy/emission/pull/903#issuecomment-353256913

orta:
These are both in now

PR #1102, opened 2018-06-17
Fresh install feedback

niklassaers:
Hi,
when running "make setup && make oss" it installs node 10.4.1 and npm 6.1.0, but package.json specifies 8.4.x and 5.4.x, and thus the install breaks.

I updated package.json to point to 10.4.x and 6.1.x, and that seems fine. Added to PR.

Yarn is installed in /usr/local/bin, the pre-commit and pre-push hooks assumes this is in path. When using Tower it was not. Adding "export PATH=$PATH:/usr/local/bin" into the hooks remedied this. Added to PR.

Another thing, compared to the manual install, these automatic steps didn't build Cocoapods. Not sure they should, but thought I'd mention it.

Not sure as to why I had to enter my GitHub credentials during the "bundle exec pod install" step, was that intended?

Cheers

   Niklas

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,23 @@
 BREW := $(shell command -v brew 2> /dev/null)
+NVM := $(shell command -v nvm 2> /dev/null)
 YARN := $(shell command -v yarn 2> /dev/null)
 WATCHMAN := $(shell command -v watchman 2> /dev/null)
 CODE := $(shell command -v code 2> /dev/null)
 
 setup:
 ifndef BREW
-    $(error "Please install homebrew before running `make oss`: https://brew.sh")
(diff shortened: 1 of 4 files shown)


orta:
Hrm, we should finally move Emission to use the same git hooks as the rest of our JS infra 

and I don't think we should update our node version either (that's a non-trivial bit of work), so I think we need to edit the install script to use nvm + 8.x

orta:
I've shipped commit that fixes a bunch of this

orta:
I meant to this PR 👍

orta:
Fixes https://github.com/artsy/emission/issues/905

alloy:
@niklassaers Thanks so much for taking the time! 👏

niklassaers:
Thank you 👍

PR #1104, opened 2018-06-18
Consoldiate types #trivial

yuki24:
We've been adding many many types since we started working on the bidding flow. Now it's time to consolidate and organize types all in `src/lib/Components/Bidding/types.d.ts`.

Let me know if there's any feedback!

#skip_new_tests

### DIFF
--- a/src/lib/Components/Bidding/types.d.ts
+++ b/src/lib/Components/Bidding/types.d.ts
@@ -1,0 +1,77 @@
+export interface Address {
+  fullName: string
+  addressLine1: string
+  addressLine2?: string
+  city: string
+  state: string
+  postalCode: string
+}
+
+export interface Bid {
+  display: string
+  cents: number
+}
+
+export interface BidderPositionResult {
+  /*
+   * Represents the status of a bidder position. The logic that determines the status is as follows:
+   *
+   *   bidder position created?
+   *     - yes
+   *       - bidder position processed?
+   *         - yes
+   *           --> OUTBID | RESERVE_NOT_MET | WINNING
+   *         - no
+   *           --> PENDING
+   *     - no
+   *       - do we know why?
+   *         - yes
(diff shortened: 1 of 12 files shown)


src/lib/Components/Bidding/types.d.ts
```diff
@@ -0,0 +1,60 @@
+export interface BidderPositionResult {
```
sweir27: #minor but can we alphabetize these?

Also I imagine that `Address` will eventually want to be extracted further, but for now 👍

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Missing Test Files:

- `src/lib/Components/Bidding/__tests__/types.d-tests.ts`

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/lib/Components/Bidding/types.d.ts
yuki24: Ideally we shouldn't have to define and maintain the types for data coming from MP (we have[ `BidderPositionResult` for this call](https://github.com/artsy/emission/blob/6cf7031c7a880e13ec1b04f64c1cc8a84f4589e6/src/lib/Components/Bidding/Screens/ConfirmBid.tsx#L71-L82)). I'd prefer to organize `PaymentCardTextFieldParams` and `StripeToken` next to each other since these two are both Stripe-related, but for now that's :+1

yuki24:
seems like `#skip_new_tests` is being ignored? Other than that, this should be good.

ashfurrow:
@yuki24 nah, it's working as designed 🙃 

https://github.com/artsy/emission/blob/40cc2739ee896a863eeb31db9e7455031eded1a8/dangerfile.ts#L73

It turns an error into a warning is all.

PR #1108, opened 2018-06-20
#trivial Updates to bid flow: require billing address/cc to be present before placing bid; move tokenization to cc form

sweir27:
This PR makes a few small changes (and was born out of my initial exploration into getting us fancy errors after stripe tokenization). I'll document my findings here:

- It _seems_ like most of the errors that would cause tokenization to fail are captured in the built-in form validations on the credit card form (things like an invalid cc number, cvc, or expiration date are already handled for us). 
- I moved the credit card tokenization on the `CreditCardForm` to that component so we get feedback on that screen if something goes wrong, but a follow-up will be to make sure we're catching networking errors and showing something generic on that screen. I'll update the story to be more explicit about this.
- I updated the `ConfirmFirstTimeBid` screen to require `billingAddress`, `creditCardToken`, and `conditionsOfSaleChecked` to be present before allowing one to place a bid.
  - Note that the **button** display isn't yet updated to look "disabled". I know @mennenia is working on consolidating our button usage, so before adding custom states I figured I'd wait for that. I'll make sure to add a follow-up story to capture this requirement.

### DIFF
--- a/src/lib/Components/Bidding/Screens/ConfirmFirstTimeBid.tsx
+++ b/src/lib/Components/Bidding/Screens/ConfirmFirstTimeBid.tsx
@@ -96,8 +96,12 @@ export class ConfirmFirstTimeBid extends React.Component<ConfirmBidProps, Confir
     })
   }
 
-  onCreditCardAdded = async (params: PaymentCardTextFieldParams) => {
-    const token = await stripe.createTokenWithCard(params)
+  canPlaceBid() {
+    const { billingAddress, creditCardToken, conditionsOfSaleChecked } = this.state
+    return billingAddress && creditCardToken && conditionsOfSaleChecked
+  }
+
+  onCreditCardAdded = async (token: StripeToken, params: PaymentCardTextFieldParams) => {
     this.setState({ creditCardToken: token, creditCardFormParams: params })
   }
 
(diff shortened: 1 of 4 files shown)


src/lib/Components/Bidding/Screens/ConfirmFirstTimeBid.tsx
```diff
@@ -96,8 +96,12 @@ export class ConfirmFirstTimeBid extends React.Component<ConfirmBidProps, Confir
     })
   }
 
-  onCreditCardAdded = async (params: PaymentCardTextFieldParams) => {
-    const token = await stripe.createTokenWithCard(params)
+  canPlaceBid() {
+    const { billingAddress, creditCardToken, conditionsOfSaleChecked } = this.state
+    return billingAddress && creditCardToken && conditionsOfSaleChecked
```
sweir27: Since the conditional below returns either a function or `null` if the button can't be pressed, it was redundant to write a test for this case. If we're unable to call `onPress()` in tests, it's because the fields aren't present. Thus the test that correctly stubs the state and then calls `onPress()` actually captures this as well.

yuki24:
It seems like this PR was sent from your fork 😅

orta:
Context:

screen shot 2018-06-20 at 3 48 10 pm

sweir27:
Hehe @yuki24 it's allowed now! I jumped at the opportunity 😄

yuki24:
Sweet, merging

PR #1110, opened 2018-06-20
Update auction registration status on bid

erikdstock:
#trivial
This is a very blunt tool right now and might be missing some edge cases. It updates the auction screen's registration status (via NSNotificationCenter) upon receiving any bid result in the ConfirmBid screen.

I'm not sure if i'm missing anything & would like to add more tests- Also considered exporting constants for these Eigen notification names since we'll likely use them in another place on the registration flow, (and LAI bids in the future?) but holding back on that for now.

### DIFF
--- a/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
+++ b/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
@@ -164,6 +164,9 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
       ARAuctionID: this.props.sale_artwork.sale.id,
       ARAuctionArtworkID: this.props.sale_artwork.artwork.id,
     })
+    NativeModules.ARNotificationsManager.postNotificationName("ARAuctionArtworkRegistrationUpdatedNotification", {
+      ARAuctionID: this.props.sale_artwork.sale.id,
+    })
     this.props.navigator.push({
       component: BidResultScreen,
       title: "",
--- a/src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
+++ b/src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
@@ -19,9 +19,9 @@ import Spinner from "../../../Spinner"
 import objectContaining = jasmine.objectContaining
 
 let nextStep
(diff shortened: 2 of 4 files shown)


src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -164,6 +164,9 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
       ARAuctionID: this.props.sale_artwork.sale.id,
       ARAuctionArtworkID: this.props.sale_artwork.artwork.id,
     })
+    NativeModules.ARNotificationsManager.postNotificationName("ARAuctionArtworkRegistrationUpdatedNotification", {
```
ashfurrow: As [discussed in Slack](https://artsy.slack.com/archives/C9YNS4X32/p1529534078000289?thread_ts=1529527038.000301&cid=C9YNS4X32), this isn't _always_ required (if the user is already registered and is only bidding) but I don't think it'll be a problem – something to watch out for in mocktions.

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -206,6 +208,19 @@ describe("polling to verify bid position", () => {
...
+      component.root.instance.setState({ conditionsOfSaleChecked: true })
+      mockphysics.mockReturnValueOnce(Promise.resolve(mockRequestResponses.pollingForBid.reserveNotMet))
+      relay.commitMutation = jest.fn((_, { onCompleted }) => {
+        onCompleted(mockRequestResponses.placeingBid.bidAccepted)
+      })
+
+      component.root.findByType(Button).instance.props.onPress()
+      jest.runAllTicks()
+
+      expect(mockPostNotificationName).toHaveBeenCalledTimes(2)
```
ashfurrow: Nice – I think either adding a comment explaining the `2`, or having two `expect()`'s for two sets of parameters, could help clarify to this. Really nice test.

erikdstock: Right- I first wanted to check that I could assert anything at all.

erikdstock:
I think this is done + good to merge.

PR #1111, opened 2018-06-21
Merge <ConfirmBid> and <ConfirmFirstTimeBid> #trivial

yuki24:
This is an attempt to consolidate the logic in the `<ConfirmBid>` and `<ConfirmFirstTimeBid>` components. Let me know what you think!

#skip_new_tests


src/lib/Components/Bidding/Screens/ConfirmFirstTimeBid.tsx
```diff
@@ -1,312 +0,0 @@
...
-      title: "",
-      passProps: {
-        onSubmit: this.onBillingAddressAdded,
-        billingAddress: this.state.billingAddress,
-        navigator: this.props.navigator,
-      },
-    })
-  }
-
-  canPlaceBid() {
```
sweir27: #minor looks like this didn't make it up to the consolidated component, but the logic would change slightly anyways  so I'll make this part of a follow-up PR.

sweir27:
Looks like this needs a rebase, but I think it's a 👍direction.

I noted one place where some logic I added in a previous PR got removed, but I think it's alright and I can follow up with something more comprehensive now that the components are merged.

src/lib/Components/Bidding/Screens/ConfirmFirstTimeBid.tsx
yuki24: I've added this back but feel free to update as you need.

yuki24:
This should be good for review.

PR #1112, opened 2018-06-21
Autocapitalizes words on billing address inputs

ashfurrow:
After pressing spacebar, users will have their shift keys enabled already for them to enter in their info. 

screen shot 2018-06-21 at 3 24 31 pm

Work tracked here: https://artsyproduct.atlassian.net/browse/PURCHASE-223

### DIFF
--- a/src/lib/Components/Bidding/Screens/BillingAddress.tsx
+++ b/src/lib/Components/Bidding/Screens/BillingAddress.tsx
@@ -94,17 +94,24 @@ export class BillingAddress extends React.Component<BillingAddressProps, Billing
               Your billing address
             </Title>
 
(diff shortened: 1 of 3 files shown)


src/lib/Components/Bidding/Screens/BillingAddress.tsx
```diff
@@ -94,17 +94,24 @@ export class BillingAddress extends React.Component<BillingAddressProps, Billing
               Your billing address
             </Title>
 
-            <StyledInput label="Full name" placeholder="Enter your full name" {...this.propsForInput("fullName")} />
+            <StyledInput
+              label="Full name"
+              placeholder="Enter your full name"
+              autoCapitalize="words"
```
sweir27: Ha! Didn't realize this was so simple. 😄

PR #1113, opened 2018-06-21
Ensures that JS is uploaded

orta:
This extra check on master isn't needed because you're on master and always have secret env vars


orta:
merge on green

PR #1114, opened 2018-06-22
Refactors timer to update on state change

sweir27:
This PR addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-210

Basically, when the state of an auction changes (or in our case, when the timer runs out), we want to update the label on the timer to reflect the new state.

In order to make this clearer, I extracted `AuctionTimerState` as its own enum and used that to define the `nextState` and `currentState`. The logic looks similar between these two methods but is slightly different so it made sense to keep them separate. But happy for ideas on how to make this make more sense!

timer

### DIFF
--- a/src/lib/Components/Bidding/Components/__tests__/Timer-tests.tsx
+++ b/src/lib/Components/Bidding/Components/__tests__/Timer-tests.tsx
@@ -24,11 +24,11 @@ beforeEach(() => {
   // Thursday, May 10, 2018 8:22:32.000 PM UTC
   Date.now = () => dateNow
   futureTime = moment(dateNow)
-    .add(1, "hour")
+    .add(1, "second")
     .toISOString()
 
   pastTime = moment(dateNow)
-    .subtract(1, "hour")
+    .subtract(1, "second")
     .toISOString()
 })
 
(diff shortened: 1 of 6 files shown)


src/lib/Components/Bidding/Components/__tests__/Timer-tests.tsx
```diff
@@ -94,3 +96,167 @@ it("shows month, date, and hour adjusted for the timezone where the user is", ()
...
+      expect(label).toEqual("In progress")
+      expect(relevantDate).toEqual(null)
+    })
+
+    it("shows the correct label for a sale that has closed", () => {
+      const { label, relevantDate } = timer.upcomingLabel(AuctionTimerState.CLOSED, pastTime, pastTime, pastTime)
+      expect(label).toEqual("Bidding closed")
+      expect(relevantDate).toEqual(null)
+    })
+  })
```
yuki24: These new tests are all testing internals and it would make it more difficult to refactor this component. It is also easier to understand if it tests the behaviour of a component. Would it be possible to write a test that doesn't depend on `state` or internal functions? I usually use `state` as the last resort and add a comment for why calling `state` is absolutely necessary.

sweir27: Since these methods have a lot of logic and depend on each other, breaking them out to test them individually in this way was super helpful while developing! But let me take a stab at trying to create tests that don't rely on the exact internals. 👍

src/lib/Components/Bidding/Components/__tests__/Timer-tests.tsx
```diff
@@ -94,3 +126,53 @@ it("shows month, date, and hour adjusted for the timezone where the user is", ()
 
   expect(getTimerLabel(timer)).toEqual("Ends May 14, 1pm")
 })
+
+describe("timer transitions", () => {
```
sweir27: @yuki24 Replaced the previous tests with these which test the outcome of advancing the timer as opposed to the individual methods. In general I like to have both, but in this case it ended up being a bit redundant. 😄 

Also.... have to look at these again before merging since these are now _super_ slow.

sweir27: Ah, looks like `advanceTimersByTime` wasn't just jumping to the future, but actually fast-forwarding through execution. I moved these to only deal with single ticks (i.e. one second) and they're now normal speed.

yuki24: This looks great! Thanks for updating the tests 😀

oxaudo:
How hard/easy do you think it would be to just drop this component into this design: https://app.zeplin.io/project/5b113f413ede4bb10dbfdc42/screen/5b114000413156335e285483 for auctions. Don't know of a good way to reuse the components from emission.

yuki24:
There are two things that need to be replaced: `<SansMedium12>` and `<SansMedium14>`. Everything else is just plain React/Javascript and can be shared very easily.

PR #1115, opened 2018-06-22
ConfirmBid screen handles whether or not you are registered #trivial

sweir27:
This PR addresses https://artsyproduct.atlassian.net/browse/PURCHASE-183 and https://artsyproduct.atlassian.net/browse/PURCHASE-218 (minor)

Before this, we weren't checking whether or not you were already registered for the sale before showing you the checkbox on the `ConfirmBid` screen. This fixes that and (hopefully) makes the logic a little clearer.

I also fixed that tiny alignment task with the edit/add buttons.


src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -383,6 +401,9 @@ export const ConfirmBidScreen = createFragmentContainer(ConfirmBid, {
   me: graphql`
     fragment ConfirmBid_me on Me {
       has_qualified_credit_cards
+      bidders(sale_id: $saleID) {
+        qualified_for_bidding
```
sweir27: We could just retrieve the `id` here instead of `qualified_for_bidding`-- it's a followup whether or not we care about your registration status once you get to this page (since theoretically you shouldn't even be able to _enter_ the bid flow if your registration is pending).

PR #1116, opened 2018-06-22
Extract payment info into a separate component #trivial

sweir27:
This PR extracts the payment info (which wraps the `BillingAddress` and `CreditCardForm` components), since it's re-used between the registration flow and the bid flow.

The parent of those components has to hold the truth about the credit card token and billing address, since we create the credit card + (bid or register) in one go on that big button. So there's still a little repeated logic but it's not too bad.

I also updated the credit card form to accept props and pre-fill the card number based on what you'd entered before, if you click the "Edit" button to go back.

Note that we don't handle _anywhere_ the ability to halfway-save your billing address or credit card info. Not sure if that's an issue, but will make sure to follow up in case that becomes a need.

Made this #trivial since it's just a refactor and doesn't add much new functionality.

#skip_new_tests

### DIFF
--- a/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
+++ b/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
@@ -235,6 +235,14 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
     SwitchBoard.presentModalViewController(this, "/conditions-of-sale?present_modally=true")
   }
 
+  onCreditCardAdded(token: StripeToken, params: PaymentCardTextFieldParams) {
+    this.setState({ creditCardToken: token, creditCardFormParams: params })
+  }
+
+  onBillingAddressAdded(values: Address) {
+    this.setState({ billingAddress: values })
+  }
+
   goBackToSelectMaxBid() {
     this.props.navigator.pop()
   }
@@ -265,7 +273,7 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
 
   render() {
     const { artwork, lot_label, sale } = this.props.sale_artwork
-    const { billingAddress, creditCardToken: token, requiresPaymentInformation, requiresCheckbox } = this.state
+    const { requiresPaymentInformation, requiresCheckbox } = this.state
 
     return (
       <BiddingThemeProvider>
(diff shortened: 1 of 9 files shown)


src/lib/Components/Bidding/Screens/__tests__/__snapshots__/ConfirmBid-tests.tsx.snap
```diff
@@ -238,6 +238,7 @@ exports[`renders properly 1`] = `
       accessibilityLabel={undefined}
       accessibilityTraits={undefined}
       accessible={true}
+      alignItems="center"
```
sweir27: Gah, this was supposed to make it into a previous PR but I think I `--amended` over it 😭. So technically _this_ PR would fix https://artsyproduct.atlassian.net/browse/PURCHASE-218, haha.

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -322,21 +298,14 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
...
-                <Divider mb={2} />
-              </View>
+              <PaymentInfo
+                navigator={this.props.navigator}
+                onCreditCardAdded={this.onCreditCardAdded.bind(this)}
+                onBillingAddressAdded={this.onBillingAddressAdded.bind(this)}
+                billingAddress={this.state.billingAddress}
+                creditCardFormParams={this.state.creditCardFormParams}
+                creditCardToken={this.state.creditCardToken}
+              />
```
yuki24: I'd rather make a `<CreditCardRow>` and `<BillingAddressRow>` components. Monolith components are a typical source of complexity and having a sinlge component to reduce 2 occurances in `<ConfirmBid>` and `<Registration>` seems less compeling:

```ts
<View>
  <CreditCardRow
    navigator={this.props.navigator}
    onCreditCardAdded={this.onCreditCardAdded.bind(this)}>

  <Divider mb={2} />

  <BillingAddressRow
    navigator={this.props.navigator}
    onCreditCardAdded={this.onCreditCardAdded.bind(this)}>

  <Divider mb={2} />
</View>
```

sweir27: Yeah, the main reason I thought this was okay was that we don't have any cases where you would have these components in isolation (i.e. you always have to have a billing address with a credit card, and vice versa). 

However, I don't feel strongly. Happy to extract.

yuki24:
The danger warning seems trivial. I'll go ahead and merge this PR.

PR #1118, opened 2018-06-25
Fix the commitMutation compatibility by not throwing errors

yuki24:
Currently, we throw [an error](https://github.com/artsy/emission/blob/50dc76ae47611da3a41e93782a8ad9a3f56ffb63/src/lib/relay/fetchQuery.ts#L44) when there is an error object present in any types of responses. This seems to have cause the `commitMutation` to invoke the `onError` callback when it should instead invoke the `onCompleted` callback. According the doc, it seems more appropriate to just resolve the promise even when an error is present:

> `onCompleted`: Callback function executed when the request is completed and the in-memory Relay store is updated with the `updater` function. Takes a `response` object, which is the "raw" server response, and `errors`, an array containing any errors from the server.


yuki24:
I wonder what kind of changelog entry should en added to the change log. If it is meat to be a list of user-facing changes, then this is an internal change and should not be added. But if it is intended to be a helpful list that developers look at, then this PR should also be part of that.

alloy:
Yeah the changelog should be helpful to devs, they need to know that this error handling changed.

Otherwise looks good, merge on green 👍

PR #1119, opened 2018-06-25
Refresh status of registration when you return to ConfirmBid screen + some tiny fixes

sweir27:
The outcome of a pairing session with @orta!

This PR addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-196

Previously, when you returned to the `ConfirmBid` screen after placing a bid, such as after you find out you'd been outbid (so after visiting `SelectMaxBidScreen` --> `ConfirmBidScreen` --> `BidResult` --> `ConfirmBidScreen`), you'd see the same options for registering/agreeing to the conditions of sale even though at this point you _must_ be registered.

This PR updates to explicitly refetch your bidder information right before navigating back to the `ConfirmBid` screen. I also added a similar refetch of the `SaleArtwork` information when revisiting the `SelectMaxBid` screen, as we found that the bid picker also became out of date.

(I just realized that we _could_ do a lot of this by passing props around... i.e. after you bid, you must be registered so we could pass a prop to not show the register info. I think that solution is worse/a little confusing as it means we're not getting the most up-to-date truth, but it's an option.)


Pod/Classes/GraphQL/ARGraphQLQueryCache.m
```diff
@@ -3,13 +3,14 @@
...
 #ifdef DEBUG
+// No caching in dev mode.
+const NSTimeInterval ARGraphQLQueryCacheDefaultTTL = 0;
 #define DLog(...) NSLog(__VA_ARGS__)
 #else
+const NSTimeInterval ARGraphQLQueryCacheDefaultTTL = 3600;
 #define DLog(...) /* */
 #endif
 
-const NSTimeInterval ARGraphQLQueryCacheDefaultTTL = 3600;
```
sweir27: @orta do you think I should leave this change in?

orta: nah, let's ditch it. Could have been a red herring.

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -343,14 +365,22 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
           </View>
 
           <View>
-            {requiresCheckbox && (
+            {requiresCheckbox ? (
```
sweir27: I know this is scope creep... but since I was finally able to test returning to this page I noticed that we weren't showing the "You agree to..." message when you were already registered. 😄

orta: 👍

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -335,36 +365,57 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
       </BiddingThemeProvider>
     )
   }
+
+  private determineDisplayRequirements(bidders: ReadonlyArray<any>, hasQualifiedCreditCards: boolean) {
+    const isRegistered = bidders && bidders.length > 0
+    const requiresCheckbox = !isRegistered
+    const requiresPaymentInformation = !(isRegistered || hasQualifiedCreditCards)
+    return { requiresCheckbox, requiresPaymentInformation }
+  }
```
orta: makes sense

orta:
Looks good - merge on green

PR #1122, opened 2018-06-26
Properly format error message for the PENDING state

yuki24:
I noticed that the white spaces in the `messageForPollingTimeout.desription` were interpreted as an actual space in the parsed markdown `Text`s. This PR removes them so the message will show up without a weird offset.


yuki24:
@sweir27 🙋‍♂️,  :shipit:❓

sweir27:
Looks great! Seems like it needs a little changelog love though 😄.

But merge on green 🍏 !

yuki24:
💥😫

PR #1123, opened 2018-06-26
Purchase-200 add register to bid flow routing

erikdstock:
https://artsyproduct.atlassian.net/browse/PURCHASE-200

This PR
- adds a `RegistrationFlow` path to our `AppRegistry`'s `BidFlow` (renamed to `BidderFlow` there, but not in the registered module naming- I'd like to settle on something consistent for these cases). This routing is handled in the `AppRegistry` itself.
- wires up the basic sale queries from `Registration_sale` to `RegistrationFlow_sale` to `QueryRenderersRegistrationFlowQuery`

image


Pod/Classes/ViewControllers/ARBidFlowViewController.m
```diff
@@ -15,7 +15,7 @@ - (instancetype)initWithArtworkID:(NSString *)artworkID saleID:(NSString *)saleI
                           @"intent": (intent == ARBidFlowViewControllerIntentBid ? @"bid" : @"register")
   };
   if ((self = [super initWithEmission:nil
-                           moduleName:@"BidFlow"
+                           moduleName:@"BidderFlow"
```
sweir27: Will this rename require a change in eigen or anything?

src/lib/AppRegistry.tsx
```diff
@@ -88,14 +90,33 @@ const Conversation: React.SFC<ConversationProps> = track<ConversationProps>(prop
...
-  intent: "bid" | "register"
+  intent: BidderFlowIntent
+}
+
+interface BidderFlow {
+  queryRenderer: React.ComponentType<BidderFlowRendererProps>
+  container: React.ReactType<any>
+}
+
+const BidderFlows: { [BidderFlowIntent: string]: BidderFlow } = {
```
sweir27: Is it possible to add tests for this? (even just to make sure that when the intent is specified a certain way, the expected component shows)?

Pod/Classes/ViewControllers/ARBidFlowViewController.m
erikdstock: At the moment I don't think so. This is internal to how the Eigen Pod connects to React Native components, and I picked the name change because `BidFlow` was a counterpart to `RegistrationFlow`, and they are both Bidder Flows. I think it would be good to change this file to `ARBidderFlowViewController.m` as well.

src/lib/AppRegistry.tsx
erikdstock: Checking in with @orta on this.

orta: Hrm, off-hand I'm not even too sure what this does, might be worth some comments TBH as it's code that's being smart to avoid some duplication from the looks of it. 

This would be very, very tricky to test, probably not worth the effort given that it would just not load at dev time rather that break in a subtle way

Pod/Classes/ViewControllers/ARBidFlowViewController.m
erikdstock: Now I'm thinking maybe this will change things - I'm assuming eigen will need to reinstall the pod. If so we can keep the existing name and make a small followup ticket to update them together.

src/lib/AppRegistry.tsx
```diff
@@ -109,4 +130,5 @@ AppRegistry.registerComponent("Inbox", () => Inbox)
 AppRegistry.registerComponent("Conversation", () => Conversation)
 AppRegistry.registerComponent("Inquiry", () => Inquiry)
 AppRegistry.registerComponent("Favorites", () => FavoritesScene)
-AppRegistry.registerComponent("BidFlow", () => BidFlow)
+// TODO: Change everything to BidderFlow? AuctionAction?
+AppRegistry.registerComponent("BidderFlow", () => BidderFlow)
```
erikdstock: I think I should probably leave this named BidFlow and update it at the same time as `ARBidFlowViewController`'s name and use of this component.

PR #1124, opened 2018-06-26
Switches from a git submodule to using the schema grabbed by the API 

orta:
To handle the stored schemas for relay.

Will highlight interesting things.

### DIFF
--- a/.graphqlconfig
+++ b/.graphqlconfig
@@ -1,0 +1,22 @@
+{
+  "projects": {
+    "metaphysics-sdl": {
+      "schemaPath": "data/schema.graphql",
+      "extensions": {
+        "endpoints": {
+          "staging": "https://metaphysics-staging.artsy.net/",
+          "local": "http://localhost:5000"
+        }
+      }
+    },
+    "metaphysics-json": {
+      "schemaPath": "data/schema.json",
+      "extensions": {
+        "endpoints": {
+          "staging": "https://metaphysics-staging.artsy.net/",
+          "local": "http://localhost:5000"
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


.graphqlconfig
```diff
@@ -0,0 +1,22 @@
...
+      "schemaPath": "data/schema.json",
+      "extensions": {
+        "endpoints": {
+          "staging": "https://metaphysics-staging.artsy.net/",
+          "local": "http://localhost:5000"
+        }
+      }
+    }
+  }
+}
```
orta: yet another GraphQL config file. Given that our graphql extension has been broken for many months it might make sense to remove the other graphql config file.

data/schema.json
```diff
@@ -1465,6 +1465,33 @@
               "isDeprecated": false,
               "deprecationReason": null
             },
```
orta: looks like this is in a different sort order

data/schema.graphql
```diff
@@ -1,11 +1,14 @@
+# source: https://metaphysics-staging.artsy.net/
+# timestamp: Tue Jun 26 2018 15:21:08 GMT-0400 (EDT)
+
 input AddAssetToConsignmentSubmissionInput {
-  # The type of the asset
+  """The type of the asset"""
```
orta: `"""` is so ugly. :-/

sweir27:
🙏 thank you @orta !!

data/schema.graphql
sweir27: Did `yarn relay` work for you? (I guess it would if you were able to push?) 

I tried running this locally and am getting syntax errors because of this change. Looks like due to: https://github.com/graphql-cli/graphql-cli/issues/199 and sounds like we have to update relay to support this new comment-style.

orta: That's a blocker for this PR today, we can't do that without serious work yet I'm afraid

orta:
in digging around, I found that we can set a flag for old style comments - so this can be re-updated

orta:
fixed by #1190

PR #1125, opened 2018-06-26
User should see an error message when a request to MP did not go through

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-188

This is a simple change for displaying an error message when an network error occurs. I'm not sure what the app should take the user to when the CONTINUE button is pressed, but that could be addressed as a separate task.

I tested this by disabling WiFi and hitting the PLACE BID button. The actual device may behave differently depending on the network condition.

## Screenshot:

network-error


src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -267,6 +272,21 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
     this.props.navigator.pop()
   }
 
+  presentErrorResult(error) {
+    console.error(error)
```
yuki24: I left this `console.error` to send the error to Sentry. In development this will show a red screen that may be surprising, but that's intentional.

src/lib/Components/Bidding/Screens/BidResult.tsx
```diff
@@ -29,13 +29,11 @@ interface BidResultProps {
...
   title: "Bid processing",
-  description: `
-    We’re receiving a high volume of traffic
-    and your bid is still processing.
-
-    If you don’t receive an update soon,
-    please contact [support@artsy.net](mailto:support@artsy.net).
-  `,
+  description:
+    "We’re receiving a high volume of traffic\n" +
```
sweir27: There's no way to just have a string of text that we wrap due to the size of the screen, right? 😬

yuki24: It's easier to use the `\n` to add line breaks rather than relying on the screen size. Though It is still possible that on iPhone SE this ends up with an unexpected line break.

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -172,7 +177,7 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
 
     commitMutation(this.props.relay.environment, {
       onCompleted: () => this.createBidderPosition(),
-      onError: errors => console.error("An error occurred in creditCardMutation: ", errors),
+      onError: this.presentErrorResult.bind(this),
```
sweir27: This error could also occur (imagining that we're only catching _certain_ errors from gravity in MP) for a variety of reasons, right? It seems a little fishy that we're always showing the "network" error. I realize these are the designs just noting since it could be confusing.

yuki24: Now that we have https://github.com/artsy/emission/pull/1118 in, the `onError` callback gets invoked only when the error type is an IO error. When Gravity returns an error to MP it will eventually be passed into the `onCompleted` callback as part of the `errors` argument. This is not handled right now, and is a case that will be covered by https://artsyproduct.atlassian.net/browse/PURCHASE-187.

src/lib/Components/Bidding/Screens/BidResult.tsx
yuki24: Actually I didn't mean to include this change as part of this PR. I'll update to remove the commit.

PR #1129, opened 2018-06-27
Adds mutations to registration flow

sweir27:
This PR adds the necessary mutations for registering for a sale.

The logic goes:
- If you have a credit card on file (doesn't have to be `qualified_for_bidding`) then we just call the `createBidder` mutation
- If you don't have a credit card on file then we first create a credit card, then call the `createBidder` mutation.

It's super similar to the registration flow and I was able to reuse a lot of the same logic. I _could_ extract some of this into a wrapper component, but it feels like that would get a bit confusing since bidding and registration share pieces but are fundamentally different flows.

Marking this as WIP until https://github.com/artsy/emission/pull/1123 is merged and I can rebase, since I'd like to test out the full flow from the initial render. But am interested in comments! I don't plan to add more functionality after the rebase.


yuki24:
I think this is mergeable, but would you prefer to wait until #1123 is merged and test on your side?

sweir27:
Yeah, actually we can probably merge this one since there will have to be some follow ups to deal with the mutation result anyways.

src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -74,6 +109,51 @@ export class Registration extends React.Component<RegistrationProps, Registratio
...
+    this.state.requiresPaymentInformation ? this.createCreditCardAndBidder() : this.createBidder()
+  }
+
+  async createCreditCardAndBidder() {
+    const { billingAddress, creditCardFormParams } = this.state
+    const token = await stripe.createTokenWithCard({
+      ...creditCardFormParams,
+      name: billingAddress.fullName,
+      addressLine1: billingAddress.addressLine1,
+      addressLine2: null,
```
yuki24: Are we not passing in `address2` at all?

yuki24: Turns out it’s me not adding it in the first place 😅

yuki24:
I'll leave the final say to you but feel free to merge! Everything looks good to me.

PR #1131, opened 2018-06-28
User should see an error message when MP returns an error

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-227 and https://artsyproduct.atlassian.net/browse/PURCHASE-187

u

### DIFF
--- a/src/lib/Components/Bidding/Screens/Registration.tsx
+++ b/src/lib/Components/Bidding/Screens/Registration.tsx
@@ -1,3 +1,4 @@
+import { isEmpty } from "lodash"
 import React from "react"
 import { NativeModules, NavigatorIOS, View, ViewProperties } from "react-native"
 import { commitMutation, createFragmentContainer, graphql, RelayPaginationProp } from "react-relay"
@@ -17,6 +18,8 @@ import { PaymentInfo } from "../Components/PaymentInfo"
 import { Timer } from "../Components/Timer"
 import { Title } from "../Components/Title"
 import { Address, PaymentCardTextFieldParams, StripeToken } from "../types"
+
+import { BidResultScreen } from "./BidResult"
 
 import { Registration_me } from "__generated__/Registration_me.graphql"
 import { Registration_sale } from "__generated__/Registration_sale.graphql"
@@ -65,6 +68,16 @@ const bidderMutation = graphql`
     }
   }
 `
+
+const msgForNetworkError = {
+  message_header: "An error occurred",
+  message_description_md: "Please\ncheck your internet connection\nand try again.",
+}
+
+const msgForGraphQLError = {
+  message_header: "An error occurred",
+  message_description_md: "Please contact support@artsy.net with any questions.",
+}
 
 export class Registration extends React.Component<RegistrationProps, RegistrationState> {
   constructor(props) {
@@ -128,10 +141,9 @@ export class Registration extends React.Component<RegistrationProps, Registratio
     })
 
     commitMutation(this.props.relay.environment, {
-      onCompleted: () => this.createBidder(),
-      onError: () => {
-        console.log("ERROR!")
-      },
+      onCompleted: (_, errors) =>
+        isEmpty(errors) ? this.createBidder() : this.presentErrorResult(errors, msgForGraphQLError),
+      onError: this.presentErrorResult.bind(this),
       mutation: creditCardMutation,
       variables: {
         input: {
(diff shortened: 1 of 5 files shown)


src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -141,7 +141,7 @@ describe("when pressing bid button", () => {
         expect(mockphysics).not.toHaveBeenCalled()
       })
 
-      it("displays an error message", () => {
+      it("displays an error message on a netwrok failure", () => {
```
sweir27: #superminor `network`

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -162,6 +162,34 @@ describe("when pressing bid button", () => {
           })
         )
       })
+
+      it("displays an error message on a mutation failure", () => {
+        const error = {
+          message:
+            'https://stagingapi.artsy.net/api/v1/bidder?sale_id=leclere-impressionist-and-modern-art - {"error":"Invalid Sale"}',
```
sweir27: A little confusing-- the url in this error message is from a `createBidder` mutation it looks like, not a `createBidderPosition` mutation (although they'll likely have the same error message, so this is #minor).

sweir27:
Could you update the registration flow to have similar error handling? (Difference being instead of being able to call `presentError` we'd just be `console.log`ing at this point.

If you want to leave that out it's 👍 I'll just try to remember to put it in a subsequent PR!

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
yuki24: Right, I had the same feeling. I'll update this to a more realistic error message.

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
yuki24: 💥

yuki24:
I'm marking this as WIP as there is still some work that needs to be done.

src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -143,17 +155,34 @@ export class Registration extends React.Component<RegistrationProps, Registratio
...
       mutation: bidderMutation,
       variables: { input: { sale_id: this.props.sale.id } },
     })
   }
 
+  presentErrorResult(error, message) {
+    console.error(error)
+
+    this.props.navigator.push({
+      component: BidResultScreen,
```
yuki24: We are re-using the `BidResultScreen` screen, but this doesn't seem right. We should switch that to the new `RegistrationResult` component once https://artsyproduct.atlassian.net/browse/PURCHASE-202 is done. Alternatively, we could add a separate `ErrorScreen` component that only shows an error message, but I'll leave it to whoever works on PURCHASE-202.

yuki24:
This is good for review.

src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -143,17 +155,34 @@ export class Registration extends React.Component<RegistrationProps, Registratio
...
   createBidder() {
     commitMutation(this.props.relay.environment, {
-      onCompleted: () => {
-        console.log("COMPLETED!")
-      },
-      onError: () => {
-        console.log("ERROR!")
-      },
+      onCompleted: (results, errors) =>
+        isEmpty(errors) ? console.log("COMPLETED!", results) : this.presentErrorResult(errors, msgForGraohQLError),
```
yuki24: I left the `console.log` so we can incrementally make changes and work on https://artsyproduct.atlassian.net/browse/PURCHASE-203.

src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -66,6 +69,16 @@ const bidderMutation = graphql`
   }
 `
 
+const msgForNetworkError = {
+  message_header: "An error occurred",
+  message_description_md: "Please\ncheck your internet connection\nand try again.",
+}
+
+const msgForGraohQLError = {
```
sweir27: #minor `msgForGraolQLError` --> `msgForGraphQLError`

sweir27:
This looks great @yuki24 ! 👏

PR #1128, opened 2018-06-27
There should be more space visually between conditions of sale checkbox and place bid button

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-181

 * Prefer `Credit card` over `Credit Card`
 * The margin between the checkbox and the Place bid button should be 20px when alreadt agreed and 40px when needs to show a checkbox
 * Changed the margin for each row to be 20px so all the elements will be accommodated well in iPhone SE's smaller screen
# Screenshots:

screen_shot_2018-06-27_at_2_53_46_pm

screen_shot_2018-06-27_at_3_43_02_pm

#skip_new_tests

PR #1133, opened 2018-06-29
Update snapshot to fix builds #trivial

yuki24:
We merged two PRs (#1123 and #1128) almost at the same time and that seemed to have caused the build to fail due to a snapshot mismatch. This PR updates the snapshot to fix that.


erikdstock:
Merge on green.

PR #1137, opened 2018-07-03
[Bidflow] Registration Result screens

mennenia:
Finally 🙈 

Fix [PURCHASE-202](https://artsyproduct.atlassian.net/browse/PURCHASE-202)

Few notes:

- I ended up following the `BidResult` pattern (and hence changed the file name to `RegistrationResult` rather than payoff), so it fits with the proj

- @sweir27 there's a `RegistrationStatus` enum that is exported, which you can use to push to the right payoff screen from the `Registration` component. The results screen just takes the status as a prop. Will this work for you?

- The ghost button ("CONTINUE") is in the old style (with the capital Avant Garde font). @briansw @katarinabatina does this need updating to [the new button styles](https://app.zeplin.io/project/5acd19ff49a1429169c3128b/screen/5b06f4b36d69399e338c74f2), like the "Secondary Outline" style? I can create another ticket and sort that this week still :). 

Pics or it didn't happen:
screen shot 2018-07-03 at 09 20 42


mennenia:
Hmm something is up with the merge masters in the commit log :S. Odd!

src/lib/Components/Bidding/Screens/__stories__/RegistrationResult.story.tsx
```diff
@@ -0,0 +1,16 @@
+import { storiesOf } from "@storybook/react-native"
+import React from "react"
+
+// import { BidderPositionResult } from "../../types"
```
sweir27: #superminor extra import

PR #1138, opened 2018-07-03
Billing address keyboard optimizations #trivial

sweir27:
This PR:
- Updates the billing address form so you can scroll the elements above the keyboard
- Updates the billing address form to have "next" as the CTA and to advance to the next field when it is pressed

It was a little tricky to get that second part working. Major props to @orta for getting me started/helping (we do a similar thing on the consignments view).

I also found some weird behavior around the border colors on the input view, which I fixed by making sure we're passing through the `onBlur` as a differently-named-prop so we still get the default handler on the `TextInput`.

It's unfortunately difficult to test these things since I don't think we have `focus` and `blur` available in our test environment (in fact jest throws a warning saying so).

billing-address

### DIFF
--- a/src/lib/Components/Bidding/Components/Input.tsx
+++ b/src/lib/Components/Bidding/Components/Input.tsx
@@ -41,6 +41,7 @@ export class Input extends Component<InputProps, InputState> {
   }
 
   onFocus() {
+    // this fails right now
     this.setState({ borderColor: "purple100" })
   }
 
--- a/src/lib/Components/Bidding/Components/Input.tsx
+++ b/src/lib/Components/Bidding/Components/Input.tsx
@@ -4,6 +4,11 @@ import { TextInput, TextInputProps } from "../Elements/TextInput"
 
 interface InputProps extends TextInputProps, TextInputProperties {
   error?: boolean
+  // In order to create a ref to a lower-level input component from a higher-level parent,
+  // we pass in a function defined on the parent and call it with a pointer to the current Input.
+  refName?: any
+  createCustomRef?: any
+  customOnBlur?: any
 }
 
 class InputState {
@@ -19,20 +24,23 @@ export class Input extends Component<InputProps, InputState> {
     }
   }
 
-  componentWillReceiveProps(nextProps) {
-    this.setState({
-      borderColor: nextProps.error ? "red100" : "black10",
-    })
+  componentDidUpdate(prevProps) {
+    if (this.props.error !== prevProps.error) {
+      this.setState({ borderColor: this.props.error ? "red100" : "black10" })
+    }
   }
 
   onBlur() {
+    if (this.props.customOnBlur) {
+      this.props.customOnBlur()
+    }
+
     this.setState({
       borderColor: this.props.error ? "red100" : "black10",
     })
   }
 
   onFocus() {
-    // this fails right now
     this.setState({ borderColor: "purple100" })
   }
 
@@ -44,6 +52,7 @@ export class Input extends Component<InputProps, InputState> {
         fontSize={3}
         onBlur={() => this.onBlur()}
         onFocus={() => this.onFocus()}
+        ref={component => this.props.createCustomRef && this.props.createCustomRef(this.props.refName, component)}
         p={3}
         pb={2}
         {...this.props}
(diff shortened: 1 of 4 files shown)


src/lib/Components/Bidding/Components/Input.tsx
```diff
@@ -19,13 +22,17 @@ export class Input extends Component<InputProps, InputState> {
     }
   }
 
-  componentWillReceiveProps(nextProps) {
-    this.setState({
-      borderColor: nextProps.error ? "red100" : "black10",
-    })
+  componentDidUpdate(prevProps) {
```
sweir27: `componentWillReceiveProps` has been deprecated (and was firing an exorbitant amount of times). This works well and seems to be the preferred way.

src/lib/Components/Bidding/Components/Input.tsx
```diff
@@ -19,13 +22,17 @@ export class Input extends Component<InputProps, InputState> {
...
+    if (this.props.error !== prevProps.error) {
+      this.setState({ borderColor: this.props.error ? "red100" : "black10" })
+    }
   }
 
   onBlur() {
+    if (this.props.customOnBlur) {
```
sweir27: For our case, we pass in a `validate` function that we want to call `onBlur`. However, passing it in as a prop named `onBlur` meant that _this_ `onBlur` method was never being called. I renamed the one we pass in to `customOnBlur`.

src/lib/Components/Bidding/Components/Input.tsx
```diff
@@ -43,6 +50,7 @@ export class Input extends Component<InputProps, InputState> {
         fontSize={3}
         onBlur={() => this.onBlur()}
         onFocus={() => this.onFocus()}
+        ref={component => this.props.createCustomRef && this.props.createCustomRef(this.props.refName, component)}
```
sweir27: This is sort of tricky because we want to define a ref _on_ the parent `BillingAddress` component, but we want it to point to this `TextInput` because it's the only element which has `focus` and `blur` methods. Here, we pass the component up to the `createCustomRef` method on the parent element.

src/lib/Components/Bidding/Screens/BillingAddress.tsx
```diff
@@ -85,55 +110,113 @@ export class BillingAddress extends React.Component<BillingAddressProps, Billing
...
+    const inputs = [
+      this.fullName,
+      this.addressLine1,
+      this.addressLine2,
+      this.city,
+      this.stateProvinceRegion,
+      this.postalCode,
+    ]
+
+    for (const input of inputs) {
```
sweir27: In @orta 's Metadata component, it selects the next non-filled input to go to next. We could do that here but for now we just advance through them. I found the other behavior to be a little finicky so I'll keep it simple for now.

src/lib/Components/Bidding/Screens/BillingAddress.tsx
```diff
@@ -85,55 +110,113 @@ export class BillingAddress extends React.Component<BillingAddressProps, Billing
...
-              label="Postal code"
-              placeholder="Enter your postal code"
-              autoCapitalize="words"
-              {...this.propsForInput("postalCode")}
-            />
-
-            <Button text="Add billing address" onPress={() => this.onSubmit()} />
-          </Container>
-        </ScrollView>
+        <KeyboardAvoidingView behavior="padding" keyboardVerticalOffset={defaultVerticalOffset} style={{ flex: 1 }}>
```
sweir27: This `KeyboardAvoidingView` is what gives us the space below the button so it will scroll above the keyboard.

src/lib/Components/Bidding/Components/Input.tsx
```diff
@@ -4,6 +4,9 @@ import { TextInput, TextInputProps } from "../Elements/TextInput"
 
 interface InputProps extends TextInputProps, TextInputProperties {
   error?: boolean
+  refName?: any
+  createCustomRef?: any
```
orta: IMO, this code should probably have comments because it's pretty far from the normal react flow

sweir27: added :)

yuki24:
LGTM! There are some conflicts that need to be resolved, but other than that this is :+1:, so merge on green.

PR #1139, opened 2018-07-03
Updates registration view to advance to correct result screen #trivial

sweir27:
This PR adds the "glue", meaning it hooks up the transition between the registration screen and the results screen.

@mennenia 's [PR](https://github.com/artsy/emission/pull/1137) added a nice abstraction around the `RegistrationResult` screen. Since the use-case is a bit simpler here (we're not trying to map error messages returned from MP with ones defined in Emission), it works nicely to have all of the statuses and messages defined in the `RegistrationResult` component. We can just pass the `status` and it will handle all of the display.

### DIFF
--- a/src/lib/Components/Bidding/Screens/Registration.tsx
+++ b/src/lib/Components/Bidding/Screens/Registration.tsx
@@ -19,10 +19,10 @@ import { Timer } from "../Components/Timer"
 import { Title } from "../Components/Title"
 import { Address, PaymentCardTextFieldParams, StripeToken } from "../types"
 
-import { BidResultScreen } from "./BidResult"
-
 import { Registration_me } from "__generated__/Registration_me.graphql"
 import { Registration_sale } from "__generated__/Registration_sale.graphql"
+
+import { RegistrationResult, RegistrationStatus } from "./RegistrationResult"
 
 const Emission = NativeModules.Emission || {}
 
@@ -68,16 +68,6 @@ const bidderMutation = graphql`
     }
   }
 `
-
-const msgForNetworkError = {
-  message_header: "An error occurred",
-  message_description_md: "Please\ncheck your internet connection\nand try again.",
-}
-
-const msgForGraphQLError = {
-  message_header: "An error occurred",
-  message_description_md: "Please contact support@artsy.net with any questions.",
-}
 
 export class Registration extends React.Component<RegistrationProps, RegistrationState> {
   constructor(props) {
@@ -142,8 +132,12 @@ export class Registration extends React.Component<RegistrationProps, Registratio
 
     commitMutation(this.props.relay.environment, {
       onCompleted: (_, errors) =>
-        isEmpty(errors) ? this.createBidder() : this.presentErrorResult(errors, msgForGraphQLError),
-      onError: this.presentErrorResult.bind(this),
+        isEmpty(errors)
+          ? this.createBidder()
+          : this.presentRegistrationResult(RegistrationStatus.RegistrationStatusError),
+      onError: error => {
+        this.presentRegistrationError(error, RegistrationStatus.RegistrationStatusError)
+      },
       mutation: creditCardMutation,
       variables: {
         input: {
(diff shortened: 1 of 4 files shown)


src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -142,8 +132,12 @@ export class Registration extends React.Component<RegistrationProps, Registratio
 
     commitMutation(this.props.relay.environment, {
       onCompleted: (_, errors) =>
-        isEmpty(errors) ? this.createBidder() : this.presentErrorResult(errors, msgForGraphQLError),
-      onError: this.presentErrorResult.bind(this),
+        isEmpty(errors)
+          ? this.createBidder()
+          : this.presentRegistrationResult(RegistrationStatus.RegistrationStatusError),
+      onError: error => {
+        this.presentRegistrationError(error, RegistrationStatus.RegistrationStatusError)
```
sweir27: @yuki24 one question here-- should this be showing the network error screen or the generic error screen?

src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -156,27 +150,37 @@ export class Registration extends React.Component<RegistrationProps, Registratio
...
       mutation: bidderMutation,
       variables: { input: { sale_id: this.props.sale.id } },
     })
   }
 
-  presentErrorResult(error, message) {
-    console.error(error)
+  presentRegistrationSuccess({ createBidder }) {
+    const qualifiedForBidding = createBidder.bidder.qualified_for_bidding
+    if (Boolean(qualifiedForBidding) === true) {
```
yuki24: Is it necessary to wrap `qualifiedForBidding` with a `Boolean(...)`?

src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -156,27 +150,37 @@ export class Registration extends React.Component<RegistrationProps, Registratio
...
       title: "",
       passProps: {
-        // TODO: BidResultScreen requires `sale_artwork` and `sale`. Switch to the RegistrationResult component once
-        //       https://artsyproduct.atlassian.net/browse/PURCHASE-202 is done so we do not have to pass in
-        //       `sale_artwork`.
-        sale_artwork: {
-          sale: {},
-        },
-        bidderPositionResult: message,
+        status,
```
yuki24: :+1:

src/lib/Components/Bidding/Screens/Registration.tsx
```diff
@@ -156,27 +150,37 @@ export class Registration extends React.Component<RegistrationProps, Registratio
...
+    const qualifiedForBidding = createBidder.bidder.qualified_for_bidding
+    if (Boolean(qualifiedForBidding) === true) {
+      this.presentRegistrationResult(RegistrationStatus.RegistrationStatusComplete)
+    } else {
+      this.presentRegistrationResult(RegistrationStatus.RegistrationStatusPending)
+    }
+  }
 
+  presentRegistrationError(error, status) {
+    console.error("Registration.tsx", error)
```
yuki24: :+1:

src/lib/Components/Bidding/Screens/Registration.tsx
yuki24: I'm ok with either - my assumption is networrk errors are the majority of the errors that actually occur so the network error screen may be appropriate, but that's still an assumption. I think we should just monitor errors closely once released and iterate quickly on improving the error case scenarios.

src/lib/Components/Bidding/Screens/RegistrationResult.tsx
```diff
@@ -59,6 +65,9 @@ export class RegistrationResult extends React.Component<RegistrationResultProps,
         title = registrationPendingMessage.title
         msg = registrationPendingMessage.description
         break
+      case RegistrationStatus.RegistrationStatusNetworkError:
+        title = registrationNetworkErrorMessage.title
+        msg = registrationNetworkErrorMessage.description
```
yuki24: This is missing a case for the `"RegistrationStatusError"` key and the result screen may end up displaying an empty title and description.

src/lib/Components/Bidding/Screens/Registration.tsx
sweir27: Oh I guess not!

src/lib/Components/Bidding/Screens/RegistrationResult.tsx
sweir27: That ones is covered in the `default` case

PR #1143, opened 2018-07-04
Refactor the <Input> component

yuki24:
* [x] Add a test for the `inputRef` prop


src/lib/Components/Bidding/Components/Input.tsx
```diff
@@ -51,12 +50,12 @@ export class Input extends Component<InputProps, InputState> {
...
         borderColor={this.state.borderColor}
         fontSize={3}
-        onBlur={() => this.onBlur()}
-        onFocus={() => this.onFocus()}
-        ref={component => this.props.createCustomRef && this.props.createCustomRef(this.props.refName, component)}
         p={3}
         pb={2}
         {...this.props}
+        ref={this.props.inputRef}
+        onBlur={() => this.onBlur()}
```
sweir27: Ah, just sticking it below the `{...this.props}` is enough to make it override if an `onBlur` is passed in as props? Didn't realize that!

sweir27:
@yuki24 from a quick glance this looks ✨ and way simpler than the implementation I tried! Maybe I can pick your brain about how you got the refs to work in this way? Although it seems to have just _worked_. 😄

yuki24:
This PR is ready for review (except I was too lazy and didn't add details)

sweir27:
Haha, we went over this in person! Looks 👍 and glad it could be simplified.

PR #1146, opened 2018-07-06
Allow for editing the max bid once on the ConfirmBid screen without losing information #trivial

sweir27:
The goal of this PR is to address https://artsyproduct.atlassian.net/browse/PURCHASE-236 -- you should be able to navigate between the `ConfirmBid` screen and the `SelectMaxBid` screen in order to update your bid, without all of your information being lost in between.

@yuki24 and I went through many potential implementations of this. This is what seemed to be simplest (a potential more ideal version in the _future_, if the interaction becomes more complex, would be to introduce something like redux to hold state globally).

This PR:
- Adds a new component `SelectMaxBidEdit` which is _not_ relay-based. It shows the `MaxBidPicker` and updates the `selectedBidIndex` on the `ConfirmBid` component.
- Changes the `ConfirmBid` component to store `increments` and a `selectedBidIndex` instead of a `bid` object, as we'll need to pass the increments and the index to the new `SelectMaxBidEdit` component, and this felt cleanest.

Some more comments are inline!

gif here:
edit-max-bid

### DIFF
--- a/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
+++ b/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
@@ -24,6 +24,7 @@ import { Title } from "../Components/Title"
 import { Address, Bid, BidderPositionResult, PaymentCardTextFieldParams, StripeToken } from "../types"
 
 import { BidResultScreen } from "./BidResult"
+import { SelectMaxBidEdit } from "./SelectMaxBidEdit"
 
 import { ConfirmBid_me } from "__generated__/ConfirmBid_me.graphql"
 import { ConfirmBid_sale_artwork } from "__generated__/ConfirmBid_sale_artwork.graphql"
@@ -35,10 +36,11 @@ stripe.setOptions({ publishableKey: Emission.stripePublishableKey })
 export interface ConfirmBidProps extends ViewProperties {
   sale_artwork: ConfirmBid_sale_artwork
   me: ConfirmBid_me
-  bid: Bid
   relay?: RelayRefetchProp
   navigator?: NavigatorIOS
   refreshSaleArtwork?: () => void
+  increments: any
+  selectedBidIndex: number
 }
 
 interface ConfirmBidState {
@@ -49,6 +51,7 @@ interface ConfirmBidState {
   isLoading: boolean
   requiresCheckbox: boolean
   requiresPaymentInformation: boolean
+  selectedBidIndex: number
 }
 
 const MAX_POLL_ATTEMPTS = 20
@@ -139,6 +142,7 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
       isLoading: false,
       requiresCheckbox,
       requiresPaymentInformation,
+      selectedBidIndex: this.props.selectedBidIndex,
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -389,6 +409,10 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
     )
   }
 
+  private selectedBid(): Bid {
+    return this.props.increments[this.state.selectedBidIndex]
```
sweir27: We never refresh the increments on these "internal" views, so this shouldn't be a problem. However, we could add logic to avoid accessing arrays with indices that are out of bounds.

src/lib/Components/Bidding/Screens/SelectMaxBidEdit.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  render() {
+    const bids = this.props.increments || []
+
+    return (
+      <BiddingThemeProvider>
+        <Container m={0}>
+          <Title>Your max bid</Title>
+
+          <MaxBidPicker
```
sweir27: I looked into extracting the internals of this component into something shared, but it seemed to be more confusing than it would be worth. All of the meat is in the `MaxBidPicker` anyways, which _is_ shared.

src/lib/Components/Bidding/Screens/SelectMaxBidEdit.tsx
```diff
@@ -0,0 +1,60 @@
...
+            bids={bids}
+            onValueChange={(_, index) => this.setState({ selectedBidIndex: index })}
+            selectedValue={this.state.selectedBidIndex}
+          />
+
+          <Flex m={4}>
+            <Button text="Select bid" onPress={this.onPressNext} />
```
sweir27: I have to ask @briansw about the language here, but `Next` seemed odd in this context.

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -193,6 +195,33 @@ describe("when pressing bid button", () => {
...
+
+    const editScreen = fakeNavigator.nextStep()
+    expect(editScreen.root.type).toEqual(SelectMaxBidEdit)
+    expect(editScreen.root.props.selectedBidIndex).toEqual(0)
+    editScreen.root.instance.setState({ selectedBidIndex: 1 })
+
+    editScreen.root.findByType(Button).instance.props.onPress()
+
+    const confirmScreen = fakeNavigator.nextStep()
+    expect(confirmScreen.root.type).toEqual(ConfirmBid)
```
sweir27: I really wanted to test that the `confirmScreen` here displays `$46,000`, but I wasn't sure how to do that in the test environment. Basically, calling `fakeNavigator.nextStep()` to return the next component passes in the component's _original_ props, thus overriding what was previously changed.

Maybe you can look at this with me on Monday? If it's testable I would love to add that.

yuki24: You could possibly do the same as the line `210`, but make sure to `find` after `onPress()` since the `selectMaxBidRow` variable may be referencing a slate object. Happy to chat about this on Monday.

sweir27:
Updated that test. I still wish there was a _simple_ way to do this in our current configuration, but in any case this should be good for review 👍.

yuki24:
Looks great, this amount of duplication is still a manageable size and is a lot better than weird abstractions.

PR #1151, opened 2018-07-06
Handle errors returned from Stripe during tokenization #trivial

sweir27:
This PR addresses https://artsyproduct.atlassian.net/browse/PURCHASE-195 by explicitly catching errors from Stripe's API.

Note that we are not _charging_ the cards that we tokenize (so most of the potential checks won't surface at this stage), and Stripe will successfully tokenize any card that passes the same validations that are implemented for free in the form element we're using. Therefore, the errors that we're catching are more likely to do with us being unable to access stripe's API at all and as such there's not really a need to clear the credit card form when this happens.

Here's a helpful stackoverflow answer that discusses the types of validations Stripe does while tokenizing (same as what the form implements): https://stackoverflow.com/a/50878951

Already discussed this behavior with @briansw in slack!

The behavior addressed by this PR is:
- When you are on the `CreditCardForm`, you will already see validation errors as you enter your card information. If Stripe's API returns an error (like I said, this won't be due to any kind of extra validation on the card), we display an error message asking you to try again. This should also handle network errors and things like that. There's a video of this below.
- When you are on the `ConfirmBid` screen or the `Registration` screen and something happens when we try to access Stripe's API, we show the generic `error` screen for that flow.

cc-error

### DIFF
--- a/src/lib/Components/Bidding/Screens/CreditCardForm.tsx
+++ b/src/lib/Components/Bidding/Screens/CreditCardForm.tsx
@@ -3,6 +3,7 @@ import React, { Component } from "react"
 import { NavigatorIOS, ScrollView, StyleSheet, View } from "react-native"
 import stripe, { PaymentCardTextField, StripeToken } from "tipsi-stripe"
 
+import { Sans12 } from "lib/Components/Bidding/Elements/Typography"
 import BottomAlignedButtonWrapper from "lib/Components/Buttons/BottomAlignedButtonWrapper"
 import { BackButton } from "../Components/BackButton"
 import { BiddingThemeProvider } from "../Components/BiddingThemeProvider"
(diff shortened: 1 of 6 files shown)


src/lib/Components/Bidding/Screens/CreditCardForm.tsx
```diff
@@ -75,10 +69,26 @@ export class CreditCardForm extends Component<CreditCardFormProps, CreditCardFor
         <Button
           text="Add credit card"
           disabled={!this.state.valid}
+          inProgress={this.state.isLoading}
```
sweir27: Noticed we weren't passing the loading state to the button.

src/lib/Components/Bidding/Screens/CreditCardForm.tsx
```diff
@@ -59,14 +49,18 @@ export class CreditCardForm extends Component<CreditCardFormProps, CreditCardFor
...
-
-    this.props.onSubmit(token, this.state.params)
-    this.props.navigator.pop()
+    try {
+      const token = await stripe.createTokenWithCard({ ...params })
+      this.props.onSubmit(token, this.state.params)
+      this.setState({ isLoading: false })
+      this.props.navigator.pop()
+    } catch (error) {
+      this.setState({ isError: true, isLoading: false })
```
yuki24: Let's add a `console.error` so we can collect errors and be data-informed™️

yuki24:
Also, would the `stripe.createTokenWithCard` raise an error when it can't tokenize the card due to network errors or it just returns an error object?

sweir27:
@yuki24 I tested by turning off my wifi and an error was raised (taking us to the error state).

sweir27:
As per the conversation in slack, I think this should be good to go, and based on what we want as far as helpful error messages from gravity's API, we can follow up with a separate story. 😄

PR #1152, opened 2018-07-07
Add @artsy/palette

yuki24:
* [x] Run a visual QA (with Brian)
 * [x] Fix the type error `Property 'p' is missing in type '{ size: "2"; mt: 2; color: "black60"; children: (string | Element)[]; }'`
* [x] Validate that the timers for bid now also show

#skip_new_tests


zephraph:
@alloy there are some _really_ weird things happening with the types and I'm not sure what's up. For some reason it looks like some of the styled-system types are being listed as required.

orta:
At a guess - emission has stricter tsconfig settings which get applied to palette

zephraph:
That could be why it's failing here and not in reaction, but it looks like the actual types that are generated are wrong. I'm not 100% sure about that though.

alloy:
Can you share why you think they are wrong? Or do you want me to pull this down and take a stab?

orta:
I've had a quick dig around, I think you'll enjoy looking into it. I tried matching TypeScript versions, and the libs available to the runtime and still got the same error (easy to reproduce with a clone of this PR and `yarn type-check`)

It looks like when running in emission, code like `<Serif size="2">Remember me</Serif>` doesn't work because the type defs which look like:

```ts
export declare const Serif: StyledComponentClass<SansProps, any, Pick<SansProps, "p" | "color" | "size" | "italic" | "weight" | "verticalAlign" | "display" | "fontFamily" | "fontSize" | "lineHeight" | "maxWidth" | "m" | "mt" | "mr" | "mb" | "ml" | "mx" | "my" | "pt" | "pr" | "pb" | "pl" | "px" | "py" | "textAlign"> & {
    theme?: any;
}>;
```

Don't actually support the `Pick` aspect of the definition. I verified this by editing the typedef to 

```ts
export declare const Serif: StyledComponentClass<SerifProps, any, Pick<SerifProps, "size"> & {
    theme?: any;
}>;
```

which passes, but if you introduce any new string inside the `Pick` then it expects it to appear. Worth noting, these return typedefs are generated at runtime [from this  func](https://github.com/artsy/palette/blob/a0a0c94c590908c93bb3ba9b9002e0fcf85cfbf8/src/elements/Typography.tsx#L117-L152) in palette

alloy:
Seeing the same issue show up in Reaction now, investigating.

orta:
One solution might be to not infer the types in palette

alloy:
I guess the reaction one was a different issue after all. Will still have to give this PR a try locally.

orta:
Continuing this PR - the only problem now is that we need the ability to declare that a `Text` has `numberOfLines` which is React Native only.

```
src/lib/Components/Bidding/Components/BidInfoRow.tsx(29,31): error TS2339: Property 'numberOfLines' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<ThemedOuterStyledProps<Pick<SerifProps, ...'.

29               <Serif size="3" numberOfLines={1}>
                                 ~~~~~~~~~~~~~~~~~


src/lib/Components/Bidding/Screens/ConfirmBid.tsx(386,50): error TS2339: Property 'numberOfLines' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<ThemedOuterStyledProps<Pick<SerifProps, ...'.

386               <Serif size="4t" weight="semibold" numberOfLines={1} ellipsizeMode={"tail"}>
                                                     ~~~~~~~~~~~~~~~~~


src/lib/Components/Bidding/Screens/ConfirmBid.tsx(393,73): error TS2339: Property 'numberOfLines' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<ThemedOuterStyledProps<Pick<SerifProps, ...'.

393               <Serif italic size="2" color="black60" textAlign="center" numberOfLines={1} ellipsizeMode={"tail"}>
                                                                            ~~~~~~~~~~~~~~~~~

error Command failed with exit code 1.
```

orta:
So the above change is something that'll need to happen inside the styled-system dts files I believe. 

At runtime everything is 👍.

orta:
Semi-blocked by https://github.com/artsy/emission/pull/1152 to handle those main three typescript errors, and I'm working on a rebase.

zephraph:
Now you're unblocked 😄

package.json
```diff
@@ -20,7 +20,7 @@
     "prettier-write": "prettier --write",
     "prettier-project": "yarn run prettier-write 'src/**/*.{ts,tsx}'",
     "stylelint-project": " stylelint -- 'src/**/*.{ts,tsx}'",
-    "ci": "yarn type-check && yarn lint && yarn prettier-project -l && yarn test:ci",
```
damassi: @orta - noticed that `yarn prettier-project -l`, when writing on circle (and thus attempting to write to disk) will error out with an unhelpful error code 1. I was able to notice this due to a file buried in the tree that was presumably committed using `--no-verify`; when running prettier locally and committing those changes everything passed just fine. Any reason this was added to the CI step?

damassi:
This is now 🍏😄. 

Follow up steps will be to move our primitives (`reaction/Elements`) over to `@artsy/palette` so that things can be shared across platforms.

package.json
orta: That's the command for the CI version of prettier, it lists if you've not shipped prettier'd code - it's just a normal CI check to keep people shipping the right code - yeah, it's STDOUT output is real bad though

I'm not attached to it, it's easy to run prettier across an entire project if it becomes inconsistent eventually

damassi: Ah, that was def confusing (I should have checked to see what `-l` did).

Readded -- best to have this stuff in place so that things are caught early.

PR #1157, opened 2018-07-11
Use new metaphysics param for pulling max bid increments

sweir27:
This PR: https://github.com/artsy/metaphysics/pull/1163 introduced a new param that we can pass to the `increments` field in order to return increments based on your most recent max bid.

This just updates emission to use that!


sweir27:
Paired with @alloy on updating graphql in this repo so we could correctly parse the new schema.

PR #1162, opened 2018-07-12
Notify eigen on completion of registration #trivial

sweir27:
I borrowed this completely from what we do in the `ConfirmBid` flow (which I _think_ works as confirmed by @erikdstock yesterday).

Addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-254

#skip_new_tests


sweir27:
Ah, the `sale` mock in the `Registration` component used to be incorrect, so now that it's actually pulling in a timestamp we lend ourselves to odd CI failures. Will update to mock that correctly 👍!

sweir27:
Phew! Finally 🍏

PR #1163, opened 2018-07-13
Show spinner when pulling up bidding and registration flow #trivial

sweir27:
This PR addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-249

The bid/registration flows are slightly different than the others because we are explicitly sending props in our query renderer (as they are different from what's received). This caused us to always [skip](https://github.com/artsy/emission/blob/master/src/lib/utils/renderWithLoadProgress.tsx#L33) showing the `Spinner` on those views unless we explicitly pass the render function with null values here.


ashfurrow:
Merge on green.

PR #1164, opened 2018-07-13
Use Gravity's system time to consider offset for auction timer

yuki24:
addresses https://artsyproduct.atlassian.net/browse/PURCHASE-208

* [x] Adjust the local timer using Gravity's timer
* [x] Fix tests by injecting a Gravity clock object using React Context
* [x] Clean up test output
* [x] Add more test coverage around `GravityClock.ts`
* [x] ~Manually test it in a condition where the devise has a very slow connection~

### DIFF
--- a/src/lib/Components/Bidding/Context/TimeOffsetProvider.tsx
+++ b/src/lib/Components/Bidding/Context/TimeOffsetProvider.tsx
@@ -1,0 +1,98 @@
+import * as PropTypes from "prop-types"
+import * as React from "react"
+import { metaphysics } from "../../../metaphysics"
+
+interface SystemResults {
+  data: {
+    system: {
+      time: {
+        unix: number
+      }
+    }
+  }
+}
+
+const getLocalTimestampInMilliSeconds = () => {
+  return Date.now()
+}
+
+const fetchSystemTime = () => {
+  return metaphysics<SystemResults>({
+    query: `
+      {
+        system {
+          time {
+            unix
+          }
+        }
+      }
+    `,
+  })
+}
+
+const getGravityTimestampInMilliSeconds = async () => {
+  const startTime = getLocalTimestampInMilliSeconds()
+  const { data } = await fetchSystemTime()
+
+  const possibleNetworkLatencyInMilliSeconds = (getLocalTimestampInMilliSeconds() - startTime) / 2
+  const serverTimestampInMilliSeconds = data.system.time.unix * 1e3 + possibleNetworkLatencyInMilliSeconds
+
+  if (__DEV__) {
+    if (typeof jest === "undefined") {
+      console.log("Network latency in msec", possibleNetworkLatencyInMilliSeconds)
+      console.log("  Gravity clock in msec", serverTimestampInMilliSeconds)
+    }
+  }
+
+  return serverTimestampInMilliSeconds
+}
+
+const getOffsetBetweenGravityClock = async () => {
+  try {
+    const gravityClock = await getGravityTimestampInMilliSeconds()
+    const localClock = getLocalTimestampInMilliSeconds()
+
+    const offsetInMilliSeconds = localClock - gravityClock
+
+    if (__DEV__) {
+      if (typeof jest === "undefined") {
+        console.log("Gravity timestamp in msec", gravityClock)
+        console.log("  Local timestamp in msec", localClock)
+        console.log("           offset in msec", offsetInMilliSeconds)
+      }
+    }
+
+    return offsetInMilliSeconds
+  } catch (error) {
+    console.error("src/lib/Components/Bidding/Context/GravityClock.ts", error)
+
+    // If something goes wrong (e.g. network error), just fall back to "no offset" since there is nothing we can do.
+    return 0
+  }
+}
+
+export class TimeOffsetProvider extends React.Component {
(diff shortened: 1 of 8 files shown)


yuki24:
This is good for review code-wide. Adjusting the timer is working when the device has a reliable connection (== less network latency), but is untested under e.g. 3G. If the connection is not reliable then the user would probably have a difficult time bidding anyway, so that may not be a big issue.

src/lib/Components/Bidding/Context/TimeOffsetProvider.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+
+const getGravityTimestampInMilliSeconds = async () => {
+  const startTime = getLocalTimestampInMilliSeconds()
+  const { data } = await fetchSystemTime()
+
+  const possibleNetworkLatencyInMilliSeconds = (getLocalTimestampInMilliSeconds() - startTime) / 2
+  const serverTimestampInMilliSeconds = data.system.time.unix * 1e3 + possibleNetworkLatencyInMilliSeconds
+
+  console.log("Network latency in msec", possibleNetworkLatencyInMilliSeconds)
```
sweir27: Do we need to keep these console.logs? (here and below)

src/lib/Components/Bidding/Context/__tests__/TimeOffsetProvider-tests.tsx
```diff
@@ -0,0 +1,65 @@
...
+const dateNow = 1525983752000 // Thursday, May 10, 2018 8:22:32.000 PM UTC in milliseconds
+
+const consoleLog = console.log
+
+afterEach(() => {
+  console.log = consoleLog
+})
+
+it("injects timeOffsetInMilliSeconds as a context", () => {
+  // Debugging a timer is very tricky, so a few console.log calls are left in the code intentionally. Remove the line
```
sweir27: Ah, I guess this answers my question above.

storybook/storyLoader.js
```diff
@@ -1,89 +1,92 @@
...
-  require("../src/lib/Components/__stories__/DottedLine.story.tsx")
-  require("../src/lib/Containers/__stories__/Artist.story.tsx")
-  require("../src/lib/Containers/__stories__/Gene.story.tsx")
-  require("../src/lib/Containers/__stories__/Inquiry.story.tsx")
-  require("../src/lib/Containers/__stories__/Sale.story.tsx")
-  require("../src/lib/Scenes/Favorites/__stories__/Favorites.story.tsx")
-  require("../src/lib/Scenes/Home/Components/Sales/Components/__stories__/LotsByFollowedArtists.story.tsx")
-  require("../src/lib/Scenes/Home/__stories__/Home.story.tsx")
-  require("../src/lib/Scenes/Settings/__stories__/Settings.story.tsx")
+  require('../src/lib/Components/Artist/__stories__/ArtistArticles.story.tsx');
```
sweir27: Is this an intentional formatting change? I thought we preferred double quotes.

yuki24: I'm not sure why these changes came from, I thought there was a change in the storybook settings so I just added them as well. I just removed and this should be good.

src/lib/Components/Bidding/Context/__tests__/TimeOffsetProvider-tests.tsx
yuki24: Yep, but I'm open to suggestions. Do you think that's a good thing? I'm also ok with just removing all the `console.log`s.

yuki24: I slightly updated this piece to make it self-documenting by wrapping with a `if (__DEV__ && typeof jest === "undefined") { ... }` guard.

storybook/storyLoader.js
ashfurrow: FYI this fine gets autogenerated by Yarn based on [this config](https://github.com/artsy/emission/blob/bd7a8b1be56467f40f9aa1e9403a32a45cd8025e/package.json#L174-L178) but sometimes Prettier messes with the quote formats.

src/lib/Components/Bidding/Context/TimeOffsetProvider.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const getGravityTimestampInMilliSeconds = async () => {
+  const startTime = getLocalTimestampInMilliSeconds()
+  const { data } = await fetchSystemTime()
+
+  const possibleNetworkLatencyInMilliSeconds = (getLocalTimestampInMilliSeconds() - startTime) / 2
+  const serverTimestampInMilliSeconds = data.system.time.unix * 1e3 + possibleNetworkLatencyInMilliSeconds
+
+  if (__DEV__ && typeof jest === "undefined") {
```
alloy: I don’t know how `__DEV__` code is usually stripped from production builds, but maybe it looks just for `if (__DEV__) { … }`, in which case you may want to nest the `if (typeof jest === "undefined") { … }` part.

yuki24: Yeah I thought about it too. Let's make it safer by nesting.

PR #1165, opened 2018-07-13
Sets artwork and artist name labels on ConfirmBid component to use only one line

ashfurrow:
Fixes [PURCHASE-239](https://artsyproduct.atlassian.net/browse/PURCHASE-239).

Here's before:

simulator screen shot - iphone se - 2018-07-13 at 15 24 11

And here's after:

simulator screen shot - iphone se - 2018-07-13 at 15 23 41


sweir27:
👏 👏 💅

PR #1166, opened 2018-07-14
Show actual error messages if an error occurs while creating a credit card

sweir27:
This PR hopefully addresses the final piece of us being able to show helpful error messages on a credit card mutation failure.

Now that we'll [return errors](https://github.com/artsy/metaphysics/pull/1165) that we _handle_ in the `creditCardMutation` payload (as opposed to in the `errors` array), this updates the confirm bid and registration screens to correctly handle showing that error if it exists.

The diff is mostly updates to the graphql schema and snapshots. I realized I hadn't updated the `Registration` component to have the error modal yet, so this includes that as well.

Current logic is:
- If credit card mutation doesn't go through at all (some kind of network error) --> show generic registration network error screen
- If credit card mutation goes through but it returns an `errors` hash --> show the modal with the generic message
- If credit card mutation goes through and returns a `mutationError` that we handle --> show the modal with the returned message

reg-cc-error

confirm-bid-cc-error


yuki24:
The union type is actually very concerning to me, but let's see how it goes!

PR #1167, opened 2018-07-16
User should not be able to click into the billing address/cc/max bid edit forms while your bid is being placed

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-260

This PR just updates the `<BillingAddress>` and `<Registration>` components to not call any callbacks or just inject a null navigator when `isLoading` is true so users won't be able to launch other screens.

### DIFF
--- a/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
+++ b/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
@@ -371,7 +371,7 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
 
   render() {
     const { artwork, lot_label, sale } = this.props.sale_artwork
-    const { requiresPaymentInformation, requiresCheckbox } = this.state
+    const { requiresPaymentInformation, requiresCheckbox, isLoading } = this.state
 
     return (
       <BiddingThemeProvider>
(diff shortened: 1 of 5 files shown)


src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -399,12 +399,12 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
             <BidInfoRow
               label="Max bid"
               value={this.selectedBid().display}
-              onPress={() => this.goBackToSelectMaxBid()}
+              onPress={isLoading ? () => null : () => this.goBackToSelectMaxBid()}
```
ashfurrow: I think you want `isLoading ? null : () => this.goBackToSelectMaxBid() `, right? Right now, it's a function that returns null, which won't disable the button.

yuki24: This function is passed in to an `<TouchableWithoutFeedback>`, and a function that does nothing basically does the same as passing nothing from a user's perspective. I'm not that opinionated though. Thoughts?

ashfurrow: Oh, hmm I think I misunderstood. There's a bigger question I want to ask: should the rows and their buttons should have a visual "disabled" state while the bid is being placed? I think we should probably at least use a gray `color` for the button text (UI convention on iOS is that disabled buttons should be desaturated). Then we can use `null` as a disabled indicator. 

@briansw Can we get your input here?

sweir27:
I'm going to merge this as-is as I am pretty sure this captures the ticket as-defined. If we want to follow up with a design update we can do so in a separate PR!

PR #1168, opened 2018-07-16
Pressing continue on "Live bidding has started" payoff screen should redirect to the LAI interface

sweir27:
This PR addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-261

The actual change is quite small-- instead of linking to `/auction/sale-id`, we link to `live.artsy.net/sale-id` (or staging) when you click the continue button on the "Live bidding has started" interface.

This did involve adding a new variable to emission (which @orta helped me out with). Since we're setting defaults depending on whether you're in staging or not, I don't think this needs any follow-ups? But you would know better. 😄 

#native_no_changes


ashfurrow:
Merge on green.

ashfurrow:
Oh, actually Sarah, this is a breaking change from the Objective-C perspective. Can you increment the `native-code-version` in `package.json`?

ashfurrow:
I've up'd the native code version in #1169, we only need one of them.

ashfurrow:
Updated PR body to include `#native_no_changes` and restarted CI.

PR #1169, opened 2018-07-17
Preloading for BidFlow.

ashfurrow:
This adds preloading capabilities to the bid flow. Once merged+released, we'll need a small Eigen PR to actually call this new method. Fix for [Purchase-270](https://artsyproduct.atlassian.net/browse/PURCHASE-270).


alloy:
Nice! 👌

I‘d love to hear any feedback about this. One thing I was thinking is that maybe we need a script that verifies that the queries and variables referenced in the native code matches the query map and run that as part of the build process?

ashfurrow:
Mmm good idea! I'll look into that further.

sweir27:
Oh cool! Will we be refreshing this pre-query every time you revisit the artwork page? I'd be worried if that was cached in any way (since the increments in particular need to be up-to-date when you're bidding).

ashfurrow:
@sweir27 Yeah exactly, I'm looking into clearing the cache right now (this will be done by Eigen).

sweir27:
Cool, in that case... merge on green!

ashfurrow:
We might need to expose a new method to clear the cache for a query instead of the query _ID_, which is opaque from Eigen's perspective. Can follow up in another PR.

ashfurrow:
Oh, never mind! It looks like the preloading code always clears the cache when we preload:

https://github.com/artsy/emission/blob/12057b19e5d5d1bb08cd63962c5ed2dc8f0ec652/Pod/Classes/GraphQL/ARGraphQLQueryPreloader.m#L126-L129

ashfurrow:
Hmm, yeah, that _could_ be a problem. Since the response is set to `nil`, which is a sentinel value to indicate a request is in-flight, we expect no existing requests to be in-flight. So if someone were to try to load the request twice in a row (through preloading other otherwise) without checking/clearing the cache first, it'd cause an assertion failure on this line:

https://github.com/artsy/emission/blob/40f1ea0100f8f77daf3abe7a2a1ca708e99fddfa/Pod/Classes/GraphQL/ARGraphQLQueryCache.m#L161

Probably not a huge deal to work around, but I want to make sure I get the other tickets resolved for our RC tomorrow. I'll return to this when I can, time permitting.

alloy:
@ashfurrow Yeah, good reading 👍 It could well be that the existing logic can be improved, it was done naively and as simple as possible. (Note the real users would never get the assertion, they are compiled out for production builds. Not sure if that actually makes the experience _worse_ in production, though.)

ashfurrow:
Thanks! Yeah, I'd like to devote more time to this but we're a bit under the wire. Likely what'll happen is I'll return to this for 20% time after the initial release of the new bid flow (we're aiming to ship a release candidate for testing tomorrow morning NYC time).

alloy:
Sounds fab 👍

PR #1170, opened 2018-07-17
WIP: Auto-advancing through the fields should scroll down enough

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-256

I'm not sure if I like this fix, but this seems to be working fine...

 * [x] Replace `any`s
 * [ ] Write tests...?

## Screenshot

scroll

### DIFF
--- a/src/lib/Components/Bidding/Screens/BillingAddress.tsx
+++ b/src/lib/Components/Bidding/Screens/BillingAddress.tsx
@@ -27,8 +27,8 @@ interface StyledInputInterface {
   }
 }
 
-const StyledInput = ({ label, error, ...props }) => (
-  <Flex mb={4}>
+const StyledInput = ({ label, error, onLayout, ...props }) => (
+  <Flex mb={4} onLayout={onLayout}>
     <Serif16 mb={2}>{label}</Serif16>
     <Input mb={3} error={Boolean(error)} {...props} />
     {error && <Sans12 color="red100">{error}</Sans12>}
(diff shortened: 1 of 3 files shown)


src/lib/Components/Bidding/Screens/BillingAddress.tsx
```diff
@@ -188,13 +208,25 @@ export class BillingAddress extends React.Component<BillingAddressProps, Billing
...
+                }
               />
 
               <StyledInput
                 {...this.defaultPropsForInput("postalCode")}
                 label="Postal code"
                 placeholder="Add your postal code"
                 onSubmitEditing={() => this.presentSelectCountry()}
+                onFocus={() =>
+                  this.scrollView.scrollTo({
```
ashfurrow: I wonder if we can be a little smarter about this, since we're repeating _nearly_ the same code each time. Something like a function named `yPosition(this.postalCodeLayout, this.titleLayout)` or similar? What do you think?

yuki24:
Alright, I updated the PR so this works on all iOS devices and this is good for review. I'm still looking into how we could test this, but could be tricky as it depends largely on view elements.

src/lib/Components/Bidding/Screens/BillingAddress.tsx
```diff
@@ -232,4 +280,21 @@ export class BillingAddress extends React.Component<BillingAddressProps, Billing
...
+    const windowHeight = Dimensions.get("window").height
+
+    return Math.max(0, y - windowHeight + height + iOSAccessoryViewHeight + this.keyboardHeight + this.iPhoneXOffset)
+  }
+
+  private get verticalOffset() {
+    return this.iPhoneXOffset + 15
+  }
+
+  // TODO: Remove this once React Native has been updated
```
ashfurrow: I like that this comment is here 👍 I think it could be more specific – if there's an issue we can link to or a version number we can specify, I think that'd be a good addition here.

yuki24: @orta Do you know any links we can specify here?

orta: It is iPhone X support stuff, not working around a specific bug / implementation issue - so updating RN will fix it

PR #1171, opened 2018-07-17
dispatch notifications on the main thread

erikdstock:
#trivial 

cc @ashfurrow - The changes you requested to prevent eigen ui updates from running off the main thread.

### DIFF
--- a/Pod/Classes/EigenCommunications/ARNotificationsManager.m
+++ b/Pod/Classes/EigenCommunications/ARNotificationsManager.m
@@ -28,7 +28,9 @@ RCT_EXPORT_MODULE();
 
 RCT_EXPORT_METHOD(postNotificationName:(nonnull NSString *)notificationName userInfo:(NSDictionary *)userInfo)
 {
-    [[NSNotificationCenter defaultCenter] postNotificationName:notificationName object:nil userInfo:userInfo];
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [[NSNotificationCenter defaultCenter] postNotificationName:notificationName object:nil userInfo:userInfo];
+    });
 }
 
 @end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 ### Master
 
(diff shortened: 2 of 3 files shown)


Pod/Classes/EigenCommunications/ARNotificationsManager.m
```diff
@@ -28,7 +28,9 @@ - (void)paymentRequestPaidReceived:(NSNotification *)notification
 
 RCT_EXPORT_METHOD(postNotificationName:(nonnull NSString *)notificationName userInfo:(NSDictionary *)userInfo)
 {
-    [[NSNotificationCenter defaultCenter] postNotificationName:notificationName object:nil userInfo:userInfo];
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [[NSNotificationCenter defaultCenter] postNotificationName:notificationName object:nil userInfo:userInfo];
+    });
 }
```
alloy: @ashfurrow @erikdstock I wonder if it wouldn’t make more sense for _all_ of these method that interact with `NSNotificationCenter` to be performed on the main queue? 🤔 See https://facebook.github.io/react-native/docs/native-modules-ios#threading

ashfurrow: It's worth thinking about – the `ARNotificationsManager` doesn't do much else except route payment notifications from `NSNotificationCenter` into a `RCTEventEmitter` interface, so the code we've merged is _okay_, but I agree that pushing them all through the main queue would be good too 👍

@erikdstock the change would be basically reverting this code but adding the code that Eloy linked to:

```objc
- (dispatch_queue_t)methodQueue
{
  return dispatch_get_main_queue();
}
```

This isn't a release blocker, but let's try for it. I'll check in tomorrow 👍

alloy: Yeah it’s really more for future notification interactions that may be added and needing to repeat it every time, so definitely not something you need to prioritise over getting the work finished.

PR #1175, opened 2018-07-19
Ignores cache when loading BidFlow query

ashfurrow:
I took a stab at [Purchase-297](https://artsyproduct.atlassian.net/browse/PURCHASE-297). Based on what I read in [the docs](https://facebook.github.io/relay/docs/en/query-renderer.html), this is how Relay expects us to indicate that we want to bypass the cache for this query. **However**, I was unable to reproduce the original bug at all (it appears intermittent). My guess is this has something to do with debug vs. production builds of the app.

So, this PR _should_ fix the bug _in theory_, but we should test in the next beta to be sure.

### DIFF
--- a/src/lib/relay/QueryRenderers.tsx
+++ b/src/lib/relay/QueryRenderers.tsx
@@ -81,6 +81,7 @@ export const RegistrationFlowRenderer: React.SFC<BidderFlowRendererProps> = ({ r
           }
         }
       `}
+      cacheConfig={{ force: true }} // We want to always fetch latest sale registration status, CC info, etc.
       variables={{
         saleID,
       }}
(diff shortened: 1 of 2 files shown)


src/lib/relay/QueryRenderers.tsx
```diff
@@ -121,6 +121,7 @@ export const BidFlowRenderer: React.SFC<BidderFlowRendererProps> = ({ render, ar
           }
         }
       `}
+      cacheConfig={{ force: true }}
```
sweir27: Can we add this to the `RegistrationFlowRenderer` above as well? It's not likely to cause an issue but is another area that shouldn't be cached.

ashfurrow: Sure.

PR #1172, opened 2018-07-18
Disable Conditions of sale link and checkbox while bid is processing

yuki24:
finishes https://artsyproduct.atlassian.net/browse/PURCHASE-296

## Screenshot

out

#skip_new_tests

PR #1177, opened 2018-07-20
Disables condition of sale link if the user has previously accepted them #trivial

ashfurrow:
This PR fixes a slight omission from #1172 where the conditions of sale link was enabled in the case where the user had previously accepted them and now checkbox was rendered. [Here](https://github.com/artsy/emission/pull/1172/files#diff-8a3a841add9adeac68c88de41a452cc4R446) is the line of code.

Self-assigning because everyone else is out and this is trivial. Wonder if merge on green works in PR bodies?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### Master
 
 * Fixes back button asset for BidFlow - ash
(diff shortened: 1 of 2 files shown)


ashfurrow:
I guess it doesn't. Merge on green, please.

PR #1179, opened 2018-08-29
Update README with full clean step

damassi:
After returning to Emission from some time away, ran into an outdated dep issue that this command fixed.


orta:
Bit worried about that fail, but this README change shouldn't affect that

PR #1182, opened 2018-09-04
Adds the labs feature from Eigen into Emission

orta:
Part 1 of 2. 

Part 2 ( tomorrow ) will include reading from this inside the JS. I got a conference party to crash though 🍡 

image


damassi:
Nicee 👍

PR #1184, opened 2018-09-05
Labs v2 with an exposed API

orta:
screen recording 2018-09-05 17_04_34

This exposes the API for feature flags to into Emission. If you change any of the lab features, then it reloads the native modules and React Native runtime so that things that don't require a restart can run without any additional work 🌮 

It's exposed as an object because in Eigen we have both [lab features](https://github.com/artsy/eigen/blob/1f8d3c00952b98aed514b3a352e60a754d9b9394/Artsy/App/AROptions.m#L1-L14) and "app features" which are server-controlled feature strings [vendored](https://github.com/artsy/eigen/blob/master/Artsy/App/Echo.json) from Artsy Echo, which we'll want to expose.


CHANGELOG.md
```diff
@@ -16,10 +16,9 @@
 
 ### Master
 
-### 1.6.0
-
```
orta: not shipped yet

Example/Emission/AppDelegate.m
```diff
@@ -98,27 +99,41 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+                                                                authenticationToken:accessToken
+                                                                          sentryDSN:nil
+                                                               stripePublishableKey:[keys stripePublishableKey]
+                                                                   googleMapsAPIKey:nil
+                                                                         gravityURL:setup.gravityURL
+                                                                     metaphysicsURL:setup.metaphysicsURL
+                                                                      predictionURL:setup.predictionURL
+                                                                          userAgent:@"Emission Example"
+                                                                            options:setup.options];
+
```
orta: The `:`'s being aligned is how obj-c is supposed to handle these long method calls. I believe the aim is to balance the label and values. I dunno.

Pod/Classes/Core/AREmission.m
```diff
@@ -87,6 +90,8 @@ - (instancetype)initWithConfiguration:(AREmissionConfiguration *)config packager
   NSParameterAssert(config.authenticationToken);
   NSParameterAssert(config.gravityURL);
   NSParameterAssert(config.metaphysicsURL);
+  NSParameterAssert(config.predictionURL);
+  NSParameterAssert(config.options);
 
   if ((self = [super init])) {
     // When adding a new native module, remember to add it
```
orta: I fixed up the nullability in this class while I was here

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "emission",
   "version": "1.5.15",
-  "native-code-version": 6,
+  "native-code-version": 7,
```
orta: I changed the native bridging API, so I've bumped the version

Example/Emission/ARLabOptions.m
```diff
@@ -1,17 +1,52 @@
...
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    //
+    // The keys in this hash are how you can access the value in Emission code:
+    //
+    // import { NativeModules } from "react-native"
+    // const { Emission } = NativeModules
+    //
+    // if (Emission.options.example) {
+    //   [something]
```
damassi: 👍

CHANGELOG.md
damassi: Ah yeah, I removed this in my other PR

damassi:
This looks very straightforward 👍 Will give it a spin in my other PR now.

PR #1185, opened 2018-09-11
[Toolchain] Update Typescript to 3.0.3

damassi:
Upgrades Typescript and TSLint to latest, and followed a few work-arounds to get it green since we're a few versions behind in terms of `react-native` / `react`.

### DIFF
--- a/patches/@types/react-native+0.52.1.patch
+++ b/patches/@types/react-native+0.52.1.patch
@@ -1,0 +1,15 @@
+patch-package
+--- a/node_modules/@types/react-native/index.d.ts
++++ b/node_modules/@types/react-native/index.d.ts
+@@ -8435,8 +8435,9 @@ export type Clipboard = ClipboardStatic;
+ export var DatePickerAndroid: DatePickerAndroidStatic;
+ export type DatePickerAndroid = DatePickerAndroidStatic;
+ 
+-export var Geolocation: GeolocationStatic;
+-export type Geolocation = GeolocationStatic;
++// https://github.com/DefinitelyTyped/DefinitelyTyped/issues/24573
++// export var Geolocation: GeolocationStatic;
++// export type Geolocation = GeolocationStatic;
+ 
+ /** http://facebook.github.io/react-native/blog/2016/08/19/right-to-left-support-for-react-native-apps.html */
+ export var I18nManager: I18nManagerStatic;
--- a/tslint.json
+++ b/tslint.json
@@ -3,10 +3,9 @@
   "rules": {
     "arrow-parens": false,
     "interface-name": [true, "never-prefix"],
-    "max-classes-per-file": [false],
+    "max-classes-per-file": false,
     "member-access": [false, "check-accessor", "check-constructor"],
     "no-console": [true, ["error", ["warn", "error"]]],
-    "no-unused-variable": [true, { "ignore-pattern": "^_" }],
     "object-literal-sort-keys": false,
     "ordered-imports": true,
     "switch-default": false,
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
 src/__generated__/*.ts
 src/__generated__/*.json
+src/__generated__/*.graphql.ts
 node_modules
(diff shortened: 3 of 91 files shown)


patches/@types/react-native+0.52.1.patch
```diff
@@ -0,0 +1,15 @@
+patch-package
+--- a/node_modules/@types/react-native/index.d.ts
++++ b/node_modules/@types/react-native/index.d.ts
+@@ -8435,8 +8435,9 @@ export type Clipboard = ClipboardStatic;
+ export var DatePickerAndroid: DatePickerAndroidStatic;
+ export type DatePickerAndroid = DatePickerAndroidStatic;
+ 
+-export var Geolocation: GeolocationStatic;
+-export type Geolocation = GeolocationStatic;
++// https://github.com/DefinitelyTyped/DefinitelyTyped/issues/24573
```
alloy: Gah, didn’t know we had this issue. This should really be worked around by getting rid of `@types/node`, because RN is not Node.

Looks like it’s being pulled in by enzyme/cheerio?? No clue why that would be a good idea of them.

```
=> Found "@types/node@8.5.1"
info Reasons this module exists
   - "enzyme#cheerio#parse5" depends on it
   - Hoisted from "enzyme#cheerio#parse5#@types#node"
```

damassi:
@alloy - Oh weird, I wonder what happened there re relay fragments; my first push only sent up a couple files. I think this might have just happened on rebase -- investigating.

patches/@types/react-native+0.52.1.patch
damassi: Enzyme discusses this a bit here: https://github.com/airbnb/enzyme/blob/master/docs/api/render.md#static-rendering-api

damassi:
Yup, this looks like a Prettier thing which I presume was triggered after Typescript re-ran post update.

tslint.json
```diff
@@ -3,10 +3,9 @@
   "rules": {
     "arrow-parens": false,
     "interface-name": [true, "never-prefix"],
-    "max-classes-per-file": [false],
+    "max-classes-per-file": false,
     "member-access": [false, "check-accessor", "check-constructor"],
     "no-console": [true, ["error", ["warn", "error"]]],
-    "no-unused-variable": [true, { "ignore-pattern": "^_" }],
```
damassi: See https://github.com/palantir/tslint/pull/3919

orta:
Part of me wonders if we should add `**/__generated__/*.graphql.ts` to the prettier ignore? Otherwise feels pretty fine to me

damassi:
@orta - I would agree with this; I don't think we should transform generated files.

patches/@types/react-native+0.52.1.patch
alloy: I actually meant that no package should list `@types/*` packages as runtime deps. In this case it was `parse5` (didn’t read properly yesterday). You can see that in v5.1.0 they removed this https://github.com/inikulin/parse5/commit/4350f6a03dcd57dd9b3d8ba3c962ebf31a5f7f54#diff-b9cfc7f2cdf78a7f4b91a753d10865a2, so the fix would be for us to force resolution of that version instead of the one that cheerio depends on.

alloy:
This LGTM. It would be great if you could address that `@types/node` issue, but not a necessity.

patches/@types/react-native+0.52.1.patch
damassi: Interesting: 

screen shot 2018-09-12 at 12 37 59 pm

I removed the patch, and updated the resolution, yet the error re-appears.

alloy: Can you do `yarn why @types/node` instead?

damassi: screen shot 2018-09-12 at 12 43 39 pm

alloy:
@damassi Sigh. Well, commit that anyways, there’s absolutely no need for `@types/node` in Emission. Then restore the RN typedef patch and go ahead with the merge 👍

damassi:
Ok mergin'

PR #1186, opened 2018-09-11
[Lib] Upgrade @artsy/palette@2.16.0

damassi:
Depends on https://github.com/artsy/emission/pull/1185

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,11 @@
 
 ### Master
 
-* Adding artwork badges to brick - matt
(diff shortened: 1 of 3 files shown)


damassi:
Hmm 🤔 Snapshots are failing on CI and passing locally. I've tried: 
- `find . -name "*.snap" -type f -delete`, and then subsequently rengenerated
- Clearing Travis caches

alloy:
The failures all seem to be related to SC class names, so maybe it’s an issue with your local cache from the SC Babel plugin?

PR #1187, opened 2018-09-12
A proof of concept of using Core Animation instead of a movie file.

alloy:
animated-carousel


alloy:
Going to close this, as we’re going to go with a movie file for now but I wanted to keep this code around for the future.

PR #1190, opened 2018-09-24
Emission deployment improvements

orta:
OK, so this PR should:

- Simplify Emission deployments so that it should "just work" - I think. So long as you're not doing a React Native update, which is an edge-case because so few PRs do that.
- Simplify Emission's repo - no more submodules. Elan is effectively dead thanks to palette, and we can rely on Metaphysics having it's schema in the right place thanks to https://github.com/artsy/README/issues/31

## Deployment faff

The idea is now that the release script does most of the work, and the release script has had a one over. This also includes the deploy queries to metaphysics, which now does all the work and you just have to fill out a PR form on GitHub.

You should be able to do everything by updating the changelog, then running `yarn release`.

It also updates prettier, because ours was over a year old and didn't support graphql files. This forces a README change too. I'm alright with this.

#skip_new_tests


orta:
looks like the prettier update also updated a bunch of the TS files - I did nothing there

orta:
Deprecates https://github.com/artsy/emission/pull/1124

alloy:
Nice 👌

PR #1199, opened 2018-10-12
Recommends the omakase extension and sets cSpell to be hints only

orta:
Now that all these extensions are consolidated

#trivial


damassi:
👍 

merge on green

PR #1203, opened 2018-10-19
Removes the BNMO feature flags

orta:
Sibling PR to https://github.com/artsy/reaction/pull/1428

### DIFF
--- a/src/lib/Scenes/Home/index.tsx
+++ b/src/lib/Scenes/Home/index.tsx
@@ -1,5 +1,5 @@
 import React from "react"
-import { AppState, NativeModules, View } from "react-native"
+import { AppState, View } from "react-native"
 import ScrollableTabView from "react-native-scrollable-tab-view"
 import styled from "styled-components/native"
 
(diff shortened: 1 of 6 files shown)


src/lib/Scenes/Home/index.tsx
```diff
@@ -133,13 +129,6 @@ export default class Home extends React.Component<Props, State> {
             <SalesRenderer render={renderWithLoadProgress(Sales)} />
           </Tab>
         </ScrollableTabView>
-
-        {showConsignmentsSash && (
-          <DarkNavigationButton
-            title="Sell works from your collection through Artsy"
-            onPress={this.openLink.bind(this)}
-          />
```
damassi: Is it ok that we're removing the consignments sash completely?

orta: I feel like that's the original logic though, Stephen is gonna be away for a while, so not sure what to do with that though. That said, I think this was the only way you can start a consignment outside of an email - so might be worth waiting on someone with a definitive answer

PR #1198, opened 2018-10-12
Add a fairs root view controller

orta:
This adds all the infra to create a new screen in the iOS app.

screen shot 2018-10-12 at 3 52 55 pm

PR #1205, opened 2018-10-23
Add shows detail root view controller

javamonn:
Inspired by #1198 and [documentation](https://github.com/artsy/emission/blob/master/docs/adding_new_components.md), adds a new ViewController and react component scaffolds for the shows detail page. Implements [LD-11](https://artsyproduct.atlassian.net/browse/LD-11). Implemented while pairing with @l2succes and @ashleyjelks.

#skip_new_tests (We'll add tests as we build those components)


src/lib/Scenes/ReferenceShow/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+        keyExtractor={(item, index) => item.type + String(index)}
+      />
+    )
+  }
+}
+
+export default createFragmentContainer(
+  ReferenceShow,
+  graphql`
+    fragment ReferenceShow_show on Show {
```
javamonn: There's some churn between `ReferenceShow` and `Show` for this container. We'd like to use `Show`, but have to refactor the [existing conflicting definition](https://github.com/javamonn/emission/blob/bdd134aa35b520bbacba0d642400043992c1bb3e/src/lib/Components/Artist/Shows/Show.tsx#L46). I got partway through doing this in this PR but reverted after running into some unrelated issues, thinking I'll follow open a follow up PR to make the naming more consistent.

javamonn: I discussed this with @alloy a bit and he told me about coming to the decision of applying better namespacing to fragments. I'll rename this to something like `LocalDiscoveryShow`, and we can ticket the follow up work to namespace other fragments, e.g.`Show_show` should become something like `ArtistsShow_show`

javamonn: Discussed this a bit more with @l2succes and ended up waffling - the local discovery show is intended to be the canonical showI think, so it seems to make sense to namespace the Artist's `show`. I'm open to discussion though.

l2succes:
merge on green

alloy:
Gotcha, makes sense 👍

PR #1206, opened 2018-10-29
Add artists view ld8

ashleyjelks:
screen shot 2018-10-30 at 2 57 30 pm
Implement Artwork grid in Show View with Infinite Scroll.

[Links to LocalDiscovery 8 in Jira](https://artsyproduct.atlassian.net/browse/LD-8)


Pod/Classes/ViewControllers/ARArtworkComponentViewController.h
```diff
@@ -0,0 +1,21 @@
+#import "ARComponentViewController.h"
```
orta: This is only needed for *root* components, you should only need these type of changes if you're planning to replace `/artworks/:id` for all of artsy

src/lib/Scenes/Show/Components/Artworks/index.tsx
```diff
@@ -1,14 +1,25 @@
+import GenericGrid from "lib/Components/ArtworkGrids/GenericGrid"
 import React from "react"
-import { Text } from "react-native"
+import { createFragmentContainer, graphql } from "react-relay"
 
 interface Props {
-  artworks: any[]
+  show: any
```
l2succes: We should actually use the `Artworks_show` type here

Pod/Classes/ViewControllers/ARArtworkComponentViewController.h
ashleyjelks: This is resolved!

l2succes:
👍  Muy bien

PR #1207, opened 2018-10-29
Show Map/Address component

l2succes:
Since this component needs to display a static map, I started exploring potential libraries/solutions for the Map library as I'd like the static map to be consistent with the main map displaying current shows.

Here are the libraries I'm currently testing
- React Native Maps: https://github.com/react-community/react-native-maps
- Mapbox RN GL: https://github.com/mapbox/react-native-mapbox-gl

Edit: Based on my research and our requirements I believe Mapbox is best option.

screen shot 2018-10-29 at 11 24 41 pm


src/lib/Scenes/Show/Components/Location/index.tsx
```diff
@@ -1,14 +1,74 @@
...
+    const { lat, lng } = location.coordinates
+
+    return (
+      <Flex>
+        <Map
+          styleURL={Mapbox.StyleURL.Light}
+          centerCoordinate={[lng, lat]}
+          zoomLevel={13}
+          logoEnabled={false}
+          scrollEnabled={false}
```
anandaroop: Not sure if it's explicitly specified in the comps, but for static map behavior would we want to set `zoomEnabled` to false as well?

package.json
```diff
@@ -76,6 +77,7 @@
     "react": "^16.3.0-alpha.1",
     "react-native": "0.54.4",
     "react-native-hyperlink": "0.0.11",
+    "react-native-maps": "^0.22.0",
```
orta: Is this still needed?

l2succes: Nope, good catch

src/lib/Scenes/Show/Components/Location/index.tsx
l2succes: I think it defaults to `false` (even though the docs say `none`) But I'll set it anyway just in case

https://github.com/mapbox/react-native-mapbox-gl/blob/master/docs/MapView.md

src/lib/Scenes/Show/index.tsx
```diff
@@ -6,8 +6,8 @@ import { createFragmentContainer, graphql } from "react-relay"
 
 import { Artists } from "./Components/Artists"
 import { ArtworksContainer as Artworks } from "./Components/Artworks"
-import { Location } from "./Components/Location"
-import ShowHeader from "./Components/ShowHeader"
+import { LocationContainer as Location } from "./Components/Location"
+import { ShowHeaderContainer as ShowHeader } from "./Components/ShowHeader"
```
l2succes: @ashleyjelks @javamonn Let's follow this naming convention for importing Relay Containers which we introduced in reaction

PR #1208, opened 2018-10-29
Add show detail header

javamonn:
Implements [LD-5](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=63&projectKey=LD&modal=detail&selectedIssue=LD-5), adds (largely unstyled) header components to the local discovery show header page.

As discussed offline, I originally wanted to include porting over reaction's relay testing devtools, but am separating that out to a separate PR due to the larger than anticipated amount of work.


src/lib/Components/Buttons/CaretButton.tsx
```diff
@@ -0,0 +1,19 @@
+import { Display, Flex } from "@artsy/palette"
+import React from "react"
+import { GestureResponderEvent, Image, TouchableWithoutFeedback } from "react-native"
```
l2succes: FYI We have a custom image component (`OpaqueImageView`) which is built on top of our native image cache.
It's necessary to use here but just so you know

src/lib/Scenes/Show/Components/ShowHeader/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <Carousel
+          sources={(images || []).map(({ url: imageURL, aspect_ratio: aspectRatio }) => ({ imageURL, aspectRatio }))}
+        />
+        <Serif size="5t" weight="semibold" textAlign="center">
+          {name}
+        </Serif>
+        <Serif size="3" textAlign="center">
+          {exhibition_period}
+        </Serif>
+        <Box px={space(2)}>
```
damassi: Is the full component tree wrapped in a `<Theme>` component from `@artsy/palette`? If so these values can be pulled in directly via styled-components' theme, e.g., 

```js
<Box px={2}>
```

src/lib/Scenes/Show/Components/ShowHeader/Components/Chip.tsx
```diff
@@ -0,0 +1,18 @@
+import { color, Flex, Sans, space } from "@artsy/palette"
+import React from "react"
+import styled from "styled-components/native"
+
+// FIXME: Replace `background-color` with `background` prop on Flex on `@artsy/palette` upgrade
```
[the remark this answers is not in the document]
javamonn: I couldn't get `bg` or `background` to work, I just tried again after wrapping with `Theme` as well. I didn't dive too far into it before, but this incompatibility was what prompted me [this discussion](https://artsy.slack.com/archives/C02BAQ5K7/p1540491403000100) - I'm figuring it's due to the old version of palette.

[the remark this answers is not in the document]
l2succes: Right, we should try again once we upgrade palette

src/lib/Scenes/Show/Components/ShowHeader/Components/Carousel.tsx
```diff
@@ -0,0 +1,143 @@
...
+    const {
+      nativeEvent: {
+        contentOffset: { x },
+      },
+    } = ev
+    const { pageWidth, animationTargetIdx } = this.state
+    const currentPageIdx = Math.round(x <= 0 ? 0 : x / pageWidth)
+    if (animationTargetIdx !== null) {
+      // Clear indicator "lock" if we've reached our destination, otherwise
+      // ignore events triggererd by intermediate animated pages.
```
alloy:
```suggestion
      // ignore events triggered by intermediate animated pages.
```

src/lib/Scenes/Show/Components/ShowHeader/__tests__/index-tests.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react"
+import "react-native"
+import * as renderer from "react-test-renderer"
+import { ShowHeader } from "../index"
+
+it("looks correct when rendererd", () => {
```
alloy:
```suggestion
it("looks correct when rendered", () => {
```

PR #1210, opened 2018-10-30
[WIP] Update React Native to 0.57.4

orta:
Re: LD-22

This is just step 1, the module + types but it's momentum. Will keep working on this branch.
Step 2 is done: having it run.

screen shot 2018-10-31 at 3 28 02 pm


fbartho:
Does this support `const enums` and TS namespaces?

src/lib/Components/Inbox/Conversations/index.tsx
```diff
@@ -111,10 +111,13 @@ export class Conversations extends Component<Props, State> {
           keyExtractor={(_item, index) => String(index)}
           scrollEventThrottle={500}
           renderItem={({ item }) => {
+            // TODO: What is up with this `as any`?
+            //       there's a type mismatch here, as item above is a `ConversationSnippetProps`
+            //       but conversation is a `ConversationSnippet_conversation`
```
alloy: [This](https://github.com/artsy/emission/blob/535d3593f5a01641bed84f2281861b26768fb308/src/lib/Components/Inbox/Conversations/index.tsx#L30) should be:

```ts
interface State {
  conversations: Array<Props["me"]["conversations"]["edges"][0]["node"]> | null
}
```

(That could be simplified by making a separate fragment/container for the individual (conversation) nodes, which would give you a new type.)

And `me` should get typed [here](https://github.com/artsy/emission/blob/535d3593f5a01641bed84f2281861b26768fb308/src/lib/Components/Inbox/Conversations/index.tsx#L52)

alloy: Btw, in that example `Props["me"]` is equal to `Conversations_me`, I go back and forth on what way to use 🤔

orta:
> Does this support const enums and TS namespaces?

No, because it uses Babel for transpiration - see ` src/lib/utils/track.ts → src/lib/utils/track/index.ts` in the diff.

zephraph:
👋 So, what can I do to help move this along? It's caused an issue where we've had to [pin react-spring](https://github.com/artsy/palette/pull/135#issue-234386685) and that same issue is blocking us from upgrading to styled-components 4.

orta:
we got further on this, looking like right now we're somewhat blocked on jest mocking a decorator.

We were seeing "decorator is not a function" from "_applyDecoratedDescriptor" in babel/runtime, which is the tracking function inside a component. Unmocking `react-tracking` fails the tests but transpires correctly. 

To continue, start by looking at either:

- manually mocking react tracking in some form
- find another way to "mock" by using a different import or changing something at runtime
- see if it's a bug in jest

iskounen:
This PR will also fix: https://artsyproduct.atlassian.net/browse/TUDI-16

l2succes:
Closing this in favor of https://github.com/artsy/emission/pull/1551

PR #1211, opened 2018-10-31
Fixes crash from possible division by zero

ashfurrow:
I was looking at Eigen's Sentry logs and came across [this new crash from 4.3.2](https://sentry.io/artsynet/eigen/issues/748150611/?query=is:unresolved):

```
Application threw exception CALayerInvalidGeometry: CALayer position contains NaN: [nan 110]
```

(It's only happened to one user so far.)

I took a look at the code and it looks like in some edge cases, we might be dividing by zero to calculate the aspect ratio. I put in a check for `isnan` first, hopefully this will fix it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Updates cocoapods (1.5.3) - luc
 - Adds Mapbox + Location Component - luc
 - Adds Artwork grids with infinite scroll to Show View - ashley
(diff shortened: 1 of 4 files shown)


ashfurrow:
Danger passed but the error comment is still there. Hmm.

orta:
We might be using an old build of danger locally - I know there was an API change for deleting posts

PR #1212, opened 2018-10-31
Upgrades Danger #trivial

ashfurrow:
[Suggested here](https://github.com/artsy/emission/pull/1211#issuecomment-434762237) to resolve a case where a PR comment wasn't deleted once the build was green.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -366,9 +366,10 @@
     call-me-maybe "^1.0.1"
     glob-to-regexp "^0.3.0"
 
-"@octokit/rest@^15.5.0":
-  version "15.8.2"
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -3442,15 +3454,17 @@ danger-plugin-jest@^1.1.0:
     serve "^5.1.5"
 
 danger@^3:
-  version "3.7.18"
-  resolved "https://registry.yarnpkg.com/danger/-/danger-3.7.18.tgz#944e0e2786b3f40c044345f8185c0e2e61131cdc"
+  version "3.9.0"
```
orta: Danger's latest version is 6.x, this probably won't be enough

PR #1214, opened 2018-11-02
Add Artists list to Show detail scene

javamonn:
Implements [LD-9](https://artsyproduct.atlassian.net/browse/LD-9).

#skip_new_tests - `Artists/index.tsx` and `ShowHeader/index.tsx` just re-export tested components.


src/lib/Scenes/Show/Components/Artists/Artists.tsx
```diff
@@ -0,0 +1,130 @@
...
+          },
+          optimisticResponse: {
+            followArtist: {
+              artist: {
+                __id,
+                is_followed: !is_followed,
+              },
+            },
+          },
+          updater: store => {
```
javamonn: Not sure if I need  both`updater` and `optimisticResponse`, I saw both used in some places and in others only one...

src/lib/Scenes/Show/Components/Artists/__tests__/Artists-tests.tsx
```diff
@@ -0,0 +1,73 @@
...
+      query={graphql`
+        query ArtistsTestsQuery {
+          show(id: "anderson-fine-art-gallery-flickinger-collection") {
+            ...Artists_show
+          }
+        }
+      `}
+      mockResolvers={{
+        Show: () => ShowFixture,
+        Artist: () => ({ name: "Hans Hofmann", is_followed: false, id: "hans-hofmann" }),
```
javamonn: Is there a better way to mock the list of artists in a show? It seems like I'm only getting 2 artists in the rendered component, when there's many more in `ShowFixture`, which makes me think something akin to the "Hello, world!" stub data is being generated.

src/lib/Scenes/Show/index.tsx
```diff
@@ -63,32 +64,46 @@ export class Show extends React.Component<Props, State> {
...
-              case "artists":
-                return <Artists artists={data} />
-              case "shows":
-                return <Shows shows={data} />
-              default:
-                return null
-            }
+          ItemSeparatorComponent={this.renderItemSeparator}
+          renderItem={data => {
+            return <Box px={2}>{this.renderItem(data)}</Box>
```
javamonn: Our separator and container styling is consistent so far, so it makes sense to lift it up into the `FlatList`.

src/lib/Scenes/Show/Components/Artists/index.tsx
```diff
@@ -1,14 +0,0 @@
-import React from "react"
```
javamonn: I renamed this as we run into issues with Relay query fragment naming in the tests. For `Artists/index.tsx`, Danger expects a new test is added for `index-tests.tsx`, which creates issues if we want to start using `MockRelayRenderer` in these tests.

src/lib/Scenes/Show/Components/Artists/Artists.tsx
l2succes: I believe the optimisticResponse should be enough but maybe there's a good reason for using both

src/lib/Scenes/Show/Components/Artists/__tests__/Artists-tests.tsx
l2succes: As of now no, it's a limitation from relay-mock-resolver not mocking GraphQLList types properly when nested in a parent type.
We have to do that for connection types as well.

l2succes:
@javamonn Nice work on the mutation 👏

src/lib/Scenes/Show/Components/Artists/Components/ArtistListItem.tsx
```diff
@@ -0,0 +1,26 @@
...
+  isFollowed: boolean
+  onPress: () => void
+  isFollowedChanging: boolean
+}
+
+export const ArtistListItem: React.SFC<Props> = ({ name, isFollowed, onPress, isFollowedChanging }) => {
+  return (
+    <Flex justifyContent="space-between" alignItems="center" flexDirection="row">
+      <Sans size="3">{name}</Sans>
+      <Box width={112} height={32}>
```
l2succes: We should probably write a todo to remove that hardcoded width

javamonn: good idea, I added a `TODO` here 👍

PR #1215, opened 2018-11-05
WIP: Try updating palette

javamonn:
Does this pass CI? It appears to type check locally...


src/lib/Components/ArtworkGrids/Artwork.tsx
```diff
@@ -76,14 +76,14 @@ class Artwork extends React.Component<Props, any> {
         <Badges>
           {is_acquireable && (
             <Badge>
-              <Sans fontSize={8} lineHeight={8} style={{ paddingTop: 1 }} weight="medium" size="1">
+              <Sans fontSize="8px" lineHeight={8} style={{ paddingTop: 1 }} weight="medium" size="1">
```
javamonn: Without the `px` hardcoding here, the values appearing in the snapshot changed dramatically from `8` to something like `72`. I think  this was the result of the issue described in https://github.com/artsy/reaction/issues/1535.

src/lib/Components/Countdown/Ticker.tsx
```diff
@@ -31,7 +31,7 @@ const durationSections = (duration: Duration, labels: [string, string, string, s
 ]
 
 const LabeledTimeSection: React.SFC<TimeSectionProps> = ({ time, label, timeTextProps, labelTextProps }) => (
-  <Flex alignItems="center" justifyConent="center">
+  <Flex alignItems="center" justifyContent="center">
```
javamonn: One advantage of upgrading: this was not caught as a type error before 👏

orta: nice one

src/lib/Components/ArtworkGrids/Artwork.tsx
l2succes: Hmm, interesting... is it only affecting the snapshot or is the line height also affected in the UI?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 ### Master
 
 - Refactor auctions countdown timer, add fair countdown timer - javamonn
+- Updates @artsy/palette (2.21.1) - javamonn
```
damassi: Right on 👍

src/lib/Components/ArtworkGrids/Artwork.tsx
damassi: I think this is related to @mzikherman updating the badges to include an item that was outside of Palette's theme file, per @briansw's request.

javamonn: It affects both:

Before fixing `px`:
screen shot 2018-11-20 at 2 18 08 pm

After:
screen shot 2018-11-20 at 2 19 00 pm

My understanding of the issue is that it only affects spots where we're passing values into `fontSize` and `lineHeight` overrides like this, though this makes me concerned about us potentially doing this in spots we don't cover with snapshot tests - I'll see if I can find any other affected spots with a search.

damassi: @javamonn - would you mind raising this in #front-end?

CHANGELOG.md
zephraph: 🙌 

You're awesome @javamonn

l2succes:
merge on green

PR #1216, opened 2018-11-06
setup Fair detail view and header

l2succes:
- Extracts Location Map view into a sharable component
- Adds tests for Fair detail view

#skip_new_tests


src/lib/Scenes/Fair/Components/FairHeader/__tests__/FairHeader-tests.tsx
```diff
@@ -0,0 +1,27 @@
...
+          }
+        }
+      `,
+      mockResolvers: {
+        Fair: () => FairFixture,
+      },
+    })
+
+    const result = `<aropaqueimageview imageurl="https://d32dm0rphc51dk.cloudfront.net/o0Zrtm-CgwbV1syD_33DuA/square.jpg" aspectratio="1" width="750"><view></view><image source="[object Object]"></image></aropaqueimageview>`
+    expect(tree.html()).toEqual(result)
```
javamonn: Does a snapshot test work here? I don't think we can snapshot test the Enzyme wrapper directly, but would `expect(tree.html()).toMatchSnapshot()` work?

src/lib/Scenes/Fair/__tests__/index-tests.tsx
```diff
@@ -1,9 +1,31 @@
...
+          }
+        }
+      `}
+      mockResolvers={{
+        Fair: () => FairFixture,
+      }}
+    />
+  )
+
+  expect(tree.html()).toMatchSnapshot()
```
javamonn: 👏

src/lib/Scenes/Fair/Components/FairHeader/__tests__/FairHeader-tests.tsx
l2succes: Yeah that would work actually, I'm doing it with the other test. I was thinking about that as well. I'll convert it

PR #1217, opened 2018-11-07
Add "More Info" screen to show detail view

javamonn:
Adds a `MoreInfo` screen and reorganizes the existing `Scenes/Show` file layout so that `Scenes/Show/Show.tsx` becomes responsible for `NavigatorIOS` and transitioning between the active "route", and the existing Show detail view moves to `Scenes/Show/Screens/Detail.tsx`.

#skip_new_tests - missing tests are for `index` files, which re-export from tested components.


src/lib/Scenes/Show/Show.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export class Show extends React.Component<Props> {
+  navigator?: NavigatorIOS
+
+  handleMoreInformationPressed = () => {
+    if (!this.navigator) {
+      throw new Error("navigator is undefined")
+    }
+
+    this.navigator.push({
```
javamonn: We could avoid holding a ref and passing this method down as a prop by doing the navigation directly from within `Detail`, but getting Relay wired up seemed awkward to me - `Detail` would have to include the `MoreInfo_show` fragment. I'm open to changing this though.

src/lib/Scenes/Show/Screens/MoreInfo.tsx
```diff
@@ -0,0 +1,85 @@
...
+  state = {
+    sections: [],
+  }
+
+  componentDidMount() {
+    const { show: _show } = this.props
+    // FIXME: opening reception field does not exist, stubbed for now
+    const show = {
+      ..._show,
+      opening_reception: "Thursday, Oct. 18th, 6pm-8pm\n\nArtists in attendance. Please RSVP.",
```
anandaroop: 👍 Makes sense to stub this for now

We do actually have structured info for opening receptions — e.g. see [this query](https://metaphysics-production.artsy.net/?query=%7B%0A%20%20show(id%3A%20%22blue-lotus-gallery-the-way-we-were%22)%20%7B%0A%20%20%20%20id%0A%20%20%20%20events%20%7B%0A%20%20%20%20%20%20event_type%0A%20%20%20%20%20%20description%0A%20%20%20%20%20%20title%0A%20%20%20%20%20%20start_at%0A%20%20%20%20%20%20end_at%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D) — but it is not exposed as a distinct field in Metaphysics. Will add a ticket for that.

src/lib/Scenes/Show/Components/Artists/__tests__/Artists-tests.tsx
```diff
@@ -3,7 +3,7 @@ import { graphql } from "react-relay"
 
 import { MockRelayRenderer } from "../../../../../tests/MockRelayRenderer"
 import { renderUntil } from "../../../../../tests/renderUntil"
-import { ShowFixture } from "../../../__tests__/fixtures"
+import { ShowFixture } from "../../../__fixtures__"
```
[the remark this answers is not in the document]
javamonn: I moved the existing Show fixture we had: https://github.com/artsy/emission/pull/1217/files#diff-e67d92e496fd2ce40462b306e0b765ff

src/lib/Scenes/Show/Show.tsx
l2succes: Na i think it's ok to do it that way

src/lib/Scenes/Show/Screens/MoreInfo.tsx
anandaroop: To follow up — I clarified this with Jun.

Although the wireframe happens to only show **Opening Reception** here, this should in fact be a list of every event that is associated with the show. 

So we should actually iterate over e.g. the data in the `events` field of the [sample query](https://metaphysics-production.artsy.net/?query=%7B%0A%20%20show(id%3A%20%22blue-lotus-gallery-the-way-we-were%22)%20%7B%0A%20%20%20%20id%0A%20%20%20%20events%20%7B%0A%20%20%20%20%20%20event_type%0A%20%20%20%20%20%20description%0A%20%20%20%20%20%20title%0A%20%20%20%20%20%20start_at%0A%20%20%20%20%20%20end_at%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D). No further follow-up is needed in Metaphysics to enable this (until/unless we have to deal with the stub show version of this).

l2succes: Got it, thanks for the follow-up explanation @anandaroop. I'll create a ticket to track this task separately.

PR #1220, opened 2018-11-07
Cleanup Jest output

javamonn:
- Restores [`console.error` monkey-patching](https://github.com/artsy/emission/commit/2a9324086d51bce4f8c989f83b7ff5ebe05005fc#diff-0b93f41d9b1048bf2a2e50b4036a898aL11) in order to silence several errors introduced by JSDOM rendering added in #1209. These are all warnings stemming from JSDOM emulating a DOM environment and us serializing RN components, props, etc to it.
- Fixes a couple of legit duplicate key warnings caused by incorrectly mocked lists
- Upgrades `relay-mock-network-layer` to 2.0.0, removing our need to patch


src/lib/Scenes/Show/Components/Artists/__tests__/Artists-tests.tsx
```diff
@@ -28,8 +28,10 @@ const renderTree = () =>
         }
       `}
       mockResolvers={{
-        Show: () => ShowFixture,
-        Artist: () => ({ name: "Hans Hofmann", is_followed: false, id: "hans-hofmann" }),
+        Show: () => ({
+          ...ShowFixture,
+          artists: () => ShowFixture.show.artists,
```
javamonn: TIL about [`MockList`](https://www.apollographql.com/docs/graphql-tools/mocking.html#MockList) while trying to figure out the right way to do this.

src/setupJest.ts
```diff
@@ -31,6 +31,23 @@ function mockedModule(path: string, mockModuleName: string) {
...
+console.error = (message?: any) => {
+  if (
+    typeof message === "string" &&
+    (message.includes("is using uppercase HTML. Always use lowercase HTML tags in React.") ||
+      /Warning: React does not recognize the `\w+` prop on a DOM element\./.test(message) ||
+      /Warning: The tag <\w+> is unrecognized in this browser\./.test(message) ||
+      /Warning: Unknown event handler property `\w+`\./.test(message) ||
+      /Warning: Received `\w+` for a non-boolean attribute `\w+`\./.test(message))
+  ) {
+    // NOOP
```
orta: probably a good call to say when we'll remove this ^

orta:
👍

PR #1221, opened 2018-11-07
Tighten the native dependencies for Emission

orta:
Tightens up the dependencies in the Emission podspec to the exact versions of the JS that's emitted.


orta:
Oh yeah, the reason for this is that a release of Emission just went out with a hard dependency on MapBox - but Eigen didn't know about it. I also spotted that tipis-stripe is attached to a commit in Eigen, and that commit (might) not be the exact tag that we're using in Emission so the bridge could be out of sync.

screen shot 2018-11-07 at 2 15 23 pm

PR #1223, opened 2018-11-13
Add HoursCollapsible component to Fair detail

javamonn:
Implements [LD-27](https://artsyproduct.atlassian.net/browse/LD-27).

The transition between collapsed and expanded is not animated currently. I think once designs are settled and palette is upgraded in emission, porting an implementation of https://github.com/artsy/palette/pull/126 to iOS is the best path forward.

This also currently causes a jump in the position of the `FlatList` on expand / collapse. I'm not sure, but I think this is due to layout issues relating to the disappearing map image.

screen shot 2018-11-13 at 11 59 24 am


src/lib/Scenes/Fair/Components/HoursCollapsible.tsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  render() {
+    const { isExpanded } = this.state
+    const { hours } = this.props
+    return (
+      <Box>
+        <Flex justifyContent="space-between" alignItems="center" flexDirection="row">
+          <Sans size="4">Hours</Sans>
+          <Box width={112}>
```
l2succes: why are we using a hardcoded width here?

javamonn: The way this `Button` works as far as I can tell is that it fills it parent, it's a bit awkward to create a button with horizontal padding. I'll rework this to be wrapped in a Flex constrained by margin so we can remove the hardcoded width.

l2succes:
Looks good, just had a minor comment. 

@anandaroop Should we ever fallback on using our structured data for hours on fairs? Because if we do, we'll have to account for that as well

src/lib/Scenes/Fair/Components/HoursCollapsible.tsx
```diff
@@ -0,0 +1,39 @@
...
+    this.setState({ isExpanded: !this.state.isExpanded })
+  }
+
+  render() {
+    const { isExpanded } = this.state
+    const { hours } = this.props
+    return (
+      <Box>
+        <Flex justifyContent="space-between" alignItems="center" flexDirection="row">
+          <Sans size="4">Hours</Sans>
```
anandaroop: Noting that this field IRL often [contains Markdown](https://user-images.githubusercontent.com/140521/48497302-dd499b00-e801-11e8-9339-a4f7a50d18e8.png) — do we have a pattern for rendering Markdown in RN views?

anandaroop: @l2succes good q — I'm not sure if we have structured data for fair hours — will investigate

orta: We do have a markdown renderer in here somewhere, yep

javamonn: Nice catch, I pulled the markdown renderer up out of [`Bidding/Components`](https://github.com/artsy/emission/pull/1223/commits/3804c06acde7435f165e1a2f45ddc9a9c19b03c0) and am using it to render this text now.

javamonn:
This PR introduced the component scaffold, but since designs have settled a bit, still TBD:

- [ ] Styling per the latest seen in the mocks: https://app.zeplin.io/project/5bef4c7dae2b6a3eb6a75f4a/screen/5bf32e8daf23577ecb31cebe
- [ ] Supporting structure hours data sometimes available on `show(id: ...) { location { day_schedules { ... } } }`, falling back to use text / markdown
- [ ] Adding this component and wiring up on `Show/Detail`

src/lib/Scenes/Show/Screens/Detail.tsx
```diff
@@ -121,6 +153,10 @@ export const DetailContainer = createFragmentContainer(
         city
         state
         postal_code
+        displayDaySchedules {
```
l2succes: We abstracted where the data for this field is coming from in metaphysics right?
(Structured / unstructured data)

l2succes:
Nice one man! There's some merge conflicts but let's merge on green

src/lib/Scenes/Show/Screens/Detail.tsx
javamonn: @anandaroop added a field to `Location` for unstructured text data, I'll create a follow up ticket to make an abstracted `hoursDisplay` field that pulls from either this field or the unstructured `Fair.hours` or `Location.day_schedule_text`.

PR #1226, opened 2018-11-20
Refactors ARNotificationManager to implicitly call all JS methods on main queue

ashfurrow:
This is a follow up from [this discussion](https://github.com/artsy/emission/pull/1171/files#r203167486). It's been on my TODO list for months and I finally did it.


ashfurrow:
No worries!

PR #1228, opened 2018-11-21
Adds mapbox api key to cocoapods

kierangillen:
This API key will need to be set with an official key for production.


kierangillen:
I’m seeing this error from the test ```This PR includes changes to the Emission Pod's native code but does not have a `package.json` change for the update to the `"native-code-version"`. If this is fine, add #native_no_changes to your PR message.``` 

Do I need to update the `package.json` file?

src/lib/Scenes/Show/Components/Artists/Components/ArtistListItem.tsx
```diff
@@ -1,20 +1,66 @@
...
+  nationality?: string
+  birthday?: string
+  deathday?: string
   isFollowed: boolean
   onPress: () => void
   isFollowedChanging: boolean
+  url: string
+}
+
+const returnTombstoneText = (nationality, birthday, deathday) => {
```
l2succes: We could test this logic a bit

l2succes:
Yeah we need to update `native-code-version` and `version` in package.json

@ashfurrow can we just bump the # from 9 to 10?

Example/Emission/AppDelegate.m
```diff
@@ -119,18 +119,19 @@ - (void)setupEmissionWithUserID:(NSString *)userID accessToken:(NSString *)acces
...
 
   EmissionKeys *keys = [[EmissionKeys alloc] init];
-  
+
   AREmissionConfiguration *config = [[AREmissionConfiguration alloc] initWithUserID:userID
                                                                 authenticationToken:accessToken
                                                                           sentryDSN:nil
                                                                stripePublishableKey:[keys stripePublishableKey]
                                                                    googleMapsAPIKey:nil
+                                                                 mapBoxAPIClientKey:[keys mapBoxAPIClientKey]
```
ashfurrow: Outside the scope of this PR, I'd like us to move to passing in `keys` directly, instead of pulling everything out. If Emission defines a protocol that the object needs to conform to (one that matches `ArtsyKeys` in Eigen), then it should be straightforward.

src/lib/Scenes/Show/Screens/Detail.tsx
```diff
@@ -66,10 +66,10 @@ export class Detail extends React.Component<Props, State> {
...
   renderItem = ({ item: { data, type } }) => {
     switch (type) {
-      case "location":
-        return <LocationMap {...data} />
-      case "artworks":
-        return <Artworks show={data} />
+      // case "location":
+      //   return <LocationMap {...data} />
+      // case "artworks":
+      //   return <Artworks show={data} />
```
ashfurrow: This is intended?

kierangillen: just fixed this

Example/Emission/AppDelegate.m
ashfurrow: emission-nebula will need _some_ env car defined for the build to work (even if the value isn’t valid).

kierangillen: https://travis-ci.org/artsy/emission-nebula/settings Added, just to be clear in cocoapods we have it beginning with a lowercase `m`, so I've added `mapBoxAPIClientKey`

ashfurrow: Thanks! If you look at the [Podfile](https://github.com/artsy/emission/blob/bb48bcfa3271e8ff10d67828e2a673215aef9530/Example/Podfile#L68-L70) that has the keys defined, the keys _always_ start with a capital, but get generated into a lower-case property during `pod install`. So you'll need `MapBoxAPIClientKey`, matching your change in the `Podfile`. Let me know if I can clarify 👍

kierangillen: Thanks for the clarification, I added it with the uppercase M :)

l2succes: @ashfurrow I don't see env vars defined for other cocoapods keys in emission-nebula TravisCI settings. Am I looking in the wrong place.

orta: Looks like they're embedded in the `travis.yml`: https://github.com/artsy/emission-nebula/blob/master/.travis.yml#L25-L29

Both should work together

PR #1230, opened 2018-11-21
More shows

l2succes:
screen shot 2018-11-26 at 1 17 48 pm

#skip_new_tests #native_no_changes


Example/Emission/AppDelegate.m
```diff
@@ -254,6 +255,10 @@ - (UIViewController *)viewControllerForRoute:(NSString *)route;
     NSString *artistID = [[route componentsSeparatedByString:@"/"] lastObject];
     viewController = [[ARArtistComponentViewController alloc] initWithArtistID:artistID];
 
+  } else if ([route hasPrefix:@"/show/"] || [route hasPrefix:@"show/"]) {
```
javamonn: This displays our new show screen instead of the web view, right?

src/lib/Scenes/Show/Components/Shows/Components/ShowItem.tsx
```diff
@@ -1,127 +1,79 @@
...
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
 import React from "react"
-import { Text } from "react-native"
+import { Dimensions, TouchableOpacity } from "react-native"
 import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components/native"
 
-const DEFAULT_ARTWORK_URL = "https://d32dm0rphc51dk.cloudfront.net/ADYdY8P1kg9vKA9ffrb4pg/larger.jpg"
-
+const { width: windowWidth } = Dimensions.get("window")
```
javamonn: This width could change, right? We might want to setup a listener to handle device rotations: https://facebook.github.io/react-native/docs/dimensions#addeventlistener

orta: Yeah, you can listen for onLayout's too I think - that's the pattern elsewhere

Example/Emission/AppDelegate.m
orta: In the Emission app -  it's a blank screen that says the path, so "/show/whatever" (in Eigen it would show the internal web view)

l2succes: Well this hooks up our ShowVC view in emission so if you tap on a show it'll render it instead of the blank screen. Will have to register ShowVC to show urls in Eigen next

src/lib/Scenes/Show/Components/Shows/Components/ShowItem.tsx
l2succes: Yes, that would only be relevant for iPad landscape and portrait orientations and we should just pick a card width that works for both in this case

PR #1233, opened 2018-11-27
Update stylelint

orta:
Updates all of the stylelint dependencies and fixes the 2 issues we had - sibling PR to https://github.com/artsy/reaction/pull/1594

#trivial


damassi:
👍

PR #1234, opened 2018-11-27
Updates pin size

kierangillen:
Updates pin size on show location map and zooms in by a fraction to be able to read street names more reliably.


src/lib/Components/HoursCollapsible.tsx
```diff
@@ -0,0 +1,91 @@
+import { Box, Flex, Sans, Serif } from "@artsy/palette"
```
l2succes: @kierangillen did you mean to add that to this PR?

kierangillen: Thanks, no, just removed it.

PR #1236, opened 2018-11-29
Pass through partnerType to LocationMap on Fair

javamonn:
Fixes a crash triggered when accessing the Fair screen

#trivial


l2succes:
merge on green

PR #1237, opened 2018-11-29
Fair header & booth sections

l2succes:
kapture 2018-12-06 at 16 50 25

- Implements new header style (White text on dark overlay)
- Adds fair booth section and infinite scroll to load more

#skip_new_tests


l2succes:
👍

PR #1238, opened 2018-11-29
Show  all artists

kierangillen:
- Adds styled AllArtist page to Shows
- Adds styling for gray bordered InvertedButton by passing grayBorder prop to InvertedButton

screen shot 2018-11-29 at 2 19 08 pm

#skip_new_tests


src/__generated__/AllArtistsFollowArtistMutation.graphql.ts
```diff
@@ -0,0 +1,118 @@
+/* tslint:disable */
+
+import { ConcreteRequest } from "relay-runtime";
```
l2succes: @kierangillen Looks like the sorting is happening again.. we should find a way to exclude those or what's causing that on your machine

PR #1243, opened 2018-12-11
Fix the artwork favourite pagination

orta:
Re: DISCO-553

We were being overly conservative with the `shouldUpdate` it would only account for when the layout changes (and the number of columns) but not when the artworks change, which is kinda the aim of the component 👍.


src/lib/Components/ArtworkGrids/GenericGrid.tsx
```diff
@@ -66,9 +66,9 @@ class GenericArtworksGrid extends React.Component<Props, State> {
     }
   }
 
-  shouldComponentUpdate(_, nextState) {
-    if (this.state.sectionCount === nextState.sectionCount) {
-      // if there's a change in columns, we'll need to re-render
+  shouldComponentUpdate(nextProps: Props, nextState: State) {
+    // if there's a change in columns, we'll need to re-render
+    if (this.props.artworks === nextProps.artworks && this.state.sectionCount === nextState.sectionCount) {
```
javamonn: This makes sense - the grid should re-render when artworks change. We briefly discussed offline about potential performance implications, particularly around `InfiniteScrollGrid` and Relay pagination containers, and whether new pages would trigger re-renders of grids backing earlier pages.

javamonn:
merge on green

src/lib/Components/ArtworkGrids/GenericGrid.tsx
orta: Looks like the InfiniteScrollGrid doesn't use the GenericGrid at all too

PR #1249, opened 2018-12-17
Link to "All works" from Fair booth, works

javamonn:
Implements [LD-93](https://artsyproduct.atlassian.net/browse/LD-93).

#skip_new_tests


src/lib/Scenes/Fair/Components/FairBooth/FairBooth.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { createFragmentContainer, graphql } from "react-relay"
+import { FairBoothHeader } from "./Components/FairBoothHeader"
+
+interface Props {
+  show: FairBooth_show
+  onViewWorksPressed: () => void
+}
+
+export const FairBoothContainer = createFragmentContainer<Props>(
+  props => {
```
l2succes: Nice use of anonymous component. I think we should use this pattern for other relay containers.

PR #1250, opened 2018-12-18
Add FairBooth screen

javamonn:
Implements [LD-51](https://artsyproduct.atlassian.net/browse/LD-51).

- Adds the [Fair Booth](https://app.zeplin.io/project/5bef4c7dae2b6a3eb6a75f4a/screen/5c185e771c6890243d6cc523) screen, linked to from the "View all works" button that appears in fair booth previews on the Fair Detail
- From the Fair Booth screen, links to the Show Artworks screen and Show Artists screen

#skip_new_tests - missing tests are for a fixture file and an index file re-exporting a component with coverage.


src/lib/Scenes/Fair/Fair.tsx
```diff
@@ -17,40 +20,48 @@ interface Props extends ViewProperties {
...
-      component: FairArtistsScreen,
-      title: "",
-      passProps: this.props,
-    })
+  handleViewFairBoothPressed = props => {
+    this.navigate({ component: FairBoothScreen, props })
+  }
+
+  handleViewFairBoothArtistsPressed = props => {
+    this.navigate({ component: ShowArtistsScreen, props })
```
javamonn: This props passing is kind of awkward but necessary with the current layout as far as I can tell - We let the root view handle navigation, but the root view is missing the data required to show screens of other types. In this case, `handleViewFairBoothPressed`, `handleViewFairBoothArtistsPressed`, and `handeleViewFairBoothArtworksPressed` all receive fragments on `Show` rather than `Fair` - at what point would it make sense to split this out into a separate query renderer?

l2succes: I would argue it makes sense to have the `FairBoothScreen` and `ShowArtworksScreen` into their own QueryRenderers for the reason you mentioned but also because currently we're overfetching data when rendering the base Fair screen since it doesn't need artists or exhibitors yet. But I haven't thought through all the repercussions of switching to nested QueryRenderers so let's keep going that route for now

PR #1251, opened 2018-12-19
Add Contextual Fair Data for Artists and Exhibitors to Fairs Header

ashleyjelks:
screen shot 2018-12-20 at 10 00 46 am

[See LD-62](https://artsyproduct.atlassian.net/browse/LD-62)
#skip_new_tests"


src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -49,10 +49,61 @@ const CountdownContainer = styled.View`
...
+  font-weight: bold;
+`
 export class FairHeader extends React.Component<Props> {
+  getArtists(artistCounts, artistsNames) {
+    const { artists } = artistCounts
+
+    return (
+      <Flex flexDirection="row" flexWrap="wrap">
+        <Sans size="3">Works by </Sans>
+        <Bold size="3">{artistsNames.edges[0].node.name + ", "}</Bold>
```
l2succes: this would fail if `edges.length == 0` for some reason. I would store `edges` in a variable and check before getting data

src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -49,10 +49,61 @@ const CountdownContainer = styled.View`
...
+    const shortExhibitorList = [<Sans size="3">{"From "}</Sans>]
+    exhibitorNamesToDisplay.forEach((exhibitor, index) => {
+      if (index === exhibitorNamesToDisplay.length - 1) {
+        shortExhibitorList.push(<Sans size="3">{" and "}</Sans>, <Bold size="3">{exhibitor}</Bold>)
+      } else {
+        shortExhibitorList.push(<Bold size="3">{exhibitor + ", "}</Bold>)
+      }
+    })
+    return shortExhibitorList
+  }
```
l2succes: add a new line between your methods

src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -49,10 +49,74 @@ const CountdownContainer = styled.View`
...
+
+    exhibitorNamesToDisplay.forEach((exhibitor, index) => {
+      if (index === exhibitorNamesToDisplay.length - 1) {
+        shortExhibitorList.push(<Sans size="3">{" and "}</Sans>, <Bold size="3">{exhibitor}</Bold>)
+      } else {
+        shortExhibitorList.push(<Bold size="3">{exhibitor + ", "}</Bold>)
+      }
+    })
+
+    return shortExhibitorList
```
l2succes: Sorry for missing this first time around but you could write this method more succinctly by handling a "short" list of exhibitors at the same time as the long one.

Taking this one step further you could actually abstract all that logic in one component for both the list of artists and the list of exhibitors as they're essentially the same like I did in the `ArtistNamesList` for the Show header view
https://github.com/artsy/emission/blob/master/src/lib/Scenes/Show/Components/ShowHeader/Components/ArtistNamesList.tsx

src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -49,11 +48,61 @@ const CountdownContainer = styled.View`
...
+    artistsCount = !lastArtistName ? artistsCount : artistsCount - 1
+    exhibitorsCount = !firstExhibitorName ? exhibitorsCount : exhibitorsCount - 1
+    exhibitorsCount = !lastExhibitorName ? exhibitorsCount : exhibitorsCount - 1
+
+    return (
+      <>
+        {(!firstArtistName && !lastArtistName) || !artistsCount ? null : (
+          <Flex flexDirection="row" flexWrap="wrap">
+            <Sans size="3">Works by </Sans>
+            <Bold size="3">{!firstArtistName ? null : firstArtistName + ", "}</Bold>
```
damassi: Rather than using `font-weight: bold` (which would then go outside of the design system spec) this should be `Sans weight='medium'` (which is the "bold" of Sans). If you want to keep the `<Bold>` component tho can also do something like this:

```js
const Bold = styled(Sans).attrs({ weight: 'medium' })``
```
Where the `attrs` method allows one to pass specific props on through.

src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -49,11 +48,61 @@ const CountdownContainer = styled.View`
...
+            <Bold size="3">{!lastArtistName ? null : lastArtistName + ", "}</Bold>
+            <Sans size="3">{!firstArtistName && !lastArtistName ? null : "and "}</Sans>
+            <Bold size="3">{artistsCount + " others."}</Bold>
+          </Flex>
+        )}
+        {(!firstExhibitorName && !lastExhibitorName) || !exhibitorsCount ? null : (
+          <Flex flexDirection="row" flexWrap="wrap">
+            <Sans size="3">{"From "}</Sans>
```
damassi: No need for brackets around the `"From "`

src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -76,7 +125,16 @@ export class FairHeader extends React.Component<Props> {
...
-
+        <Box mx={2}>
+          {this.getContextualDetails(
+            artistCount,
+            firstArtist,
+            lastArtist,
+            exhibitorCount,
+            firstExhibitor,
+            lastExhibitor
+          )}
```
damassi: Rather than extracting data from props and passing it into a method that returns something right back, best to just pull the data from props from within the method itself. So the render method becomes super clean: 

```jsx
<Box mx={2}>
  {this.getContextualDetails()}
</Box>
```
And then in the method itself, grab the data since it can be easily accessed via `this.props`. This reduces complexity: 
```js
getContextualData() {
  const { keep all data colocated where its needed } = this.props
  ...
}

File: src/lib/Scenes/Fair/Components/FairHeader/index.tsx [hunk 7]
```diff
@@ -49,61 +48,81 @@ const CountdownContainer = styled.View`
...
+            <Sans size="3">Works by </Sans>
+            {firstArtistName && (
+              <Sans weight="medium" size="3">
+                {firstArtistName + ", "}
+              </Sans>
+            )}
+            {lastArtistName && (
+              <Sans weight="medium" size="3">
+                {lastArtistName + ", "}
+              </Sans>
```
damassi: 👍

src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -127,10 +146,7 @@ export class FairHeader extends React.Component<Props> {
           </CountdownContainer>
         </BackgroundImage>
         <Spacer m={1} />
-        <Box mx={3}>
-          {this.getArtists(counts, artists_names)}
-          {this.getExhibitors(exhibitors_grouped_by_name)}
-        </Box>
+        <Box mx={2}>{this.getContextualDetails()}</Box>
```
damassi: 👍

damassi:
@ashleyjelks - Looks great 👍

PR #1252, opened 2018-12-20
Map & City container view

l2succes:
Sets up native bottom sheet container with React Native Map and City Components.

Todo (in a separate PR)
- [ ] Set up user location service and nearest city matcher
- [ ] Add event emitter for applying filters across Map & City view
- [ ] Slide top bar (filters) up when bottom sheet is in full screen

Design spec: https://zpl.io/VYGRAwJ

#skip_new_tests

### DIFF
--- a/Example/Emission/Placeholder.swift
+++ b/Example/Emission/Placeholder.swift
@@ -1,11 +1,3 @@
-//
-//  Placeholder.swift
-//  Emission
-//
-//  Created by Luc Succes on 12/19/18.
-//  Copyright © 2018 CocoaPods. All rights reserved.
-//
-
 import Foundation
 
 // Dummy swift file


Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -0,0 +1,38 @@
...
+
+@interface ARMapContainerViewController ()
+
+@property (nonatomic, readwrite) BottomSheetViewController *bottomSheetVC;
+@property (nonatomic, readwrite) ARMapComponentViewController *mapVC;
+@property (nonatomic, readwrite) ARCityComponentViewController *cityVC;
+
+@end
+
+@implementation ARMapContainerViewController
```
l2succes: This class will also be responsible for inter City / Map view communication such as applying filters across both and changing cities / location

src/lib/AppRegistry.tsx
```diff
@@ -171,3 +173,5 @@ AppRegistry.registerComponent("Favorites", () => FavoritesScene)
 AppRegistry.registerComponent("BidFlow", () => BidderFlow)
 AppRegistry.registerComponent("Fair", () => Fair)
 AppRegistry.registerComponent("Show", () => Show)
+AppRegistry.registerComponent("Map", () => GlobalMap)
```
l2succes: Will switch to MapRenderer when ready

Pod/Classes/ViewControllers/ARMapComponentViewController.m
```diff
@@ -0,0 +1,15 @@
+#import "ARMapComponentViewController.h"
+
+@implementation ARMapComponentViewController
+
+- (instancetype)init
```
l2succes: Will need to use `appProperties` (https://facebook.github.io/react-native/docs/communication-ios) to set location onto map from native User location module

src/lib/Scenes/Map/Components/FiltersBar.tsx
```diff
@@ -0,0 +1,104 @@
...
+    })
+  }
+
+  render() {
+    return (
+      <>
+        <Box pt={4} pb={2} px={3}>
+          <Serif size="8">Hong Kong</Serif>
+        </Box>
+        <Tabs>
```
l2succes: Should be a horizontal scrollview

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+interface Props {
+  location: LocationMap_location
+  partnerType: PartnerType
+  partnerName?: string
+}
+
+export class GlobalMap extends React.Component<Props> {
+  render() {
+    const { lat, lng } = { lat: 40.770424, lng: -73.981233 }
```
l2succes: Get location from props passed in from native city service

Example/Emission.xcodeproj/project.pbxproj
```diff
@@ -799,13 +804,15 @@
 				PRODUCT_BUNDLE_IDENTIFIER = net.artsy.Emission;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE_SPECIFIER = "match AppStore net.artsy.Emission";
+				SWIFT_VERSION = 4.2;
```
l2succes: This links swift libs in Emission but I think that's ok since we already are linking swift in Eigen

src/lib/relay/fetchQuery.ts
```diff
@@ -26,6 +26,7 @@ function _fetchQuery(queryID: string, variables: object, isQuery: boolean) {
   let body
   if (__DEV__) {
     body = { query: require("../../__generated__/complete.queryMap.json")[queryID], variables }
+    console.log(body)
```
l2succes: This turns out to be useful to debug graphQL queries when it doesn't work in emission for some reason

Example/Emission/Placeholder.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  Placeholder.swift
+//  Emission
+//
+//  Created by Luc Succes on 12/19/18.
+//  Copyright © 2018 CocoaPods. All rights reserved.
+//
```
alloy: Always remove these headers please. We have SCM for blaming and the copyright lines are harder to maintain then the single notice in the LICENSE file.

src/lib/relay/fetchQuery.ts
alloy: What do you mean by “when it doesn't work in emission”?

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -0,0 +1,60 @@
...
+      </Flex>
+    )
+  }
+}
+
+export const GlobalMapContainer = createFragmentContainer(
+  GlobalMap,
+  graphql`
+    fragment GlobalMap_viewer on Viewer
+      @argumentDefinitions(near: { type: "Near", defaultValue: { lat: 22.3964, lng: 114.1095 } }) {
```
alloy: This is Hong Kong, right? Why is that a default / can you add a comment for that?

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -0,0 +1,60 @@
...
+        />
+      </Flex>
+    )
+  }
+}
+
+export const GlobalMapContainer = createFragmentContainer(
+  GlobalMap,
+  graphql`
+    fragment GlobalMap_viewer on Viewer
```
alloy: Just for my understanding, this is on `Viewer` so that the `near` argument can be kept local to this module, right?

src/lib/Scenes/Map/GlobalMap.tsx
l2succes: Yes, I set it to Hong Kong by default for now as that's the city where we're launching the feature but we should remove the default value when the location service is ready

src/lib/Scenes/Map/GlobalMap.tsx
l2succes: Yes that's right, but we may decide to change that as the coordinate values may be pulled from the `GeoLocation` module (https://facebook.github.io/react-native/docs/geolocation) or we may end up writing a custom native module that is aware of our city usecase

src/lib/Scenes/Map/GlobalMap.tsx
anandaroop: Can you elaborate on the "location service" you mentioned here? Is that what will map a devices's currents coords to an appropriate LD city?

Also, I'm wondering if you want to fetch shows the way shown below, or if you want to use a named city entry point, [such as this](https://metaphysics-production.artsy.net/?query=%7B%0A%20%20city(slug%3A%20%22london-united-kingdom%22)%20%7B%0A%20%20%20%20slug%0A%20%20%20%20name%0A%20%20%20%20shows(discoverable%3A%20true%2C%20first%3A%205)%20%7B%0A%20%20%20%20%20%20edges%20%7B%0A%20%20%20%20%20%20%20%20node%20%7B%0A%20%20%20%20%20%20%20%20%20%20id%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D)? That root query is intended to be a convenient entry point for LD-city related data (and assumes the device's current position has been mapped to a city slug). If that won't be useful we should discuss.

l2succes: So what I meant by location service is really a client side native library that abstracts getting the user device's location and provides the initial coordinates the map should be centered on. Basically the following pseudo logic

```
  // Logic for whether to show a city or cities list
  //
  // 1. Get user location (lat, lng) from native code
  // 2. do a lookup to get nearest city up to a certain threshold
  // 3. if within threshold select nearest city
  // 4. else show cities list
```

Good thing you mention that though @anandaroop because I forgot with had a `city` root field in MP was gonna bring that up. What would be great is if we also support querying for a city using coordinates such as
```
city(near: $near) {
}
```
which would return null if there is no city nearby and that way we just show the list of cities instead

anandaroop: 👍 So in that case, the logic to select an appropriate LD city happens server-side over in MP, rather than client side yeah? And the pseudo-logic above becomes something like:

```
1. Get user location (lat, lng) from native code
2. Request `city(near: lat, lng)` from MP
3. If the response is null, or something representing "no nearest city", render the picker
4. Else render the returned data
```

(MP specifics are debatable, but the flow is something like the above?)

l2succes: Yes, exactly the pseudo logic you described

Pod/Classes/BottomSheetComponent/BottomSheetViewController.swift
```diff
@@ -0,0 +1,1521 @@
+//
+//  BottomSheetViewController.swift
+//  Emission
+//
+//  Code reused from https://github.com/52inc/Pulley
```
ashfurrow: Is it possible to use the Pulley library as-is from their repo? It has a podfile. If not, we should document why somewhere.

l2succes: I looked into it and it didn't work out for a few reasons
1. I wasn't able to import the swift interfaces from Objective-C using `#import <Pulley/Pulley-Swift.h>`
2. There's some tweaks we will need to do for our use case that will be less straightforward to it with Pulley

l2succes:
Alright I believe this is now ready to be merged.

Based on @ashfurrow and @orta suggestions. I ended up creating a fork of Pulley to fix the Objective-C compatibility issue and we'll be able to apply needed changes to the maintained version. So no need to add swift code to emission for now.

l2succes:
👍 definitely aware of our RFC process for new dependencies, I'll follow the process on the next one. Thanks for helping getting this one in a mergeable state @ashfurrow.

alloy:
@l2succes Thanks for resolving the objc vs swift issue! I wonder, though, if we should make sure forked code exists on the Artsy org for long term stability of our projects? /cc @ashfurrow

ashfurrow:
@alloy I told Luc he could use his personal account, which is what I'd have done, following the same practice as creating libraries on personal repos ([example here](https://github.com/ashfurrow/forgeries), [and here](https://github.com/orta/ARAnalytics)). But I see your point about longterm stability. 

@orta I'm curious about your thoughts on this. We could even leverage Peril to detect when we're pointing to a fork of someone no longer in the GitHub org, but that might be overkill.

orta:
The distinction that's super important here is OSS vs Private, a fork on a private repo which is used would get deleted when someone leaves vs an OSS one which is very likely to stick around (because no-one really deletes their forks).

We could do something like the eigen-artefacts for Emission, if we made it a private repo then it could contain fonts and we could simplify people's local setup. I'm not really sure if people are good at updating that submodule etc but native changes are generally rare

ashfurrow:
Any Emission native dependencies will end up in `eigen-artefacts` anyway, right?

alloy:
Good point re the distinction @orta 👍 

And yeah @ashfurrow, indeed we do already pull it into Eigen's artefacts, which I forgot about. Nonetheless, perhaps something to consider in a general sense, as we don't store artefacts for all of Artsy's projects.

PR #1253, opened 2019-01-02
Update artwork grid item text spacing

javamonn:
Implements [LD-104](https://artsyproduct.atlassian.net/browse/LD-104)

Note that this updates the styling of the Artwork grid item globally in Emission - this is intended.

An example of the tightened grid item can be seen here: https://app.zeplin.io/project/5bef4c7dae2b6a3eb6a75f4a/screen/5c1ae7411253daaef2213dcb

Before:
screen shot 2019-01-04 at 3 02 07 pm

After:
screen shot 2019-01-02 at 2 45 51 pm


src/lib/Components/ArtworkGrids/Artwork.tsx
```diff
@@ -164,10 +164,12 @@ export class Artwork extends React.Component<Props, any> {
 
 const styles = StyleSheet.create({
   image: {
-    marginBottom: 10,
+    marginBottom: 8,
   },
   text: {
-    fontSize: 12,
+    fontSize: 14,
+    lineHeight: 18,
```
javamonn: The text styling seen in the [designs](https://app.zeplin.io/project/5bef4c7dae2b6a3eb6a75f4a/screen/5c1ae7411253daaef2213dcb) isn't available in palette as far as I can tell, so we hardcode sizes and margins.

javamonn: `ArtworkGrids/Artwork`  isn't a new component, it's just the common grid item that gets rendered by `GenericGrid` and `InfiniteScrollGrid`. We're using it in LD and it's also used in other spots in Emission, so these styling tweaks are global.

javamonn:
@ashfurrow Great point on this causing visual inconsistency in the app. My understanding of this new artwork grid design is that it's brand-wide, e.g. we've already switched over in the web: https://www.artsy.net/collect, though the text there looks even tighter than the change here - I'll double check the margins here. 

It seems ideal to take this opportunity and also propagate this grid item change to Eigen as well, I'll coordinate the PRs so that the changes are timed together.

ashfurrow:
@javamonn great, sounds good. Even if we can ticket the work for Eigen, that'd be helpful. Here are the two places I know of offhand that have similar visuals – the code is kind of all over the place (because of how UIKit and UICollectionView's APIs are structured) so please grab me if I can help. 

## The Auction View

img_3496

Code is here:

- https://github.com/artsy/eigen/blob/master/Artsy/View_Controllers/Embedded/Sale_Artworks/ARSaleArtworkItemMasonryModule.m
- https://github.com/artsy/eigen/blob/master/Artsy/Views/Collection_View_Cells/ARSaleArtworkMasonryCollectionViewCell.m#L149-L153

## The Artwork View

In "Related Works", though for artworks in sales, we use the code above.

img_3497

Code is here:

- https://github.com/artsy/eigen/blob/master/Artsy/View_Controllers/Embedded/Artworks/Modules/ARArtworkMasonryModule.m
- https://github.com/artsy/eigen/blob/master/Artsy/Views/Collection_View_Cells/ARArtworkWithMetadataThumbnailCell.m#L29-L36

javamonn:
@ashfurrow Opened a PR on Eigen to address the related artworks grid item styling: https://github.com/artsy/eigen/pull/2752. I'm going to create a ticket to track the work needed for the Auction view, as I'm unsure what the metadata styling should look like with the lot number and paddle.

ashfurrow:
@javamonn sounds great, thanks!

orta:
Inreresting, this failed when merged - what's with the CI not giving accurate results?

orta:
The only difference between this and a normal Eigen PR is that it's running from a fork

javamonn:
This one in particular looks like it could be transient and due to Circle's outage yesterday (I'll retry it) - Eigen PR here is one that failed the Eigen build: https://github.com/artsy/eigen/pull/2754

PR #1254, opened 2019-01-02
Update Show MoreInfo screen with events

javamonn:
Implements [LD-43](https://artsyproduct.atlassian.net/browse/LD-43)

[Design](https://app.zeplin.io/project/5bef4c7dae2b6a3eb6a75f4a/screen/5c1ae7415ff90824363b8f42)

#skip_new_tests - missing tests are for an updated fixture data file


src/lib/Scenes/Show/Components/EventSection.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { EventSection_event } from "__generated__/EventSection_event.graphql"
+import moment from "moment"
+import React from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+
+interface Props {
+  event: EventSection_event
+}
+
+const formatDate = startAt => moment(startAt).format("dddd, MMMM D")
```
javamonn: The designs seem to only take into account `start_at` and `end_at` within a single day, though it seems like the data could come down with an event starting and ending on different days. Not sure if / how we want to handle this.

anandaroop: Great point. Looking at staging I see that a significant fraction of `PartnerShowEvents` have start date != end date, something like 15%.  (Not clear how much, if any, of this is faulty data entry.)

Seems like it would be worth it to display this explicitly.

PR #1258, opened 2019-01-10
Split Fair screens into their own QueryRenderers

l2succes:
Currently, the query to render the initial Fair detail view is fetching more data than needed. This separates disparate screens into their own views


l2succes:
This has already been reviewed so I'm gonna go ahead and merge it

PR #1268, opened 2019-01-25
Fair more info

ashleyjelks:
Depends on: artsy/metaphysics#1459

Implements LD-99.

Adds the Fair More Information screen. Note that BMW art activations section is ignored for now.

#skip_new_tests 
- missing test is for an index file re-exporting a component with coverage


l2succes:
🙏🏾

PR #1272, opened 2019-01-28
updates Fair More Info query renderer and tests

ashleyjelks:
This is a refactor this code to update tests  and the Relay Query Render in a PR previously created by @javamonn 

Links to:

#1268 

#skip_new_tests
#trivial


l2succes:
Could you reset the changes to `Podfile.lock` as this would update versions for a many deps that shouldn't be affected by this change.

Example/Emission.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
l2succes: I think that's an IDE artifact that doesn't need to be checked in

ashleyjelks: Yep, checked it out of this branch.

PR #1275, opened 2019-01-29
Fixes CI

ashfurrow:
Native tests on CI are currently failing ([example](https://circleci.com/gh/artsy/emission/79)). I'm not able to reproduce the failure locally, so I've changed the Circle CI cache keys to force a clean build. We'll see if that works.


ashleyjelks:
LGTM!

PR #1276, opened 2019-01-29
LD-119 - Fair Exhibitor list now links to partner

kierangillen:
- ( https://artsyproduct.atlassian.net/browse/LD-119 )


src/lib/Scenes/Fair/Screens/FairExhibitors.tsx
```diff
@@ -34,18 +36,22 @@ export class FairExhibitors extends React.Component<Props, State> {
...
-  renderExhibitor(item) {
-    const { count, index } = item
+  renderExhibitor(data) {
+    const { item, index, section } = data
+    const { partnerIds, count } = section
     const generatedKey = count - index
     return (
       <Box mb={2} key={generatedKey}>
-        <Serif size="3">{item}</Serif>
+        <TouchableOpacity onPress={() => SwitchBoard.presentNavigationViewController(this, partnerIds[index])}>
```
l2succes: I would add a check in case `partnerIds.length === 0` for some strange reason

PR #1279, opened 2019-01-30
Removes "View on Map" text from More Shows section

ashleyjelks:
Small PR to remove text from map area in Shows view.
screen shot 2019-01-30 at 12 34 23 pm
screen shot 2019-01-30 at 12 34 04 pm

#skip_new_tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,7 @@
 - Reordering Fair sections to match design specs -ashley
 - Add contextual data to Fair view - ashley
 - Enable navigation on contextual data in Show view - ashley
+- Removes "View on Map" text from More Shows section in Show view -ashley
 
 ### 1.7.1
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -102,6 +102,7 @@
 - Reordering Fair sections to match design specs -ashley
 - Add contextual data to Fair view - ashley
 - Enable navigation on contextual data in Show view - ashley
+- Removes "View on Map" text from More Shows section in Show view -ashley
```
l2succes: @ashleyjelks You shouldn't add things under this section. We should remove this `User Facing` and `Not user facing` as it's not consistent with other versions & a source of confusion. It should go under **Master** all the way at the top. Also you should have a space after `-` 🙏🏾

ashleyjelks: Ooh, good catch, I've been putting these in the wrong section. I'll make a new branch and update the other notes here accordingly.

PR #1281, opened 2019-01-30
WIP: City view filtering and cross-component communication

ashfurrow:
This is a WIP to establish just the outline of what I'm thinking. Instead of providing a separate event for filter changes, we'll filter the objects at the `GlobalMap` component level, and send the filtered results alongside the full list each time with the `map:change` event. My next steps are to get the filtering logic implemented and a basic proof-of-concept for the cities UI (showing it responds to new data, etc).


src/lib/Scenes/Map/Bucket.ts
```diff
@@ -0,0 +1,49 @@
+import { GlobalMap_viewer } from "__generated__/GlobalMap_viewer.graphql"
+import { filter } from "lodash"
+
+export type BucketKey = "all" | "saved" | "fairs" | "galleries" | "museums"
```
l2succes: 👍

src/lib/Scenes/Map/Bucket.ts
```diff
@@ -0,0 +1,49 @@
...
+export type BucketResults = { [key in BucketKey]: GlobalMap_viewer }
+
+export const bucketCityResults = (viewer: GlobalMap_viewer): BucketResults => {
+  // TODO: Saved isn't supported yet, so we'll return nothing.
+  const saved = {
```
l2succes: We'll have to add a new field to our GraphQL query to resolve saved shows

PR #1282, opened 2019-01-31
Styles Fair view

kierangillen:
- First pass on styling fair view to match designs
- Still needs search styling

## Before

screen shot 2019-01-31 at 12 59 21 pm

## After

screen shot 2019-01-31 at 2 14 54 pm


src/lib/Components/HoursCollapsible.tsx
```diff
@@ -77,13 +77,15 @@ export class HoursCollapsible extends React.Component<Props, State> {
...
         <TouchableWithoutFeedback onPress={this.handleToggleIsExpanded}>
-          <Flex justifyContent="space-between" alignItems="center" flexDirection="row">
+          <Flex alignItems="center" flexDirection="row">
             <Sans size="3t" weight="medium">
-              Hours
+              Opening hours
             </Sans>
-            <Chevron source={require("../../../images/chevron.png")} isExpanded={isExpanded} />
+            <Box ml={1}>
+              <Chevron source={require("../../../images/chevron.png")} isExpanded={isExpanded} />
```
l2succes: I think we should consider using the svg version if only for consistency sake

kierangillen: cool I think this will need to be refactored in several places, I can do this in a separate PR?

l2succes: yup sounds good

PR #1283, opened 2019-01-31
Fair performance improvements 

ashleyjelks:
This PR introduces performance enhancements to the requests made to Metaphysics to populate the initial Fair view in local discovery.  

MP requests were timing out due to large responses that didn't need to be requested on the initial load of the Fair screen.

[Links to LD-169](https://artsyproduct.atlassian.net/browse/LD-169)

#skip_new_tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 - Centers Fair titles in fairbooth preview - kieran
 - Adds some small styling adjustments to the hoursCollapsible - kieran
 - Fix overfetching data on Fair Detail - luc
+- Update Show detail view spacing & copy - luc
 
 ### 1.8.1
 
--- a/src/lib/Scenes/Show/Components/ShowHeader/ShowHeader.tsx
+++ b/src/lib/Scenes/Show/Components/ShowHeader/ShowHeader.tsx
@@ -47,7 +47,7 @@ export class ShowHeader extends React.Component<Props> {
             sources={(images || []).map(({ url: imageURL, aspect_ratio: aspectRatio }) => ({ imageURL, aspectRatio }))}
           />
         )}
-        <Box px={2}>
+        <Box px={2} mb={4}>
           <ArtistNamesList artists={artists} Component={this} viewAllArtists={onViewAllArtistsPressed} />
           <Spacer mt={1} />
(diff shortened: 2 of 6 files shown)


src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -66,22 +66,29 @@ export class FairHeader extends React.Component<Props> {
...
-        profile_ids: [lastExhibitorProfileSlug],
-      },
-    ] = exhibitors_grouped_by_name
-
-    artistsCount = !firstArtistName ? artistsCount : artistsCount - 1
-    artistsCount = !lastArtistName ? artistsCount : artistsCount - 1
-    exhibitorsCount = !firstExhibitorName ? exhibitorsCount : exhibitorsCount - 1
-    exhibitorsCount = !lastExhibitorName ? exhibitorsCount : exhibitorsCount - 1
+    // @ts-ignore
+    artistsCount = artistsCount - Boolean(firstArtistName) - Boolean(lastArtistName)
```
l2succes: 😎 clever, didn't know `10 - true == 9`. Another interesting JS type interoperability

src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -201,16 +208,25 @@ export const FairHeaderContainer = createFragmentContainer(
...
-        exhibitors
-        profile_ids
-      }
-
       counts {
         artists
         partners
       }
 
+      partner_names: shows_connection(first: 2) {
```
l2succes: 👍

src/lib/Scenes/Fair/Screens/FairDetail.tsx
```diff
@@ -206,8 +206,11 @@ export const FairDetailContainer = createPaginationContainer(
             cursor
             node {
               ...FairBoothPreview_show
-              ...FairBooth_show
               ...ShowArtworks_show
+              ...FairBoothHeader_show
```
l2succes: We shouldn't need any of those fragments right? these come from `FairBooth_show` which is the unneeded fragment we want to eliminate from the initial query

PR #1285, opened 2019-02-01
Filter bar UI fix

ashfurrow:
This has been bugging me 😅 When tapping elements on the right side of the list, they'd be centred beyond the natural content offset of the scroll view. When tapping elements on the left side, they'd be ignored completely. This is most obvious on a smaller screen size:

bad gifcask 2019-01-31 18_58_12

The new code keeps the offset we scroll to within a bound corresponding to the scroll view's natural maximum content offset:

good gifcask 2019-01-31 18_58_45

Oh, and I also fixed a badly named file from #1281, which was [correctly reported](https://github.com/artsy/emission/pull/1281#issuecomment-459121808) as a failure but I ignored because it was surrounded by noise and I #skip_new_tests'd. I'll fix that tomorrow.


ashfurrow:
The map view looks off, too, like it's not tall enough.

l2succes:
😅 Thanks for that one @ashfurrow, been meaning to fix that

ashfurrow:
@l2succes it was a lot of fun to take something I've done a bunch in Objective-C, and put it in React 😄

PR #1289, opened 2019-02-04
City view data marshalling

ashfurrow:
I haven't finished the Objective-C code yet but wanted to get this in. It's not much, but it sets things up to be filled in:

screen shot 2019-02-04 at 4 16 43 pm

### DIFF
--- a/src/lib/Scenes/City/City.tsx
+++ b/src/lib/Scenes/City/City.tsx
@@ -2,26 +2,20 @@ import { Box, color, Flex, Serif } from "@artsy/palette"
 import React, { Component } from "react"
 import styled from "styled-components/native"
 import { EventEmitter } from "../Map/EventEmitter"
-
-export interface City {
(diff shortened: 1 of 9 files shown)


src/lib/Scenes/City/City.tsx
```diff
@@ -1,52 +1,59 @@
...
-    fairs: [],
-    title: "",
+    city: null,
+    filter: { id: "all", text: "All events" },
   }
 
   componentWillMount() {
     EventEmitter.subscribe("map:change", ({ filter, city }) => {
-      const { shows, fairs } = city
       console.log(city)
```
kierangillen: I'm assuming this is still needed?

ashfurrow: It's not _needed_ but it does help give us an idea of the shape of the data while we build the UI.

PR #1290, opened 2019-02-04
Polishes pulley view experience

ashfurrow:
This PR relies on https://github.com/l2succes/Pulley/pull/1 and is built on top of https://github.com/artsy/emission/pull/1289, so let's merge those two first. 
https://github.com/artsy/emission/commit/e573751a4e030c77a1a269076923552f1941f58e adds better support for the pulley drawer, including setting the scroll view's content inset (so that content doesn't extend below the bottom of the screen), disabling scrolling while the drawer is closed, and scrolling to the top when the drawer is closed by the user.

---

Edit: this is done now. The two big fixes are:

- Drawer now defaults to "partially closed" instead of totally closed (small at the bottom of the screen).
- The scroll view within the drawer now sets its content insets correctly to not extend content below the bottom edge of the screen.

Default when opening:

screen shot 2019-02-05 at 10 09 48 am

Opened and scrolled down:

screen shot 2019-02-05 at 10 09 55 am

#skip_new_tests


l2succes:
@ashfurrow Thanks for this. I been meaning to start setting that up. I think we should consider the slide up animation of the filter bar also when the bottom sheet VC goes past the mid position threshold. Not saying the animation has to go in this PR, but we should consider it as part of the API setup

ezgif-1-d72808d6aa5d

ashfurrow:
@l2succes oh nice, yeah. That'll be totally doable with the setup I've started here 👍 The container VC (the Pulley delegate) can refer to and set props on the two child VCs/components.

l2succes:
Let's go ahead and merge this one to make sure no one is blocked

ashfurrow:
@l2succes Kieran and I were just reviewing it – let's merge on green 👍

kierangillen:
Looks like theres a conflict @ashfurrow

kierangillen:
screen shot 2019-02-05 at 3 57 58 pm


Seeing this on this branch locally when building

ashfurrow:
After running a `pod install`, this is working for me locally. I'm going to merge and if we run into problems, I'm here to help.

PR #1291, opened 2019-02-05
Replaces PNG Chevron icon for SVG

kierangillen:
screen shot 2019-02-05 at 12 47 34 pm

screen shot 2019-02-05 at 12 47 43 pm

#skip_new_tests


src/lib/Icons/ChevronIcon.tsx
```diff
@@ -0,0 +1,35 @@
+import { color } from "@artsy/palette"
+import React from "react"
+import Svg, { G, Path } from "react-native-svg"
+
+const ChevronIcon = props => {
```
l2succes: More of a stylistic suggestion but a good pattern is to use object destructuring on `props` so `({expanded, initialDirection, width, height}) => `

l2succes:
merge on green

PR #1296, opened 2019-02-07
Show styling

kierangillen:
- Moves `Description` and `View more information` below map.
- Decreases size of pin slightly

screen shot 2019-02-07 at 12 30 18 pm

#skip_new_tests


src/lib/Components/Show/ShowArtworksPreview.tsx
```diff
@@ -21,7 +21,7 @@ export class ShowArtworksPreview extends React.Component<Props> {
     return (
       <>
         {title && (
-          <Serif size="6" mt={3} mb={3}>
+          <Serif size="6" mt={2} mb={3}>
```
damassi: again super minor (and i see that yr updating other code), but better to wrap <Serif> / typography in <Box> and adjust margins there since box is more structural / layout-like

PR #1298, opened 2019-02-08
LD-182 - Show address updates

kierangillen:
- Adds city and zip code to show address ( https://artsyproduct.atlassian.net/browse/LD-182 )
- Checks for works before rendering works section
- Checks for hours before rending hours section 

## Updates with 3t sizing

screen shot 2019-02-08 at 12 50 16 pm


src/lib/Components/LocationMap/index.tsx
```diff
@@ -16,6 +16,10 @@ const Map = styled(Mapbox.MapView)`
   height: 120;
 `
 
+const StyledSerif = styled(Serif)`
```
l2succes: I think we should use a value from Palette where the line height is smaller like `2t` or `3t` if possible

kierangillen: ok ill check that out

l2succes:
👍

PR #1300, opened 2019-02-08
Fixes Relay compiler invocation

ashfurrow:
A [failed emission deploy](https://travis-ci.org/artsy/emission-nebula/builds/490726056) led me to look at why `yarn relay` was failing. When we [renamed a test file](https://github.com/artsy/emission/commit/1957bf6410716a0b0ebd50ce4498a0c5ad1c5d59#diff-853dcc232c54ab09d1017908a0a753c1) to get around a Danger limitation (when making sure components had matching tests), we inadvertently caused a failure in the Relay compiler:

```
Parse error: Error: RelayFindGraphQLTags: Operation names in graphql tags must be prefixed with the module name and end in "Mutation", "Query", or "Subscription". Got `FairHeaderTestsQuery` in module `indexTests`. in "lib/Scenes/Fair/Components/FairHeader/__tests__/index-tests.tsx"
```

(This got merged into master because our only `yarn relay` CI invocation was running at a point where it cannot fail the build, which I also fixed in this PR.)

Interesting. I tried renaming the query to `indexTestQuery`, but that caused a conflict with yet _another_ `indexTestQuery`, at which point I stepped back to ask _hey why is Relay even running against our tests?_

I couldn't get the `--exclude` flag to work, I think because of how the Relay compiler parses array arguments, but I noticed that we're basically using the defaults anyway, so why not just remove it altogether?

That caused a problem with the `mockRelayRenderer` tests, which have a query renderer and was using the Relay compiler for type safety in the tests. I think that sacrificing some type safety for the sake of not shipping our unit test queries is a good trade-off.


orta:
less is less

ashfurrow:
@orta I'm definitely going to compare the size of our bundles before and after.

alloy:
Seems like a worthy trade-off 👍

ashfurrow:
Hmm! Our JS tests blew up. I'm going to PR a workaround to unblock LD while I look into this.

l2succes:
🙌🏽 ++++

ashfurrow:
(Just making a few notes so I can pick up on this on Monday...)

It looks like `Relay.graphql` in unit tests is causing problems. Here's code that fails:

```tsx
        query={graphql`
          query ShowArtistsPreviewTestsQuery {
            show(id: "anderson-fine-art-gallery-flickinger-collection") {
              ...ShowArtistsPreview_show
            }
          }
        `}
```

It fails because `graphql` the function is trying to do some module resolution based on who is calling the function (in this case, the test) but it can't look up a module that we no longer generate with Relay. This is an interesting problem, and I'm curious how other teams writing unit tests with Relay components handle this. Maybe they _do_ just have different Relay compiler settings for development-vs-deploy 🙃

ashfurrow:
This has kind of fallen off my radar, and it's a pretty minor improvement. I'm going to close this PR and we can revisit the issue if it comes up again.

PR #1301, opened 2019-02-08
Adds logging for relay queries and metaphysics extensions used in query

l2succes:
Related: https://github.com/artsy/reaction/pull/2002

screen shot 2019-02-08 at 4 52 03 pm


src/lib/relay/middlewares/cacheMiddleware.ts
```diff
@@ -0,0 +1,52 @@
...
+      return response
+    } else {
+      cache.clear(queryID, req.variables)
+
+      const error = new NetworkError(response.statusText)
+      error.response = response
+      throw error
+    }
+  }
+}
```
orta: good work

package.json
```diff
@@ -129,6 +129,7 @@
     "react-dom": "^16.3.0-alpha.1",
     "react-native-storybook-loader": "^1.6.0",
     "react-native-typescript-transformer": "^1.2.3",
+    "react-relay-network-modern": "^2.5.1",
```
damassi: Rad, so this was pretty straightforward to drop in

src/lib/relay/createEnvironment.ts
```diff
@@ -1,8 +1,32 @@
...
+import { metaphysicsURL } from "./config"
+import { cacheMiddleware } from "./middlewares/cacheMiddleware"
+import { metaphysicsExtensionsLoggerMiddleware } from "./middlewares/metaphysicsMiddleware"
+
+const Emission = NativeModules.Emission || {}
 
 export default function createEnvironment() {
-  const network = Network.create(fetchQuery)
+  const network = new RelayNetworkLayer([
+    cacheMiddleware(),
```
damassi: this one works quite well :)

src/lib/relay/middlewares/cacheMiddleware.ts
damassi: ahh i see -- rather than using the one that ships with the middleware you worked in yr own 👍

package.json
l2succes: Yup it was pretty straightforward aside from the cache middleware

src/lib/relay/middlewares/cacheMiddleware.ts
l2succes: Thank you sir 🙇🏾‍♂️

src/lib/relay/createEnvironment.ts
```diff
@@ -1,8 +1,32 @@
-import { Environment, Network, RecordSource, Store } from "relay-runtime"
-import { fetchQuery } from "./fetchQuery"
+import { NativeModules } from "react-native"
+import { errorMiddleware, loggerMiddleware, RelayNetworkLayer, urlMiddleware } from "react-relay-network-modern/node8"
```
ashfurrow: I don't know if it matters, but we're on node 10 instead of 8.

ashfurrow:
This looks great!

src/lib/relay/middlewares/cacheMiddleware.ts
orta: The Emission one takes into account that the native app can make queries *before* JS launches (And from other screens etc) so we need a unique cache model 👍

l2succes: Yes, precisely it uses a native cache which can pre warmed before views get rendered

src/lib/relay/createEnvironment.ts
l2succes: It doesn't matter here. This version makes use of new features that became available in node 8 (i.e. native async/await)

orta:
You want these changes: https://github.com/artsy/reaction/pull/2008

PR #1303, opened 2019-02-08
LD-189 - EntityList partner links

kierangillen:
- Parter names in EntityList now link to LD view
- https://artsyproduct.atlassian.net/browse/LD-189


src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -15,6 +15,7 @@ import { CountdownTimer } from "./CountdownTimer"
 interface Props {
   fair: FairHeader_fair
   onSaveShowPressed?: () => Promise<void>
+  onViewFairBoothPressed: () => Promise<void>
```
l2succes: we don't need this anymore right

PR #1310, opened 2019-02-12
Fixes show save

kierangillen:
- Uses correct ID for unfollow


src/lib/Scenes/Show/Components/ShowHeader/ShowHeader.tsx
```diff
@@ -48,28 +48,30 @@ export class ShowHeader extends React.Component<Props, State> {
...
-                is_followed: !is_followed,
+                _id: showID,
+                is_followed: !isShowFollowed,
+                id: showSlugID,
               },
             },
           },
           updater: store => {
-            store.get(__id).setValue(!is_followed, "is_followed")
+            store.get(relayID).setValue(!isShowFollowed, "is_followed")
```
ashfurrow: I really like the disambiguated naming here, nice.

PR #1313, opened 2019-02-12
Refactor CityView to accept all buckets and adds basic styling

l2succes:
this allows us to render "All" or filtered views depending on selected filter

screen shot 2019-02-13 at 2 45 09 pm

#skip_new_tests


ashfurrow:
I trust you on this, but we lose some contextual information (what city am I in?) in the filtered views with this change. The designs might not require that info, but just thought I'd mention 👍

l2succes:
@ashfurrow Yup that's true but we actually don't need the `city` info in the City view for now, if we do we can always pass a shallow object as a prop. I refactored it that way because the `all` filter needs every bucket in order to render each sections

ashfurrow:
@l2succes makes sense, thanks 👍

PR #1315, opened 2019-02-13
Fair/Show QA fixes

kierangillen:
- Show header title now links to gallery ( https://artsyproduct.atlassian.net/browse/LD-210 )
- Show image now full width if one image ( https://artsyproduct.atlassian.net/browse/LD-212 )
- Adds spacer below links in show about page ( https://artsyproduct.atlassian.net/browse/LD-218 )

## Single image example

screen shot 2019-02-13 at 4 45 07 pm

## Spacer below Show about links

screen shot 2019-02-13 at 5 08 15 pm


Example/Emission/AppDelegate.m
```diff
@@ -280,7 +280,8 @@ - (UIViewController *)viewControllerForRoute:(NSString *)route;
     NSURLComponents *components = [[NSURLComponents alloc] initWithString:route];
     NSString *artistID = [self valueForKey:@"artist_id" fromQueryItems:components.queryItems];
     viewController = [[ARHomeComponentViewController alloc] initWithSelectedArtist:artistID tab:ARHomeTabArtists emission:nil];
-
+  } else if ([route hasSuffix:@"entity=gallery"]){
+    return [[UnroutedViewController alloc] initWithRoute:route];
   } else {
```
l2succes: Can you actually make that change for fair? right now it's catching any url that doesn't fall in one of the VCs specified.

kierangillen: sure ill add that now

kierangillen: Do we need to add this to all links to Fairs in the RN code? May need to comb through the codebase for those..

l2succes: Yeah we also need to update links to include that in the RN code

ashfurrow: @l2succes yeah that's a better way to go, good thinking. In Eigen, everything should "just work" since it does a Gravity request to determine the entity. This new query string @kierangillen added will help us avoid a roundtrip on Eigen's side, but it's only a performance optimization (things should work as-is once integrated into Eigen).

PR #1317, opened 2019-02-14
Removes empty opening hours for emtpy schedules

ashfurrow:
This PR is a follow up to https://github.com/artsy/metaphysics/pull/1511 , which modified Metaphysics to return an empty `location.displayDaySchedules` array for shows missing schedules. This PR is a follow-up that checks for that empty value.

Unit tests are blocked by [LD-158](https://artsyproduct.atlassian.net/browse/LD-158) so I've left a note about that.

screen shot 2019-02-14 at 5 07 47 pm


src/lib/Scenes/Show/Screens/Detail.tsx
```diff
@@ -63,7 +64,7 @@ export class Detail extends React.Component<Props, State> {
       },
     })
 
-    if (show.location && show.location.displayDaySchedules) {
+    if (show.location && !isEmpty(show.location.displayDaySchedules)) {
```
kierangillen: @ashfurrow as discussed we should also look for `day_schedule_text` here and also make sure we're handling that field in the component

ashfurrow: Good catch! I took a closer look and this relates to some work that @anandaroop has [ticketed in LD-221](https://artsyproduct.atlassian.net/browse/LD-221) for Metaphysics to do this check, and then have the clients consume that (instead of having the clients consume both and make a guess at which to use). This approach centralizes a lot of UI logic that would otherwise need to be replicated across each client. 

(My guess is that Metaphysics will have a field that's a union of either a string or structured data, and the client uses the `__typename` field to determine which it is.)

I think it makes sense to continue with this change until we complete LD-221, but let me know what I can clarify.

PR #1319, opened 2019-02-15
Fair QA fixes

kierangillen:
- Fixes inconsistent padding between items in show and fair views
- Removes fair save button
- Adds params to fair link
- Fixes Exhibitor name overflows and pushes follow button off the screen ( https://artsyproduct.atlassian.net/browse/LD-206 )

## Removes save button

screen shot 2019-02-15 at 1 38 57 pm

## Fixes exhibitor overflow

screen shot 2019-02-15 at 2 53 55 pm


src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -1,7 +1,7 @@
 import { Box, Flex, Sans, space, Spacer } from "@artsy/palette"
 import { FairHeader_fair } from "__generated__/FairHeader_fair.graphql"
 import { FairHeaderMutation } from "__generated__/FairHeaderMutation.graphql"
-import InvertedButton from "lib/Components/Buttons/InvertedButton"
+// import InvertedButton from "lib/Components/Buttons/InvertedButton"
```
ashfurrow: We should remove this if it's no longer needed.

src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -162,7 +162,8 @@ export class FairHeader extends React.Component<Props, State> {
     } = this.props
     const { width: screenWidth } = Dimensions.get("window")
     const imageHeight = 567
-    const { isSavedFairStateUpdating } = this.state
+    // const { isSavedFairStateUpdating } = this.state
+    // FIXME: When saving fairs is defined we can use commented code
```
ashfurrow: I'd prefer if we removed the `isSavedFairStateUpdating` state value and `handleSaveFair`. If it's not getting called, it should be removed – but we can still access it in git history.

We should add a reference to this pull request in the ticket to re-add the fair saving, though, to make it easy to add back in 👍

src/lib/Scenes/Home/Components/ForYou/Components/FairsRail.tsx
```diff
@@ -58,7 +58,7 @@ export class FairsRail extends Component<Props, null> {
       }
 
       const selectionHandler = () => {
-        Switchboard.presentNavigationViewController(this, fair.profile.href)
+        Switchboard.presentNavigationViewController(this, `${fair.profile.href}?entity=fair`)
```
ashfurrow: Exxxxxxcellent

src/lib/Scenes/Fair/Components/FairHeader/index.tsx
kierangillen: ok sounds good

PR #1325, opened 2019-02-20
FairBooth routing

kierangillen:
- Fixes https://artsyproduct.atlassian.net/browse/LD-209 and https://artsyproduct.atlassian.net/browse/LD-208


l2succes:
Nice one 👌🏾

PR #1326, opened 2019-02-20
Sets up making sure there's a phone number on a users's account when signing up for an auction

orta:
Supports ensuring that the phone number is set when signing up for an auction


sweir27:
@orta 🤞 that the tests are green now! I also added a few more for the reg flow.

orta:
OK, yep, I've given it a run through locally since upgrading to RN 0.57  (and made a minor test refactor) 

All looks good - merge on green

orta:
This'll be blocked on a beta until LD launch, but will come out in the next patch after that

sweir27:
Thanks @orta !

PR #1327, opened 2019-02-21
[WIP] - Fair analytics

kierangillen:
- Adds fair analytics

#skip_new_tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 - Fixes FairBooth routing - kieran
 - Fair exhibitors overview looks up to spec - orta
 - Adds Shows to Global Saves and Follows view -ashley
+- Adds seperator between fair entities - kieran
 
 ### 1.8.2
 
--- a/src/lib/Scenes/Fair/Components/FairHeader/index.tsx
+++ b/src/lib/Scenes/Fair/Components/FairHeader/index.tsx
@@ -83,6 +83,7 @@ export class FairHeader extends React.Component<Props, State> {
           onItemSelected={this.handleWorksPress}
           onViewAllPressed={viewAllArtists}
         />
+        <Spacer mt={1} />
         <EntityList
           prefix="From"
           list={partnerList}


src/lib/Components/FilteredInfiniteScrollGrid/FilteredInfiniteScrollGrid.tsx
```diff
@@ -8,6 +9,8 @@ import { FiltersContainer as Filters } from "./Filters"
 interface Props {
   onRefetch: (params: object) => any
   filteredArtworks: any
+  fairSlug?: string
```
l2succes: hmm.. this is meant to be a generic component so adding fair specific props isn't great.
Maybe name this `entityData: {id: string, slug: string}`

l2succes: That being said I think it's ok for now

src/lib/Scenes/Fair/Components/FairBoothHeader.tsx
```diff
@@ -28,16 +29,43 @@ const ButtonWrapper = styled(Box)`
...
+const track: Track<Props, State> = _track
+
+@track()
 export class FairBoothHeader extends React.Component<Props, State> {
   state = {
     isFollowedChanging: false,
   }
 
+  @track((props, _, args) => {
+    const partnerSlug = args[0]
```
l2succes: I think this could fail for fairs that have partner === `null` but in this case we shouldn't show the follow button at all.

kierangillen: Are you sure? This function is only called when a partner is followed or unfollowed so I dont think the partner can be `null`

src/lib/Components/FilteredInfiniteScrollGrid/FilteredInfiniteScrollGrid.tsx
kierangillen: Updated this to make it more generic

PR #1329, opened 2019-02-22
New Map filter design 

l2succes:
#skip_new_tests

### DIFF
--- a/Pod/Classes/ViewControllers/ARComponentViewController.h
+++ b/Pod/Classes/ViewControllers/ARComponentViewController.h
@@ -7,7 +7,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @protocol ARComponentFullBleedViewSizing
 
-@property (nonatomic, assign) BOOL fullBleed;
+@property (nonatomic, readonly) BOOL fullBleed;
 
 @end
 
--- a/Pod/Classes/ViewControllers/ARComponentViewController.h
+++ b/Pod/Classes/ViewControllers/ARComponentViewController.h
@@ -5,7 +5,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@protocol ARComponentFullBleedViewSizing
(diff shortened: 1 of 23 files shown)


Pod/Classes/ViewControllers/ARComponentViewController.h
```diff
@@ -5,10 +5,19 @@
...
 NS_ASSUME_NONNULL_BEGIN
 
-@interface ARComponentViewController : UIViewController
+@protocol ARComponentFullBleedViewSizing
+
+@property (nonatomic, assign) BOOL fullBleed;
+
+@end
+
+@interface ARComponentViewController : UIViewController <ARComponentFullBleedViewSizing>
```
orta: I'm not sure that this protocol adds much value - as every component VC supports it, the Eigen subclass can use the ARMenuAware protocol on that side of the fence though

Pod/Classes/ViewControllers/ARMapComponentViewController.m
```diff
@@ -7,6 +7,21 @@ - (instancetype)init
...
+                                [NSLayoutConstraint constraintWithItem:self.rootView
+                                                             attribute:NSLayoutAttributeTop
+                                                             relatedBy:NSLayoutRelationEqual
+                                                                toItem:self.view
+                                                             attribute:NSLayoutAttributeTop
+                                                            multiplier:1
+                                                              constant:0]
+                                ]];
+}
+
```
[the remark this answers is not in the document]
l2succes: No, that's not needed anymore

Pod/Classes/ViewControllers/ARComponentViewController.h
orta: This is also used on a non-ARComponentVC, so there's value

orta:
Alright, this has been rebased and brought up to master post https://github.com/artsy/emission/pull/1345

orta:
OK, looking good. Merging 👍

PR #1331, opened 2019-02-22
Enable viewing saved shows in global saves/follows view

ashleyjelks:
This PR exposes the Shows tab in the Favorites screen and enables users to view all their saved shows there. 

[Links to LD-194](https://artsyproduct.atlassian.net/browse/LD-194)

kapture 2019-02-22 at 10 40 31

#skip_new_tests


src/lib/Components/Lists/SavedShowItemRow.tsx
```diff
@@ -0,0 +1,111 @@
...
+  status: string
+  images: {
+    url: string
+  }
+  partner: {
+    name: string
+  }
+  is_followed: boolean
+  relayEnvironment: RelayProp
+}
```
orta: I feel like this should be generated for you somewhere?

ashfurrow: I think it's a nested part of the type generated by a fragment, so I don't think it's accessible from a "this object is this shape" perspective. 

A better approach could be to define the fragment in `SavedFairItemRow` and then spread across that fragment in the `SavedFairs` component instead. I'm working in this code for analytics, so I can take on that refactor after this PR is merged.

l2succes: Yup that'd be preferable indeed. sounds good @ashfurrow

l2succes:
I agree we should have a relay container for `SavedShowItemRow` but we can merge as is for now and tackle during analytics work

PR #1333, opened 2019-02-22
Global Saves&Follows tracking, SavedShowItemRow fragmentification

ashfurrow:
This PR is for [LD-196](https://artsyproduct.atlassian.net/browse/LD-196) and does two things:

- Adds tracking 
  - for tapping into the saves tab
  - for tapping save/unsave
- Refactors the `SavedShowItemRow` component into a fragment container [as discussed in 1331](https://github.com/artsy/emission/pull/1331#discussion_r259403891)

I'm going to leave a few comments inline describing the fragment refactor – please let me know what I can clarify. 

#skip_new_tests


src/lib/Scenes/Favorites/Components/Shows/index.tsx
```diff
@@ -75,23 +75,7 @@ export default createPaginationContainer(
...
-                  ... on ExternalPartner {
-                    name
-                  }
-                }
-                href
-                status
-                images(size: 1) {
-                  url
-                }
+                ...SavedShowItemRow_show
```
ashfurrow: This is the core change – defining the data for the `SavedShowItemRow` "closer" to the code that uses it. The parent component becomes more focused while the child component gets the benefits of generated types.

src/lib/Scenes/Favorites/Components/Shows/index.tsx
```diff
@@ -54,7 +54,7 @@ export class SavedShows extends Component<Props, State> {
       <FlatList
         data={shows}
         keyExtractor={item => item.__id}
-        renderItem={item => <SavedShowItemRow {...item.item} relayEnvironment={this.props.relay} />}
+        renderItem={item => <SavedShowItemRow show={item.item} />}
```
ashfurrow: Notice here how the `relay` prop is passed in implicitly by Relay for us.

src/lib/Components/Lists/SavedShowItemRow.tsx
```diff
@@ -1,38 +1,45 @@
...
-  status: string
-  images: {
-    url: string
-  }
-  partner: {
-    name: string
-  }
-  is_followed: boolean
-  relayEnvironment: RelayProp
+  show: SavedShowItemRow_show
```
ashfurrow: By defining our properties in a fragment (below), the Relay compiler will generate the `SavedShowItemRow_show` type for us, which we import above. This is great because we don't have to define our types anymore – they're defined by the Metaphysics schema for us 🎉

ashleyjelks: Nice!

ashleyjelks:
Looks great Ash!

PR #1334, opened 2019-02-22
Fixes rendering markdown in Fair/Show (but also everywhere else

orta:
This never worked, so now if anyone creates a new item in the markdown config a test will fail if they don't remember to do the `...default`.

Fixes LD-214

### DIFF
--- a/src/lib/Components/Markdown.tsx
+++ b/src/lib/Components/Markdown.tsx
@@ -50,12 +50,14 @@ export const defaultRules = {
   },
 
   newline: {
+    ...SimpleMarkdown.defaultRules.newline,
     react: (_node, _output, state) => {
       return <Text key={state.key}>{"\n"}</Text>
     },
   },
 
   strong: {
+    ...SimpleMarkdown.defaultRules.strong,
     react: (node, output, state) => {
       return (
         <Sans size="3t" weight="medium" key={state.key}>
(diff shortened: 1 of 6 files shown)


src/lib/Components/Markdown.tsx
```diff
@@ -66,6 +68,7 @@ export const defaultRules = {
   },
 
   br: {
+    ...SimpleMarkdown.defaultRules.br,
     react: (_node, _output, state) => {
       return <Text key={state.key}>{"\n\n"}</Text>
```
orta: these were what was missing, none of the above would actually render and so we never showed br or newlines

anandaroop: Whoa, nice catch.

src/lib/Components/__tests__/Markdown-tests.tsx
```diff
@@ -92,3 +92,14 @@ describe("Markdown", () => {
...
   })
 })
+
+describe("defaultRules", () => {
+  Object.keys(defaultRules).forEach(key => {
+    if (key === "Array") {
+      return
+    }
+    it(`has a match rule for ${key}`, () => {
+      expect(defaultRules[key].match).toBeTruthy()
```
anandaroop: Smart 😎

PR #1335, opened 2019-02-22
Checks for whether we can show more info about a fair

orta:
After:
screen shot 2019-02-22 at 6 46 46 pm

I didn't want to just allow someone to end up in a weird state where they were then told to go directly to a page with only one option (to go to another page) so I changed the spec somewhat and if the fair just has a website but no metadata then it'll take you directly to the site instead.


CHANGELOG.md
```diff
@@ -16,6 +16,12 @@
 
 ### Master
 
+- Fair exhibitors overview looks up to spec - orta
+- Hides additional info when a fair doesn't have it - orta
+- When there's only a website for the fair additional info, the app goes there - orta
+
+### 1.8.3
```
orta: This got shipped today, no idea about the boundaries for everyone else work, but mine is all post-1.8.3

src/lib/Scenes/Fair/Fair.tsx
```diff
@@ -106,6 +110,12 @@ export const FairContainer = createFragmentContainer(
       id
       __id
       ...FairDetail_fair
+
+      organizer {
+        website
+      }
+      about
+      ticketsLink
     }
```
orta: This gets put in a few places now, it increases the main payload a bit sure - but additional will now be instant thanks to caching.

src/lib/Scenes/Fair/Screens/FairMoreInfo.tsx
```diff
@@ -24,6 +24,20 @@ interface State {
...
+  ticketsLink?: string
+}
+
+export const shouldShowFairMoreInfo = (data: ShowMoreMetadataForFairs) => {
+  return data.ticketsLink || data.about
+}
+
+export const shouldGoStraightToWebsite = (data: ShowMoreMetadataForFairs) => {
+  return !shouldShowFairMoreInfo(data) && (data.organizer && data.organizer.website)
+}
```
orta: puts the "should we do this" functions next to the data it should show

src/lib/Scenes/Fair/Screens/FairDetail.tsx
```diff
@@ -150,10 +151,13 @@ export class FairDetail extends React.Component<Props, State> {
         )
       case "information":
         return (
-          <>
-            <CaretButton onPress={() => data.onViewMoreInfoPressed()} text="View more information" />
-            <Separator mt={2} />
-          </>
+          shouldGoStraightToWebsite(this.props.fair) ||
```
kierangillen: Can we move this logic into the `updateSections` function and apply the conditional before we push the `location` type into the sections array? The padding is still rendering around the box wrapper even if we return `null` here.

orta: Yep, done 👍

PR #1339, opened 2019-02-25
Adds a componentshouldupdate to the Markdown component

orta:
This could/should improve the animation performance on fairs


l2succes:
👍🤞🏾this fixes our animation performance issue

PR #1340, opened 2019-02-25
Add docs about troubleshooting installs

ashkan18:
Ran into few issues during install, adding the solutions to troubleshooting doc.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,5 +1,63 @@
 # Troubleshooting
 
+## Installation Issues
+
+- Failed `make artsy` with
+
+```
+  You will need to run:
+    yarn install
+```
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -21,3 +79,7 @@ console.log(source) // => { uri: "pods/Assets/.../some-video.mp4" }
 ```
 
 See [Video.tsx](https://github.com/artsy/emission/tree/master/src/lib/Components/Video.tsx) for an example implementation and [here](https://facebook.github.io/react-native/docs/images#static-non-image-resources) for a list of supported file formats.
+
+```
+
+```
```
orta: You can merge after killing this

PR #1341, opened 2019-02-25
Add cursor saved shows

ashleyjelks:
The Relay query that handled Saved/Followed Shows in the Global Favorites view was throwing a cursor error because it was not being included in the query. I suspect the missing cursor also caused the bug reported in [LD-244](https://artsyproduct.atlassian.net/browse/LD-244) where saved shows were not consistently rendering in the Saved Shows tab 

#trivial
#skip_new_tests


src/lib/Scenes/Favorites/Components/Shows/index.tsx
```diff
@@ -66,13 +66,17 @@ export class SavedShows extends Component<Props, State> {
...
+  Shows,
   {
     me: graphql`
       fragment Shows_me on Me
-        @argumentDefinitions(count: { type: "Int", defaultValue: 10 }, cursor: { type: "String", defaultValue: "" }) {
+        @argumentDefinitions(count: { type: "Int", defaultValue: 10 }, cursor: { type: "String" }) {
         followsAndSaves {
           shows(first: $count, after: $cursor) @connection(key: "SavedShows_shows") {
+            pageInfo {
+              endCursor
```
l2succes: 👍

PR #1345, opened 2019-02-26
Moves Fair sub screens to separate Native VCs

l2succes:
Fixes:
- https://artsyproduct.atlassian.net/browse/LD-246
- https://artsyproduct.atlassian.net/browse/LD-240
- https://artsyproduct.atlassian.net/browse/LD-239
- https://artsyproduct.atlassian.net/browse/LD-217

#skip_new_tests


src/lib/Scenes/Show/Screens/Detail.tsx
```diff
@@ -134,19 +134,6 @@ export class Detail extends React.Component<Props, State> {
     return null
   }
 
-  handleAnimationFrame = animatedValue => {
-    /**
-     * If children change their size on animation (e.g. HoursCollapsible), we need a sentinel value
-     * in state in order to trigger a re-render, as FlatList statically sizes child cells.
-     */
-    this.setState({
```
l2succes: @orta @kierangillen This is the root cause of the animation jerkiness. This callback was being called at every frame (~60fps) and calling `setState` which would potentially trigger a render pass (only potentially because it's an async call). After removing this, the animation performed

orta: heeyyyyyyy

CHANGELOG.md
```diff
@@ -61,7 +62,9 @@
 - Fair exhibitors overview looks up to spec - orta
 - Adds Shows to Global Saves and Follows view -ashley
 - Adds seperator between fair entities - kieran
+- Fixes FairBooth routes - luc
 - Adds Fair analytics - kieran
+- Adds Component VCs for Artists, Artworks, Exhibitors screens - luc
```
orta: This isn't in master

orta:
Nice find on the callbacks 👍

orta:
I fixed the CHANGELOG myself, and this is ready to merge on green

PR #1347, opened 2019-02-26
Fixes partner link on Fair booth header

l2succes:
Fixes: https://artsyproduct.atlassian.net/browse/LD-247


src/lib/Scenes/Fair/Components/SearchLink.tsx
```diff
@@ -30,7 +30,7 @@ export class SearchLink extends React.Component<Props> {
 
   render() {
     return (
-      <TouchableWithoutFeedback onPress={() => this.handlePress}>
+      <TouchableWithoutFeedback onPress={() => this.handlePress()}>
```
l2succes: Just noticed the search bar on Fair was broken and bundled that fix there as it's very small

l2succes:
@ashleyjelks Gonna merge this one as it's #trivial

src/lib/Scenes/Show/Screens/Detail.tsx
```diff
@@ -100,10 +100,12 @@ export class Detail extends React.Component<Props, State> {
       },
     })
 
-    sections.push({
-      type: "shows",
-      data: show,
-    })
+    if (show.location) {
```
ashkan18: should we use `show.has_location`? the logic behind `has_location` is slightly different and thats what defines if the show is online exclusive or not. What do you think? it's already exposed in MP for `Show`

l2succes: Ahhh, just saw this. Yeah that makes sense. I'll make a separate PR to use `has_location` instead

PR #1350, opened 2019-02-26
Fix timezone tests

orta:
This is the technique they use in the [timer tests](https://github.com/artsy/emission/blob/fcb48de8de7d728cc9a1e29f17e67d7221b0e1e1/src/lib/Components/Bidding/Components/__tests__/Timer-tests.tsx).

#trivial


kierangillen:
Thanks. Merge on green.

PR #1351, opened 2019-02-26
Show/Fair Routing Migration

ashfurrow:
This PR is for [LD-250](https://artsyproduct.atlassian.net/browse/LD-250) and relates to [LD-233](https://artsyproduct.atlassian.net/browse/LD-233). This PR is a long story.

We want to use the existing infrastructure to get our React Native scroll views playing nicely with Eigen's back button. This _should_ involve making the component controllers confirm to a particular protocol in Eigen:

https://github.com/artsy/eigen/blob/d9188d3816142a0a0835c15088e4be6f4fc54639/Artsy/App/ARAppDelegate%2BEmission.m#L340-L341

I went to extend `ARFairComponentViewController` in the same way, but ran into a problem. Our solution relies on the `viewDidLayoutSubviews` `UIViewController` callback, but having a nested navigation controller inside our RN component controller breaks that callback. Nesting navigation controllers is _never_ a good idea, and normally will crash an iOS app, but we nest things a level deeper than UIKit checks for (`ARFairComponentViewController` -> `UIViewController` -> `UINavigationController`).

This is the third time I've run into a bug that was caused by or blocked by our nesting, so I finally removed `NavigatorIOS` from our components. This wasn't too much work, and works in concert with an Eigen change to fix the bug in LD-250.

*This requires new routing logic in Eigen* and I don't want to block our current release with a huge refactor, so *this PR is a draft* until we get v0 released.

#skip_new_tests


ashfurrow:
Oh, I need to review this to make sure we aren't loosing any tracking information, too.

src/lib/Scenes/Fair/Components/FairBoothPreview/index.tsx
```diff
@@ -120,6 +119,14 @@ export class FairBoothPreview extends React.Component<Props, State> {
     SwitchBoard.presentNavigationViewController(this, `show/${showID}?entity=fair-booth`)
   }
 
+  viewFairBoothPressed() {
+    console.log("HI ASH")
```
ashfurrow: Ha, oops.

ashfurrow:
"About the fair" component needs to be moved down:

screen shot 2019-02-26 at 4 58 25 pm

ashfurrow:
Okay, except for the two visual problems and [a change we'll need to make on Eigen's side](
https://github.com/artsy/eigen/pull/2782#issuecomment-467631890), this looks pretty good! I expected to find more places that broke without the extra navigation controller, but yeah!

ashfurrow:
Okay, this is ready for review. I'm going to leave it as a draft PR until the new versions of Eigen/Emission are submitted.

ashfurrow:
We discussed this today, but we're going to review+merge this and do more QA.

ashfurrow:
Argh, CI failure is just Danger complaining about missing tests.

src/lib/AppRegistry.tsx
```diff
@@ -158,6 +167,34 @@ const Show: React.SFC<ShowProps> = track<ShowProps>(props => {
   return <ShowRenderer showID={showID} render={renderWithLoadProgress(Containers.Show, { showID })} />
 })
 
+interface ShowArtistsProps {
+  showID: string
+}
+const ShowArtists: React.SFC<ShowArtistsProps> = ({ showID }) => {
```
l2succes: We'll have to add analytics for those

ashfurrow: Makes sense – can we follow-up on that in another PR once the native changes are made?

src/lib/Scenes/Show/Screens/MoreInfo.tsx
```diff
@@ -98,6 +97,8 @@ export class MoreInfo extends React.Component<Props, State> {
   }
 
   render() {
+    const topMargin = 54 // Palette's space() is out of date.
+    console.log({ topMargin })
```
l2succes: stray console.log

src/lib/AppRegistry.tsx
l2succes: Yeah for sure

ashfurrow:
Merge on green.

src/lib/AppRegistry.tsx
ashfurrow: Am I correct in understanding that the analytics _weren't_ there before? I'll look up the schema doc and make sure everything is accounted for.

l2succes: Yeah actually that's right, we didn't have those screens before so definitely not a blocker

PR #1352, opened 2019-02-26
Make sure event name is ellipsis(ed?)

ashkan18:
https://artsyproduct.atlassian.net/browse/LD-265

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 -->
 
 ### Master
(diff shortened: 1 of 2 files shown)


orta:
Perfect 👍

orta:
Might want a changelog entry?

kierangillen:
Awesome, can you also add it to the Gallery/Partner line? We will need to add this elsewhere where it's not present in Show/Fair views, but can be separate PR.

After added to first line we can merge on green.

PR #1353, opened 2019-02-26
Adds clickable links for a map preview

orta:
Fixes LD-187

Allows clicking on a map and opening up a set of different maps providers, and lets you copy it for whatever obscure app you use also. 

screen shot 2019-02-26 at 3 20 19 pm


l2succes:
🔥💃  We can switch `Copy Address` with `Open in app` when it's ready

orta:
Exactly, yah

PR #1356, opened 2019-02-26
Fixes to the fair icon so that we correctly get transparency

orta:
These are the safely rendered pngs which are used in fair contexts around Artsy. They're 140x140, which is _ok_ but not great and feels a bit blurry because it'd should be a 2x image. To try reduce this blurriness, I reduced its size on screen to exactly 140 instead of 150. 

To do this right, we need square280 (or square300 to fit design spec exactly) which are rendered from the original images given in waves.


l2succes:
Makes sense, let's see if we can get the fair team to upload higher res png's. we can also look at what size is used on web as those are not blurry

PR #1357, opened 2019-02-26
City tabs

kierangillen:
- Adds tab data to city tabs
- Adds Upcoming and Closing shows to All events tab
- Adds `Saved` to buckets

#skip_new_tests


src/lib/Scenes/Map/Bucket.ts
```diff
@@ -1,22 +1,36 @@
...
+export type BucketKey = "saved" | "fairs" | "galleries" | "museums" | "closing" | "opening"
 export type BucketResults = { [key in BucketKey]: any[] }
 
 export const bucketCityResults = (viewer: GlobalMap_viewer): BucketResults => {
-  // TODO: Saved isn't supported yet, so we'll return nothing.
-  const saved = []
-
-  const fairs = (viewer.city.fairs as unknown) as Tab[]
+  const saved = filter(viewer.city.shows.edges, e => e.node.is_followed === true)
+  const oneWeekFromNow = moment(new Date()).add(1, "week")
```
ashkan18: #minor i wonder if we should make these constants somewhere, to define the thresholds for upcoming and closing.

src/lib/Scenes/City/Components/TabListItem/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+                  <Sans size="3t" color="black60" numberOfLines={1} ellipsizeMode="tail">
+                    {formatDuration(item.node.start_at, item.node.end_at)}
+                  </Sans>
+                )}
+            </Box>
+          </>
+        )
+      default:
+        const galleryImage = item.node.cover_image ? item.node.cover_image.url : null
+        return (
```
l2succes: Is that just the same as the other ShowItem component from the global Saves & Follows? If yes we might want to abstract that

PR #1362, opened 2019-02-27
Improvements to fair handling for map preview taps

orta:
- The title now will take into account fairs, and is always the same as the text copied. 
- Google maps now only gets lat/lng and under the hood that is shown as an address (on the site at least)
- Better handling of no postal code existing

screen shot 2019-02-27 at 1 56 19 pm

#trivial


orta:
Spotted a bug, fixed it

screen shot 2019-02-27 at 2 12 41 pm

l2succes:
You too quick with it @orta  🔥👌🏾

PR #1363, opened 2019-02-27
Fixes fair/shows links throughout the app

ashfurrow:
Discussed in Slack: https://artsy.slack.com/archives/CFNFWD5F0/p1551296889331300

#skip_new_tests


src/lib/Components/Show/ShowArtistsPreview.tsx
```diff
@@ -24,7 +24,7 @@ export class ShowArtistsPreview extends React.Component<Props> {
     owner_type: Schema.OwnerEntityTypes.Show,
   }))
   handlePress(url: string) {
-    Switchboard.presentNavigationViewController(this.props.Component, url)
+    Switchboard.presentNavigationViewController(this.props.Component || this, url)
```
ashfurrow: Fixed a nasty bug here.

src/lib/Components/Lists/SavedShowItemRow.tsx
```diff
@@ -18,7 +18,8 @@ const track: Track<Props, {}> = _track
 @track()
 class SavedShowItemRow extends React.Component<Props> {
   handleTap() {
-    Switchboard.presentNavigationViewController(this, this.props.show.href)
+    console.log({ show: this.props.show })
+    Switchboard.presentNavigationViewController(this, this.props.show.href || `show/${this.props.show.id}`)
```
ashfurrow: This was why we were seeing some shows in the Favourites tab direct back to the Home tab.

l2succes: Ah... got it

l2succes:
Nice catch @ashfurrow

src/lib/Components/Show/ShowArtistsPreview.tsx
orta: that was the bug?

src/lib/Scenes/Home/Components/ForYou/Components/FairsRail.tsx
```diff
@@ -58,7 +58,7 @@ export class FairsRail extends Component<Props, null> {
       }
 
       const selectionHandler = () => {
-        Switchboard.presentNavigationViewController(this, `${fair.profile.href}?entity=fair`)
+        Switchboard.presentNavigationViewController(this, `${fair.id}?entity=fair`)
```
orta: I just did something like this - doesn't it need to be different depending on whether it has a profile or not? I just had a bug like this in native code

orta: https://github.com/artsy/eigen/pull/2772

PR #1365, opened 2019-02-27
Fixes fair view crash on fairs with no profiles #trivial

ashfurrow:
For example: https://www.artsy.net/affordable-art-fair-battersea-spring-2019


orta:
nice call

PR #1372, opened 2019-03-01
Show favorite artists in the list of participating artists first

ashkan18:
# Problem
We want to show the artists followed by current user first in the list of artists of a fair.

# Solution
Fetch `followed_content` for the current `Fair` from MP which should return list of artists that current user follows and have works in this fair.
When showing the list concat the list ☝️ with all artists.

Fixes artist part of: https://artsyproduct.atlassian.net/browse/LD-108


src/lib/Scenes/Fair/Components/FairHeader/index.tsx
```diff
@@ -88,7 +88,7 @@ export class FairHeader extends React.Component<Props, State> {
       <>
         <EntityList
           prefix="Works by"
-          list={artistList}
+          list={fairfollowedArtistList.concat(artistList)}
```
kierangillen:
```var a1 = [1, 2, 3]
var a2 = [3, 2, 1]
a1.concat(a2)
(6) [1, 2, 3, 3, 2, 1]```

I dont think we want to concat the two arrays because there could be duplicate names and it could display the same name twice.

Comment (kierangillen) [reply src/lib/Scenes/Fair/Components/FairHeader/index.tsx]: Try something like this

```
const concatArray = fairfollowedArtistList.concat(artistList)
const uniqArray = concatArray.filter((v,i) => concatArray.indexOf(v) == i)

...

list={uniqArray}
```

Comment (kierangillen) [reply src/lib/Scenes/Fair/Components/FairHeader/index.tsx]: @orta Do you think this is heavy for the client? Should we move this to it's own MP field?

Comment (orta) [reply src/lib/Scenes/Fair/Components/FairHeader/index.tsx]: Nah, I think this level is fine enough - good call on making sure it's unique :

Comment (orta) [reply src/lib/Scenes/Fair/Components/FairHeader/index.tsx]: ( Mainly because I can see the usefulness in having two separate queries for this data )

Comment (ashkan18) [reply src/lib/Scenes/Fair/Components/FairHeader/index.tsx]: good catch!!! yes makes total sense @kierangillen

**Comment by kierangillen** [comment]
Merge on green

File: src/lib/Scenes/Fair/Components/FairHeader/index.tsx [hunk 2]
```diff
@@ -184,6 +185,18 @@ export const FairHeaderContainer = createFragmentContainer(
         artists
         partners
       }
+      followed_content {
```
ashkan18: we could limit this to 2

PR #1375, opened 2019-03-01
Display ShowCard when pin is tapped

l2succes:
kapture 2019-03-01 at 17 19 27

#skip_new_tests

### DIFF
--- a/src/lib/Scenes/Map/types.ts
+++ b/src/lib/Scenes/Map/types.ts
@@ -1,3 +1,5 @@
+import { GlobalMap_viewer } from "__generated__/GlobalMap_viewer.graphql"
+
 export interface Coordinates {
   lat: number
   lng: number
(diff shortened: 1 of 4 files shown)


kierangillen:
Nice!

src/lib/Scenes/Map/types.ts
```diff
@@ -12,3 +14,5 @@ export interface Tab {
   id: string
   text: string
 }
+
+export type Show = GlobalMap_viewer["city"]["shows"]["edges"][0]["node"]
```
kierangillen: Any chance this could be null?

l2succes: It could be null if we change the graphql fragment `GlobalMap_viewer` but that happens typescript will complain anyway so it's safe

l2succes:
Closing in favor of https://github.com/artsy/emission/pull/1379

PR #1376, opened 2019-03-01
City tabs styling

kierangillen:
- Show `save` button now spins when saving
- Small adjustments to `All` tab styling after QA with Jun and Owen

screen shot 2019-03-01 at 5 23 35 pm

screen shot 2019-03-01 at 5 23 30 pm

#skip_new_tests


l2succes:
merge on green

PR #1374, opened 2019-03-01
City picker ld 162

ashleyjelks:
This PR handles the UI and styling to support moving the CityPicker view controller to Objective-C

[Links to LD-162](https://artsyproduct.atlassian.net/browse/LD-162)

kapture 2019-03-01 at 15 15 35

#skip_new_tests

PR #1377, opened 2019-03-01
Shows city picker modal

ashfurrow:
Follows up on @ashleyjelks's work in #1374. 

This PR does a lot, here's the Cole's Notes:

- Moves the logic for user location and the city picker modal presentation to the `ARMapContainerViewController` class (in Objective-C) and documents why in that file.
- Adds the logic for switching cities.
- Adds the logic for handling user location denial, looking for a closest city, etc.
- Exposes the list of cities (in `cities.ts`) to Objective-C with new script/yarn task.
- Implements a really cool modal animation.

I'd like to get some tests in here for this, but this feature is blocking our ability to manually test LD, so I'm going to follow-up with tests later. 

Here's what it looks like:

## When the user accepts

initial good

## When the user declines

initial denied

## When changing cities

changing

Notice that changing has a considerable lag. We'll have to look into that further. 

Fixes LD-295, LD-296.

#skip_new_tests.


src/lib/Scenes/City/cities.ts
```diff
@@ -1,44 +1,44 @@
...
   },
-  berlin: {
-    name: "Berlin",
+  {
+    name: "Hong Kong",
     epicenter: {
-      lat: 52.52,
-      lng: 13.405,
+      lat: 22.3964,
+      lng: 114.1095,
```
ashfurrow: Looks like a merge conflict.

ashfurrow:
The `selectedCity` is no longer getting passed in, I can fix that.

Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -1,26 +1,55 @@
...
 {
     [super viewDidLoad];
 
+    self.locationManager = [[CLLocationManager alloc] init];
+    self.locationManager.delegate = self;
+    self.locationManager.desiredAccuracy = kCLLocationAccuracyThreeKilometers;
+    [self.locationManager requestWhenInUseAuthorization];
+
+    __weak typeof(self) sself = self;
+    [[NSNotificationCenter defaultCenter] addObserverForName:@"ARLocalDiscoveryOpenCityPicker" object:nil queue:nil usingBlock:^(NSNotification * _Nonnull note) {
```
l2succes: Only question is why are we using `NSNotificationCenter` if events will only be emitted from this class and child React Native view controllers. Can't we just rely on communication channels between native and react native. 

Calling an Objective-C method from Javascript: https://facebook.github.io/react-native/docs/native-modules-ios#content

ashfurrow: Unfortunately, our setup doesn’t make this easy (how would we call a method on this specific controller instance?). I confirmed on Slack that this is how we’d want to approach this (especially because of imminent RN bridge changes): https://artsy.slack.com/archives/C02BAQ5K7/p1551463932014200

l2succes: Right yeah that makes, very true the RN bridge API will change once we upgrade as well. Sounds good

src/lib/Scenes/City/CityPicker.tsx
```diff
@@ -1,85 +1,59 @@
...
     this.setState({ selectedCity: city })
-
-    // checks that comp is mounted before calling timeout method here to prevent timer
-    // from firing after component is unmounted
-    if (isCityPickerMounted) {
-      setTimeout(() => {
-        this.setState({ overlayVisible: false })
-      }, 1000)
-    }
+    NativeModules.ARNotificationsManager.postNotificationName("ARLocalDiscoveryUserSelectedCity", { cityIndex: index })
```
l2succes: I get the usage of notification center now, makes perfect sense

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -85,8 +96,27 @@ export class GlobalMap extends React.Component<Props, State> {
     )
   }
 
-  componentWillReceiveProps() {
+  componentWillReceiveProps(nextProps: Props) {
     this.emitFilteredBucketResults()
+    if (nextProps.hideMapButtons !== this.props.hideMapButtons) {
+      if (nextProps.hideMapButtons) {
+        this.moveButtons = new Animated.Value(0)
```
l2succes: I was hoping we could use `react-spring` for the animation like the following as I think it would simplify this but we can update it later
https://github.com/artsy/emission/blob/ccdf39004720c404c12824935c40cce053982fdb/src/lib/Scenes/Map/GlobalMap.tsx#L173-L197

ashfurrow: Yup, definitely makes sense! To be honest, this was my first React Native animation 😅 So I'm not surprised there's a better way. I'm going to be polishing the city picker today, I'll take a look into using `react-spring` instead.

PR #1379, opened 2019-03-03
Adds ability to cluster nearby shows

l2succes:
- Allows us to update the drawer position which is needed to collapse the drawer when the user taps a pin
- Converts shows data in GeoJSON format for consumption by the clustering engine

As of now, `react-native-mapbox-gl` doesn't expose an API to get points inside a cluster. I've attempted using supercluster, the underlying clustering library, to get the data but I still haven't gotten it to return the proper data. 

What we do have is the cluster coordinates and it's size, so another potential solution is sort coordinates by distance and select the N first points which will correspond to the points in the cluster. The downside of this solution is it would become more inefficient as the data size grows.

TODO:
- Handle tapping on a cluster which should show the following UI: https://zpl.io/29oy5Bw

**Screengrabs**

*Clustering Demo*

*Collapsing drawer*

#skip_new_tests


src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -230,6 +257,19 @@ export class GlobalMap extends React.Component<Props, State> {
               trackUserLocation: true,
             })
           }}
+          onPress={async event => {
+            this.setState({
+              activeShowID: null,
+            })
+            const { screenPointX, screenPointY } = event.properties
```
l2succes: The idea here is we use user's tap x and y values and the map zoom level to fetch the cluster from our cluster engine at which we map the points back to the shows

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -47,10 +60,16 @@ const ButtonAnimation = {
   },
 }
 
+enum DrawerPosition {
+  open = "open",
+  closed = "closed",
+  collapsed = "collapsed",
+  partiallyRevealed = "partiallyRevealed",
+}
```
ashfurrow: Shame that TypeScript can't infer the underly enum values à la Swift.

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -230,6 +257,19 @@ export class GlobalMap extends React.Component<Props, State> {
...
+              activeShowID: null,
+            })
+            const { screenPointX, screenPointY } = event.properties
+            const screenCoords = [screenPointX, screenPointY]
+            const features = await this.map.queryRenderedFeaturesAtPoint(screenCoords, ["==", "type", "Cluster"])
+            console.log(features)
+            const zoom = await this.map.getZoom()
+            const [x, y] = await this.map.getCoordinateFromView(screenCoords)
+            const data = this.clusterEngine.getTile(zoom, x, y)
+            console.log(data)
```
ashfurrow: Looks like a few `console.log` calls here.

src/lib/utils/convertCityToGeoJSON.ts
```diff
@@ -0,0 +1,37 @@
...
+    features: data.map(({ node }) => {
+      const {
+        coordinates: { lat, lng },
+      } = node.location
+
+      return {
+        type: "Feature",
+        geometry: {
+          type: "Point",
+          coordinates: [lng, lat],
```
ashfurrow: This is long/lat, which appears to be correct GeoJSON, but our codebase uses lat/long. I'd recommend including a comment so someone doesn't come in later and "fix" this 😄

orta: so weird, but yeah - GeoJSON spec denotes the other way around. Weird

anandaroop: Yeah this will always be weird until the end of time 

https://macwright.org/lonlat/

😭

l2succes: lol yeah I was getting confused just working on this, it's even more inconsistent than I thought lol

src/lib/Scenes/Map/GlobalMap.tsx
l2succes: Yup, that would be nice indeed

l2succes:
Alright thanks @ashfurrow, I addressed your comments. Gonna merge this one

PR #1380, opened 2019-03-04
Fixes render failure with missing city coords.

ashfurrow:
Follow-up to LD-295 / #1377. This is a workaround until LD–309 is addressed.

#skip_new_tests


ashleyjelks:
Looks good!

ashleyjelks:
Merge on green

PR #1382, opened 2019-03-04
Add followedContent to show query to get list of followed artists and…

ashkan18:
… show them first in the list of participating artists on show page

# Depends on
https://github.com/artsy/metaphysics/pull/1541


src/lib/Scenes/Show/Components/ShowHeader/ShowHeader.tsx
```diff
@@ -132,8 +132,11 @@ export class ShowHeader extends React.Component<Props, State> {
   render() {
     const { isFollowedSaving } = this.state
     const {
-      show: { artists, images, is_followed, name, partner, exhibition_period },
+      show: { artists, images, is_followed, name, partner, exhibition_period, followedArtists },
     } = this.props
+    const fairfollowedArtistList =
+      (followedArtists && followedArtists.edges && followedArtists.edges.map(fa => fa.node.artist)) || []
+    const uniqArtistList = [...new Set(fairfollowedArtistList.concat(artists))]
```
orta: this will remove the order, so you'll probably want something like `_.uniq`

ashkan18: good call! i'll update

kierangillen:
Merge on green

PR #1383, opened 2019-03-04
Ensures there's a consistent way to route to shows in Emission

orta:
We had three places with 2 different ways of routing - this consolidates them all and makes sure we have the entity business set up too. 

#trivial #skip_new_tests


orta:
Ah yeah, this also gets LD-301

ashfurrow:
Nice!

PR #1384, opened 2019-03-04
Remembers users' city selections

ashfurrow:
Fixes [LD-159](https://artsyproduct.atlassian.net/browse/LD-159). Also fixes a crash for shows with missing `cover_image` properties. Also fixes the problem where the checkmark wasn't being displayed in the city picker. 

#native_no_changes #skip_new_tests


ashleyjelks:
Looks good!

PR #1385, opened 2019-03-04
CitySectionList

kierangillen:
- Waiting on Metaphysics query
- Paired with @ashfurrow on this
- Adds `CitySectionList` and new view controller to All events page
- ( https://artsyproduct.atlassian.net/browse/LD-307 )

#skip_new_tests.


src/lib/Scenes/City/CitySectionList.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return <EventList key={name + section} bucket={shows} type={section} relay={this.state.relay} />
+  }
+}
+
+export default createFragmentContainer(
+  CitySectionList,
+  graphql`
+    fragment CitySectionList_city on City {
+      name
+      shows(discoverable: true, first: 50, sort: START_AT_ASC) {
```
kierangillen: This should pull all shows and needs to be passed the `section` param which corresponds to the type MUSEUM/GALLERY

ashfurrow: We're going to turn this into a pagination container, too.

ashfurrow:
Okay, I wrapped this up:

src/lib/Scenes/City/CitySectionList.tsx
```diff
@@ -0,0 +1,131 @@
...
+  {
+    city: graphql`
+      fragment CitySectionList_city on City
+        @argumentDefinitions(
+          count: { type: "Int", defaultValue: 10 }
+          cursor: { type: "String", defaultValue: "" }
+          partnerType: { type: "PartnerShowPartnerType" }
+        ) {
+        name
+        # TODO: Any way to exclude partnerType if it's not specified? When null it causes errors on MP.
```
ashfurrow: It's not a blocker on merging this PR, but I'd like to figure this out sometime. I couldn't find other examples of including an argument or not based on if it was present, so this might be something we need to change on MP (to not complain if `partnerType` is `null`. Thoughts @alloy?

ashfurrow: The error is from Gravity:

```
"https://stagingapi.artsy.net/api/v1/shows?at_a_fair=false&displayable=true&has_location=true&include_local_discovery=true&max_distance=25&near=51.51%2C-0.13&page=3&partner_types=&size=10&sort=start_at&status=current&total_count=true - {\"type\":\"param_error\",\"message\":\"Invalid parameters.\",\"detail\":{\"partner_types\":[\"does not have a valid value\"]}}"
```

So if we just don't send that from MP, I think that'd be the right choice.

alloy:
Nice @ashfurrow 👌 

One thing that’s missing is a spinner when loading more items, though, and perhaps it should start paging earlier? Not sure, that could just be because of your fast scrolling.

ashfurrow:
Yeah good call.

src/lib/Scenes/City/CitySectionList.tsx
alloy: I guess you could inject a boolean variable like `hasPartnerType` and use that in a conditional directive:

```graphql
... @include(if: $hasPartnerType) {
  shows(partnerType: $partnerType, …) {
    ... CitySectionList_shows
  }
}
... @skip(if: $hasPartnerType) {
  shows(…) {
    ... CitySectionList_shows
  }
}
```

Seems a little burdensome, so perhaps a MP change would lead to cleaner code.

ashfurrow: I think that accepting `null` would behave more accurately according to the schema, and would avoid a crash. I've opened a PR to metaphysics here and will remove the TODO comment: https://github.com/artsy/metaphysics/pull/1560

kierangillen:
Thanks @ashfurrow !!

PR #1386, opened 2019-03-05
Fetch all city shows and fairs

alloy:
Closes https://artsyproduct.atlassian.net/browse/LD-285

#trivial


src/lib/Components/Lists/SavedShowItemRow.tsx
```diff
@@ -86,7 +86,7 @@ export class SavedShowItemRow extends React.Component<Props, State> {
               },
             },
             updater: store => {
-              store.get(relayID).setValue(!isShowFollowed, "is_followed")
+              store.get(nodeID).setValue(!isShowFollowed, "is_followed")
```
ashkan18: oh nice! so we this `store...setValue` we don't need to refetch everything after save?

alloy: Yeah there is an imperative API to the Relay store.

PR #1387, opened 2019-03-05
Adds styling/sizing based on screen size to contents of CityPicker modal

ashleyjelks:
This PR handles the dynamic height and sizing of the City Picker modal content to support the different screen sizes in modern iPhone models (iPhone 5s and later).

[Links to  LD-314:](https://artsyproduct.atlassian.net/browse/LD-314)

screen shot 2019-03-05 at 11 17 09 am
screen shot 2019-03-05 at 11 52 13 am
screen shot 2019-03-05 at 10 46 09 am

#skip_new_tests

PR #1390, opened 2019-03-05
Fixes rounded corners on city picker component

ashfurrow:
~This PR is a WIP because it needs to be integrated with the changes in #1387 once that PR is merged. Otherwise it will look like this:~

This got messy (`rootView` vs `view`, etc) but the fix was just to take the safe area into account. React Native does this and now that we do too, our corners line up.

screen shot 2019-03-06 at 11 16 01 am

screen shot 2019-03-06 at 11 17 01 am

Fixes LD-319. #native_no_changes


orta:
Cool, yep - looks good to me, nice digging

PR #1391, opened 2019-03-05
LD-323 - All event section events now link to show

kierangillen:
- https://artsyproduct.atlassian.net/browse/LD-323


src/lib/Scenes/City/Components/Event/index.tsx
```diff
@@ -43,39 +45,48 @@ export class Event extends React.Component<Props, State> {
     console.log("handleSaveChange")
   }
 
+  handleTap = () => {
+    const {
+      node: { id },
+    } = this.props.event
+    SwitchBoard.presentNavigationViewController(this, `/show/${id}`)
```
orta: IMO, this syntax is more faffing (for not too much value) than just doing `SwitchBoard.presentNavigationViewController(this, `/show/${this.props.event.node.id}`)` - not a blocker, but this syntax is more overhead when reading (because you have to mentally parse backwards)

src/lib/Scenes/City/Components/EventSection/index.tsx
```diff
@@ -8,34 +8,37 @@ export interface Props {
...
+  render() {
+    const { data, title } = this.props
+    return (
+      <>
+        <Box my={2} px={2}>
+          <Serif size="8">{title}</Serif>
+        </Box>
+        {this.renderEvents()}
+        <Box px={2} mb={2}>
+          <CaretButton onPress={() => this.viewAllPressed()} text={`View all ${data.length} shows`} />
```
orta: we sure this can't be 1 show?

src/lib/Scenes/City/Components/Event/index.tsx
kierangillen: agreed

ashfurrow:
Going to merge this to get it into the next beta 👍

PR #1394, opened 2019-03-06
Adds BMW sponsoredContent to all events in city view

kierangillen:
- Adds BMW sponsoredContent to all events in city view


src/lib/Scenes/City/Components/BMWEventSection/index.tsx
```diff
@@ -1,57 +1,67 @@
...
+        <Box my={2} px={2}>
+          <Serif size="8">{this.props.title}</Serif>
+        </Box>
+        <Box mb={2} px={2}>
+          <Sans weight="medium" size="3">
+            Presented by BMW
+          </Sans>
+          <Box mt={1}>
+            <Serif size="3" color={color("black60")}>
+              {introText}
```
anandaroop: Yep, so implicit in the way this is all set up now is that these sponsored content text blurbs do not contain any markup. Judging from the comps that seems like a safe assumption for now, but wanted to call it out.

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -409,6 +410,10 @@ export const GlobalMapContainer = createFragmentContainer(
     fragment GlobalMap_viewer on Viewer @argumentDefinitions(citySlug: { type: "String!" }, maxInt: { type: "Int!" }) {
       city(slug: $citySlug) {
         name
+        sponsoredContent {
+          introText
+          artGuideUrl
+        }
```
anandaroop: 👍

src/lib/Scenes/City/Components/BMWEventSection/index.tsx
kierangillen: 👍

kierangillen:
@orta I dont see rules corresponding to this style, `Serif` with `color("black60")` -- maybe need to update the component? Or should we just not use the `Markdown` component for now?

orta:
Cool, sure, makes sense 👍

PR #1402, opened 2019-03-07
Adds new view controller and path to expose bmw art activations content

ashleyjelks:
This PR adds the BMW Art activations section the Fairs view

[Links to LD-131](https://artsyproduct.atlassian.net/browse/LD-131)

kapture 2019-03-06 at 22 56 52

#skip_new_tests


Pod/Classes/ViewControllers/ARFairBMWArtActivationComponentViewController.m
```diff
@@ -0,0 +1,29 @@
...
+#import "AREmission.h"
+#import "ARGraphQLQueryPreloader.h"
+
+@interface ARFairBMWArtActivationComponentViewController ()
+
+@end
+
+@implementation ARFairBMWArtActivationComponentViewController
+
++ (NSArray<ARGraphQLQuery *> *)preloadQueriesWithFairID:(NSString *)fairID;
```
ashfurrow: We can remove this method from this file and the header, actually. It's only really used in a few situations and doesn't really apply here.

src/lib/Scenes/Fair/Screens/FairBMWArtActivation.tsx
```diff
@@ -0,0 +1,147 @@
...
+          }
+          ItemSeparatorComponent={this.renderItemSeparator}
+          keyExtractor={(item, index) => item.type + String(index)}
+        />
+      </Theme>
+    )
+  }
+}
+
+export const FairBMWArtActivationRenderer: React.SFC<{ fairID: string }> = ({ fairID }) => (
```
ashfurrow: Usually, we put the query renderer component into the `QueryRenderers.tsx` file or somewhere separate from the component, and then use `createFragmentContainer` to wrap the plain React component (in this case, `FairBMWArtActivation`). This lets us separate out the query (`query FairBMWArtActivationQuery($fairID: String!) { ...`) from the data fragment.

Is there a reason we'd like to do things this way instead? I'm happy to pair as well 😄

src/lib/Scenes/Fair/Screens/FairBMWArtActivation.tsx
```diff
@@ -0,0 +1,147 @@
...
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import React from "react"
+import { FlatList, Image, ViewProperties } from "react-native"
+import { graphql, QueryRenderer } from "react-relay"
+import styled from "styled-components/native"
+import { defaultEnvironment } from "../../../relay/createEnvironment"
+import renderWithLoadProgress from "../../../utils/renderWithLoadProgress"
+
+interface Props extends ViewProperties {
+  fair: FairBMWArtActivationQuery["response"]["fair"]
```
ashfurrow: The `["response"]["fair"]` is a little confusing to me, but I think it's caused by the Relay issue I mentioned lower in this file. 

Also, I didn't know that TypeScript would let you refer to nested types with this subscript notation – very cool!

src/lib/Scenes/Fair/Screens/FairBMWArtActivation.tsx
l2succes: We should probably discuss this in our platform meeting on async in slack or a RFC as it's proposed change in how we organize our code. But `QueryRenderers.tsx` is getting fairly large (> 500 LOC) so we've been co locating QueryRenderers with their corresponding top level relay fragment / react components for our Local Discovery Component VCs.

I just synced with @ashleyjelks as she was having issues after moving the QueryRenderer, I suggested we leave it there and sync later on best practices around code organization.

ashfurrow: Sure, if this works for you.

PR #1405, opened 2019-03-07
Zero state for tabs and save to all tab

kierangillen:
- Adds zero state to `CityTab`
- Renames `SavedShowItemRow` to `ShowItemRow`
- Adds Save button functionality to button in `All` tab

## Example zero state
screen shot 2019-03-07 at 1 15 53 pm

- Note: This zero state needs `Message` in palette to accept weight, multiple lines, and centered text to more closely reflect design. Got OK from Owen for now on this.

#skip_new_tests #native_no_changes


src/lib/Scenes/City/Components/Event/index.tsx
```diff
@@ -1,26 +1,26 @@
...
 
 interface Props {
+  relay: RelayProp
   event: {
     node: {
       name: string
       __id: string
+      _id: string
       id: string
+      is_followed: boolean
```
alloy: @kierangillen There is no need to type all these typings manually, they map 1:1 to the query made in GlobalMap, right? So you should be able to get that typing like so (as if you were getting a property from a regular JS object):

```ts
import { GlobalMap_viewer } from "__generated__/GlobalMap_viewer.graphql"

interface Props {
  relay: RelayProp
  event: GlobalMap_viewer["city"]["shows"]["edges"][0]
}
```

Alternatively you can make use of a fragment to emit separate typings–and that might be a good step to making all these components actual Relay containers–but that’s not functionally going to be any different than the above. Nonetheless, ping me if you want to pair on that.

PR #1406, opened 2019-03-07
Sets up ScrollableTabView for City Guide

orta:
WIP, but now the tabs now use our infra. It's still slow, so I think the slowness probably comes from the fact that we set the active state of the tabs via the event emitter? Will take a look

screen shot 2019-03-07 at 3 03 28 pm

#native_no_changes


ashfurrow:
The CI failure is spurious, this is good to review+merge @l2succes

src/lib/Scenes/City/Components/EventList.tsx
```diff
@@ -81,6 +80,7 @@ export class EventList extends React.Component<Props> {
 
   render() {
     const { bucket } = this.props
+    // debugger
```
l2succes: stray comment

src/lib/Scenes/City/City.tsx
```diff
@@ -123,62 +132,63 @@ export class CityView extends Component<Props, State> {
...
-                ref={r => {
-                  if (r) {
-                    this.scrollView = r as any
-                  }
+              <ScrollableTabView
+                initialPage={this.props.initialTab || AllCityMetaTab}
+                onChangeTab={selectedTab => this.setSelectedTab(selectedTab)}
+                renderTabBar={props => (
+                  <>
+                    <TabBar {...props} />
```
l2succes: Can we use the same tab bar as the Saves & Follows VC https://github.com/artsy/emission/blob/master/src/lib/Scenes/Favorites/index.tsx#L65

src/lib/Scenes/City/Components/EventList.tsx
l2succes: stray comment

ashfurrow:
Yeah, we've got nested scroll views:

Screen Shot 2019-03-12 at 11 13 46 AM

ashfurrow:
Yup, so I confirmed that it's grabbing the _innermost_ scrollview instead of the outermost. If you grab the pointer address of `rnScrollView` from the VC, and look in Reveal, you'll see that the controller is pointing to the innermost one (the `FlatList`, which always has scroll disabled anyway). I'll clean up `FindFirstScrollView` to fix.

Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -30,15 +30,15 @@ @interface ARMapContainerViewController () <PulleyDelegate, PulleyDrawerViewCont
 @end
 
 static UIScrollView *
-FindFirstVerticalScrollView(UIView *view)
+FindFirstScrollView(UIView *view)
 {
     for (UIView *subview in view.subviews) {
-        if ([subview isKindOfClass:UIScrollView.class] && ([(UIScrollView *)subview contentSize].height > subview.frame.size.height)) {
```
ashfurrow: The issue here was that the outermost scroll view has a wide content size, at least at first. Since it contains the rest of the scroll views (now), this check wasn't necessary anymore.

ashfurrow:
@l2succes I've got those changes made, this should be ready for review again.

ashfurrow:
Here's a screenshot:

Screen Shot 2019-03-12 at 1 32 11 PM

I'd recommend following any other visual tweaks up in another PR, since this one has grown in size and duration.

src/lib/Scenes/Favorites/index.tsx
```diff
@@ -56,7 +56,7 @@ interface Props {
 class Favorites extends React.Component<Props, null> {
   render() {
     return (
-      <View style={{ flex: 1 }}>
+      <View style={{ flex: 1, backgroundColor: "red" }}>
```
l2succes: backgroundColor: "red"

l2succes:
👍 Yup agreed, let's get this one merged in.

src/lib/Scenes/City/City.tsx
```diff
@@ -103,16 +110,17 @@ export class CityView extends Component<Props, State> {
...
-    }
-
     if (this.state.buckets) {
       // We have the Relay response; post a notification so that the ARMapContainerViewController can finalize the native UI.
       NativeModules.ARNotificationsManager.postNotificationName("ARLocalDiscoveryQueryResponseReceived", {})
     }
   }
 
+  setSelectedTab(selectedTab) {
+    this.setState({ selectedTab: selectedTab.i }, this.fireScreenViewAnalytics)
```
alloy: @orta Do you recall why this was needed? No other code in this component seems to use it, but calling `setState` does lead to an extra render pass of this component.

orta: I took it directly from the homescreen - not sure if it's needed

alloy: Gotcha, thanks!

PR #1407, opened 2019-03-07
Fixes saving not updating props

kierangillen:


src/lib/Scenes/City/Components/AllEvents.tsx
```diff
@@ -28,11 +29,38 @@ export class AllEvents extends React.Component<Props, State> {
...
+        nextProps.buckets.museums.map(g => g.is_followed)
+      ) ||
+      !isEqual(
+        this.props.buckets.opening.map(g => g.node.is_followed),
+        nextProps.buckets.opening.map(g => g.is_followed)
+      ) ||
+      !isEqual(
+        this.props.buckets.galleries.map(g => g.node.is_followed),
+        nextProps.buckets.galleries.map(g => g.is_followed)
+      )
```
alloy: Seeing as these all look for the same prop, you could make this a bit DRYer by creating a function that takes the before and after props and the bucket key to compare.

l2succes: Yup you could do something like (double check bool logic for me)

```javascript
const shouldUpdate = ["saved", "closing", "museums", "opening", "closing"].map(key => {
  return !isEqual(
        this.props.buckets[key].map(g => g.node.is_followed),
        nextProps.buckets[key].map(g => g.is_followed)
      )
  )
}).some(a => a)

if (shouldUpdate) { this.updateSections(nextProps)) }
```

kierangillen: ah nice, ill update now.

l2succes: Actually even better
`Object.keys(this.props.buckets).map`

kierangillen: Problem with that is it would include the `fairs`, we don't want all the bucket keys

src/lib/Scenes/City/Components/AllEvents.tsx
```diff
@@ -28,11 +29,26 @@ export class AllEvents extends React.Component<Props, State> {
...
-    const { buckets, cityName } = this.props
+  componentWillReceiveProps(nextProps) {
+    const shouldUpdate = ["saved", "closing", "museums", "opening", "closing"]
+      .map(key => {
+        return !isEqual(
+          this.props.buckets[key].map(g => g.node.is_followed),
+          nextProps.buckets[key].map(g => g.is_followed)
+        )
+      })
+      .some(a => a)
```
alloy: This is not important at all, but as an FYI—replacing the `map` call with `some` would save on extra comparisons after a bucket did not pass the comparator.

l2succes: Oh that's true, even more efficient!

PR #1408, opened 2019-03-07
Sync city coordinates with Metaphysics

anandaroop:
This is a one-off change to bring Emission's city data inline with [Metaphysics' version](https://github.com/artsy/metaphysics/blob/master/src/schema/city/city_data.json) of the same.

A further enhancement would be to consolidate this source of truth via the build process, an item which is ticketed elsewhere.

#skip_new_tests

### DIFF
--- a/src/lib/Scenes/City/cities.ts
+++ b/src/lib/Scenes/City/cities.ts
@@ -3,24 +3,24 @@ export const cityList = [
     name: "New York",
     slug: "new-york-ny-usa",
     epicenter: {
-      lat: 40.7128,
-      lng: -74.006,
+      lat: 40.71,
+      lng: -74.01,
     },
   },
   {
     name: "Los Angeles",
     slug: "los-angeles-ca-usa",
     epicenter: {
-      lat: 34.0522,
-      lng: 118.2437,
+      lat: 34.05,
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/City/cities.ts
```diff
@@ -3,48 +3,48 @@ export const cityList = [
...
     },
   },
   {
     name: "Los Angeles",
     slug: "los-angeles-ca-usa",
     epicenter: {
-      lat: 34.0522,
-      lng: 118.2437,
+      lat: 34.05,
+      lng: -118.24,
```
anandaroop: This is the fix for the Los Angeles issue — sign needed to be flipped for decimal west longitude.

(The reduction in precision here and throughout is appropriate for city-scale mapping.)

ashfurrow: Good catch! Is there a name for this specific kind of bug? Where a latitude or longitude is accidentally inverted?

anandaroop: Not aware of any special name for this, just a good ol fashioned sign error 😄

l2succes:
Thanks for the fix @anandaroop!

PR #1409, opened 2019-03-07
Narrow iphone fix

ashleyjelks:
This PR handles narrow iphone screens so the Saved Shows text does not get truncated off the screen on smaller iphone devices. PR involved a refactor to the Flexbox code that contained the Saved Shows contents. 
screen shot 2019-03-07 at 5 09 33 pm
screen shot 2019-03-07 at 5 10 48 pm

#skip_new_tests

### DIFF
--- a/src/lib/Components/Lists/ShowItemRow.tsx
+++ b/src/lib/Components/Lists/ShowItemRow.tsx
@@ -9,7 +9,7 @@ import SwitchBoard from "lib/NativeModules/SwitchBoard"
 import { hrefForPartialShow } from "lib/utils/router"
 import { Schema, Track, track as _track } from "lib/utils/track"
 import React from "react"
-import { TouchableWithoutFeedback } from "react-native"
+import { Dimensions, TouchableWithoutFeedback } from "react-native"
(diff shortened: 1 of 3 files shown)


src/lib/Components/Lists/ShowItemRow.tsx
```diff
@@ -119,7 +121,7 @@ export class ShowItemRow extends React.Component<Props, State> {
               ) : (
                 <OpaqueImageView width={58} height={58} imageURL={imageURL} />
               )}
-              <Flex flexDirection="column" flexGrow="1" width="197">
+              <Flex flexDirection="column" flexGrow="1" width={isScreenNarrow ? width / 7 : 197}>
```
l2succes: Just trying to understand why we're hardcoding a width. Was flexbox not enough to dynamically resize the row?

ashleyjelks: If I don't set a width then the Gallery Infor container overflows to cut off the Save/Saved container:

screen shot 2019-03-08 at 9 10 27 am

l2succes:
merge on green

PR #1411, opened 2019-03-08
[EventEmitter] Add unsubscribing capabilities.

alloy:
Fixes https://artsyproduct.atlassian.net/browse/LD-375

#skip_new_tests

### DIFF
--- a/src/lib/Scenes/City/City.tsx
+++ b/src/lib/Scenes/City/City.tsx
@@ -42,31 +42,34 @@ export class CityView extends Component<Props, State> {
     { id: "museums", text: "Museums" },
   ]
 
+  handleEvent = ({
+    filter,
+    buckets,
+    cityName,
+    relay,
+    sponsoredContent,
+  }: {
+    filter: Tab
+    buckets: BucketResults
+    cityName: string
+    relay: RelayProp
+    sponsoredContent: { introText: string; artGuideUrl: string }
+  }) => {
+    this.setState({
+      buckets,
+      filter,
+      cityName,
+      relay,
+      sponsoredContent,
+    })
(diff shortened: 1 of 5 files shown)


src/lib/Scenes/City/City.tsx
```diff
@@ -42,31 +42,34 @@ export class CityView extends Component<Props, State> {
...
+    sponsoredContent: { introText: string; artGuideUrl: string }
+  }) => {
+    this.setState({
+      buckets,
+      filter,
+      cityName,
+      relay,
+      sponsoredContent,
+    })
+  }
```
alloy: Made this a bound method so that it will be the exact same instance when we subscribe and unsubscribe.

src/lib/Scenes/Map/EventEmitter.ts
```diff
@@ -1,19 +1,41 @@
-export const EventEmitter = {
-  _events: {},
-  dispatch(event, data) {
-    if (!this._events[event]) {
+type Callback = (data: any) => void
+
+type Event = "map:change" | "filters:change"
```
l2succes: Just a thought but could be interesting to use an enum instead
```typescript
enum Event {
  MapChanged = "map:change"
  FiltersChanged = "filters:change"
}
```

alloy: Ah yes, could be! For now I didn't want to change the API too much, as I’m not sure of the plans. Let me know if you feel strongly about doing so now.

l2succes: For sure, my reasoning as well. Let's leave it as is for now

PR #1412, opened 2019-03-08
Fixes scroll view getting stuck

ashfurrow:
This is a fix for [LD-321](https://artsyproduct.atlassian.net/browse/LD-321) and will require an update to Pulley, which I'm about to submit a PR for. I'll link back to here from it, but the important part of _this_ PR is that we are moving responsibility for enabling/disabling scrolling in the city view into ObjC; this decouples it from the React lifecycle, which is critical for the fix.

#skip_new_tests


orta:
This is great, because my current work in 
https://github.com/artsy/emission/pull/1406  is kinda blocked on some of this stuff as the nest ed scrollviews are pretty complicated

ashfurrow:
Okay, opened that PR here: https://github.com/l2succes/Pulley/pull/3

Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -60,6 +77,15 @@ - (void)viewDidLoad
         [sself.bottomSheetVC setDrawerPositionWithPosition:[PulleyPosition partiallyRevealed] animated:YES completion:nil];
         sself.initialDataIsLoaded = YES;
     }];
+    [[NSNotificationCenter defaultCenter] addObserverForName:@"ARLocalDiscoveryCityGotScrollView" object:nil queue:nil usingBlock:^(NSNotification * _Nonnull note) {
+            if (!self.rnScrollView) {
+                self.rnScrollView  = FindFirstVerticalScrollView(self.cityVC.view);
+                self.rnScrollView.scrollEnabled = NO;
+                if (self.rnScrollView) {
+                    [self.bottomSheetVC.drawerPanGestureRecognizer requireGestureRecognizerToFail:self.rnScrollView.panGestureRecognizer];
```
l2succes: 🙌🏾 amazing

Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -60,6 +77,15 @@ - (void)viewDidLoad
         [sself.bottomSheetVC setDrawerPositionWithPosition:[PulleyPosition partiallyRevealed] animated:YES completion:nil];
         sself.initialDataIsLoaded = YES;
     }];
+    [[NSNotificationCenter defaultCenter] addObserverForName:@"ARLocalDiscoveryCityGotScrollView" object:nil queue:nil usingBlock:^(NSNotification * _Nonnull note) {
+            if (!self.rnScrollView) {
```
alloy: Shouldn't this be a weak ref to `self`?

(Also, looks like the `sself` var above should be `wself`.)

Pod/Classes/ViewControllers/ARMapContainerViewController.m
alloy: Makes sense 👍

src/lib/Scenes/City/City.tsx
```diff
@@ -102,8 +102,10 @@ export class CityView extends Component<Props, State> {
               />
               <ScrollView
                 contentInset={{ bottom: bottomInset }}
-                onLayout={layout => (this.scrollViewVerticalStart = layout.nativeEvent.layout.y)}
-                scrollEnabled={isDrawerOpen}
+                onLayout={layout => {
+                  this.scrollViewVerticalStart = layout.nativeEvent.layout.y
+                  NativeModules.ARNotificationsManager.postNotificationName("ARLocalDiscoveryCityGotScrollView", {})
```
alloy: I wonder if this could get called too often and if moving this to the `ref` callback would work too?

alloy:
Good sleuthing, @ashfurrow! 👏

Pod/Classes/ViewControllers/ARMapContainerViewController.m
ashfurrow: Yup to both 👍

src/lib/Scenes/City/City.tsx
ashfurrow: I initially tried the `ref` callback but it's invoked before the scroll view has been added to the UIView hierarchy. I'll profile to see how often this gets called – my impression from the docs was that it would be fine, but let's verify.

ashfurrow: Yup, confirmed that this only gets called when switching tabs. I think that's acceptable.

alloy: Ah gotcha

alloy:
Merge on green

PR #1413, opened 2019-03-08
LD-377 - Fixes console error about textAlign

kierangillen:
- https://artsyproduct.atlassian.net/browse/LD-377
#trivial
#skip_new_tests

### DIFF
--- a/src/lib/Scenes/City/Components/CityTab.tsx
+++ b/src/lib/Scenes/City/Components/CityTab.tsx
@@ -3,12 +3,7 @@ import { ShowItemRow } from "lib/Components/Lists/ShowItemRow"
 import React from "react"
 import { FlatList } from "react-native"
 import { RelayProp } from "react-relay"
-import styled from "styled-components/native"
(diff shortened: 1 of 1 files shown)


src/lib/Scenes/City/Components/CityTab.tsx
```diff
@@ -46,19 +41,19 @@ export class CityTab extends React.Component<Props> {
       case "Saved":
         return (
           <Box py={2}>
-            <StyledMessage textSize="3t">{`You haven’t saved any shows in ${cityName}. When you save shows, they will show up here.`}</StyledMessage>
+            <Message textSize="3t">{`You haven’t saved any shows in ${cityName}. When you save shows, they will show up here.`}</Message>
```
alloy: So these are already centered?

kierangillen: No the component actually doesn't support centered text but got the OK from Owen that it's not necessary.

alloy: Okeydokey

PR #1415, opened 2019-03-08
Styles cards for clusters pins and clusters

l2succes:
**Cluster**

**Pin**

#skip_new_tests


orta:
nice, looks much better

kierangillen:
nice!

PR #1416, opened 2019-03-08
Remove cluster limits & runtime protection

orta:
This is only c384c45 but it contains #1414 because I needed those types

But I believe this fixes LD-350 

#trivial #native_no_changes #skip_new_tests

### DIFF
--- a/src/lib/Scenes/Map/GlobalMap.tsx
+++ b/src/lib/Scenes/Map/GlobalMap.tsx
@@ -86,11 +86,11 @@ enum DrawerPosition {
 
 export class GlobalMap extends React.Component<Props, State> {
   /** Makes sure we're consistently using { lat, lng } internally */
-  static coordArrayToLocation(arr: [number, number] | undefined) {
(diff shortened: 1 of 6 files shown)


src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -72,6 +85,19 @@ enum DrawerPosition {
 }
 
 export class GlobalMap extends React.Component<Props, State> {
+  /** Makes sure we're consistently using { lat, lng } internally */
+  static coordArrayToLocation(arr: [number, number] | undefined) {
```
anandaroop: Love the move towards being more explicit about this **eternal** problem 💯 

As long as you're doing so what about changing this to `lngLatArrayToLocation` since otherwise it's not _immediately_ clear what the order of the array is - `[lat, lng]` vs `[lng, lat]`.

orta: Good idea, done'd

PR #1417, opened 2019-03-08
[CityTab] Increase FlatList’s window size to avoid blank views.

alloy:
Fixes https://artsyproduct.atlassian.net/browse/LD-362

#skip_new_tests

### DIFF
--- a/src/lib/Scenes/City/Components/CityTab.tsx
+++ b/src/lib/Scenes/City/Components/CityTab.tsx
@@ -31,6 +31,7 @@ export class CityTab extends React.Component<Props> {
         keyExtractor={item => item.node.id}
         renderItem={({ item }) => this.renderItem(item)}
         scrollEnabled={false}
+        windowSize={50}
       />
     )
   }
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/City/Components/CityTab.tsx
```diff
@@ -31,6 +31,7 @@ export class CityTab extends React.Component<Props> {
         keyExtractor={item => item.node.id}
         renderItem={({ item }) => this.renderItem(item)}
         scrollEnabled={false}
+        windowSize={50}
```
alloy: An hour of faffing with Xcode and all I got was this 1 line change

orta: bah

alloy: To be fair, that was mostly just me catching up with the status quo in general and specifically on my iMac

orta: hab

l2succes: wow, one liner fixes 😩

PR #1418, opened 2019-03-08
Adds Fairs to the map

orta:
This shows everything for now, about to rebase.

#skip_new_tests

### DIFF
--- a/src/lib/Scenes/Map/Bucket.ts
+++ b/src/lib/Scenes/Map/Bucket.ts
@@ -1,15 +1,11 @@
 import { GlobalMap_viewer } from "__generated__/GlobalMap_viewer.graphql"
 import { filter } from "lodash"
 import moment from "moment"
-import { MapTab } from "./types"
 
-export type BucketKey = "saved" | "fairs" | "galleries" | "museums" | "closing" | "opening"
-export type BucketResults = { [key in BucketKey]: any[] }
-
-export const bucketCityResults = (viewer: GlobalMap_viewer): BucketResults => {
+export const bucketCityResults = (viewer: GlobalMap_viewer) => {
   const saved = filter(viewer.city.shows.edges, e => e.node.is_followed === true)
   const oneWeekFromNow = moment(new Date()).add(1, "week")
-  const fairs = (viewer.city.fairs.edges as unknown) as MapTab[]
+  const fairs = viewer.city.fairs.edges
   const galleries = filter(viewer.city.shows.edges, e => e.node.partner.type === "Gallery")
   const museums = filter(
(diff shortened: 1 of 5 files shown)


src/lib/Scenes/Map/Bucket.ts
```diff
@@ -1,15 +1,11 @@
 import { GlobalMap_viewer } from "__generated__/GlobalMap_viewer.graphql"
 import { filter } from "lodash"
 import moment from "moment"
-import { MapTab } from "./types"
 
-export type BucketKey = "saved" | "fairs" | "galleries" | "museums" | "closing" | "opening"
-export type BucketResults = { [key in BucketKey]: any[] }
-
-export const bucketCityResults = (viewer: GlobalMap_viewer): BucketResults => {
+export const bucketCityResults = (viewer: GlobalMap_viewer) => {
```
orta: OK, so I removed explicit typing all of the show here - all of the responses were correctly inferred in this function, so I could ditch the top one

src/lib/Scenes/Map/Bucket.ts
```diff
@@ -38,6 +34,9 @@ export const bucketCityResults = (viewer: GlobalMap_viewer): BucketResults => {
   }
 }
 
+export type BucketKey = keyof ReturnType<typeof bucketCityResults>
+export type BucketResults = ReturnType<typeof bucketCityResults>
+
```
orta: `ReturnType` has been my fav for a while, it's what powers the types for resolvers in MP. It can give a name to a inferred set of types. So the keys and the results are now entirely "dynamic" types - change the return value, and these will all change too 🎉

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -99,7 +100,8 @@ export class GlobalMap extends React.Component<Props, State> {
 
   map: Mapbox.MapView
   clusterEngine: Supercluster
-  featureCollection: any
+  showsGeoJSONFeatureCollection: MapGeoFeatureCollection
+  fairsGeoJSONFeatureCollection: MapGeoFeatureCollection
```
orta: There's a geocoding types library in our app, so if I have to faff more, I'll switch to those, that's why there's an `as any` coming up.

src/lib/Scenes/Map/types.ts
```diff
@@ -31,7 +32,7 @@ export interface MapGeoFeature {
     pitch: number
     animated: boolean
   }
-  type: string
+  type: string | any
```
orta: to match the geocoding lib also

PR #1420, opened 2019-03-09
Switch to includeStubShow instead of using discoverable

ashkan18:
This is follow up on https://github.com/artsy/metaphysics/pull/1552 where we removed `discoverable` and switched to `includeStubShow`, after this gets merged we are safe to remove `discoverable` from MP.

#trivial


orta:
I bet prettier is running on it, perhaps ashcan has global prettier settings in vsocde?

PR #1423, opened 2019-03-09
Map & city analytics v1.0

kierangillen:
- https://artsyproduct.atlassian.net/browse/LD-230
- Will need to revisit analytics once all city list views are enabled.
- I've marked a "y" next to rows that are finished in the spreadsheet, and all "green" rows are finished

#native_no_changes

### DIFF
--- a/src/lib/Scenes/Map/Components/ShowCard.tsx
+++ b/src/lib/Scenes/Map/Components/ShowCard.tsx
@@ -29,7 +29,6 @@ interface ShowCardProps {
   shows: ShowItemRow_show[]
   onSaveStarted?: () => void
   onSaveEnded?: () => void
-  onSave?: () => void
 }
 
 interface ShowCardState {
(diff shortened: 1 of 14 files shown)


src/lib/Scenes/Map/Components/ShowCard.tsx
```diff
@@ -29,6 +29,7 @@ interface ShowCardProps {
   shows: ShowItemRow_show[]
   onSaveStarted?: () => void
   onSaveEnded?: () => void
+  onSave?: () => void
```
l2succes: We don't need this anymore

src/lib/Scenes/Map/Components/ShowCard.tsx
```diff
@@ -52,7 +53,7 @@ export class ShowCard extends Component<ShowCardProps, ShowCardState> {
     currentPage: 1,
     isSaving: false,
   }
-
+  Background
```
l2succes: ?

l2succes:
merge on green

src/lib/Scenes/Map/Components/ShowCard.tsx
orta: Addressed, and the other commetn

PR #1425, opened 2019-03-09
Use relay validate on the CI

orta:
This means that not shipping the relay changes would fail the build.

#trivial


orta:
I wonder if this doesn't work with the typescript type emission instead. Closing.

alloy:
Hmm, I didn’t even know of `--validate`. Are you saying that perhaps there’s a limitation with the TS language plugin that would prohibit this from working? If so, can you file a ticket on the plugin repo?

alloy:
Bump

PR #1426, opened 2019-03-10
Support filtering the Shows/Fairs based on the current selected tab in the map

orta:
I originally had some unique functions for doing the filtering per-tab, but but I realized that the bucket holds the source of truth for this already. Instead the tabs for the map have the ability to declare where to find the map data for shows and for fairs.

Screen Shot 2019-03-09 at 7 32 34 PM

Screen Shot 2019-03-09 at 7 32 31 PM

Screen Shot 2019-03-09 at 7 32 28 PM


src/lib/AppRegistry.tsx
```diff
@@ -38,7 +38,7 @@ import {
 import FavoritesScene from "./Scenes/Favorites"
 import HomeScene from "./Scenes/Home"
 import { MapContainer } from "./Scenes/Map"
-import { BucketKey } from "./Scenes/Map/Bucket"
+import { BucketKey } from "./Scenes/Map/bucketCityResults"
```
orta: It wasn't a class or anything, so I renamed the file

src/lib/Scenes/City/City.tsx
```diff
@@ -20,7 +21,7 @@ interface Props {
 
 interface State {
   buckets?: BucketResults
-  filter: Tab
+  filter: MapTab
```
orta: was renamed for this file, made it consistent

src/lib/Scenes/City/City.tsx
```diff
@@ -62,14 +63,6 @@ export class CityView extends Component<Props, State> {
...
   scrollView: ScrollView = null
 
-  filters: Tab[] = [
-    { id: "all", text: "All" },
-    { id: "saved", text: "Saved" },
-    { id: "fairs", text: "Fairs" },
-    { id: "galleries", text: "Galleries" },
-    { id: "museums", text: "Museums" },
-  ]
-
```
orta: these all live in a separate place, which means we can set the default (and that it can be shared across the map and the city page

src/lib/Scenes/City/Components/AllEvents.tsx
```diff
@@ -37,10 +37,7 @@ export class AllEvents extends React.Component<Props, State> {
   componentWillReceiveProps(nextProps) {
     const shouldUpdate = ["saved", "closing", "museums", "opening", "closing"]
       .map(key => {
-        return !isEqual(
-          this.props.buckets[key].map(g => g.node.is_followed),
-          nextProps.buckets[key].map(g => g.is_followed)
-        )
+        return !isEqual(this.props.buckets[key].map(g => g.is_followed), nextProps.buckets[key].map(g => g.is_followed))
       })
```
orta: a rare case of 120 looking bad :D

src/lib/Scenes/City/Components/BMWEventSection/__tests__/index-tests.tsx
```diff
@@ -6,18 +6,16 @@ import { BMWEventSection } from "../index"
...
+    id: "pacita-abad-art-estate-palay-trapunto-murals-by-pacita-abad",
+    cover_image: {
+      url: "",
+    },
+    end_at: "2001-12-15T12:00:00+00:00",
+    start_at: "2001-11-12T12:00:00+00:00",
+    partner: {
+      name: "Pacita Abad Art Estate",
     },
   },
```
orta: de-nodifying, expected to see a lot of this

src/lib/Scenes/City/cityTabs.ts
```diff
@@ -0,0 +1,34 @@
...
+    getShows: bucket => bucket.galleries,
+    getFairs: _ => [],
+  },
+  {
+    id: "museums",
+    text: "Museums",
+    getShows: bucket => bucket.museums,
+    getFairs: _ => [],
+  },
+]
```
orta: maybe these could be renamed to clarify they're on the map? Unsure, not a blocker for me though

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -637,6 +638,8 @@ export const GlobalMapContainer = createFragmentContainer(
         fairs(first: $maxInt) {
           edges {
             node {
+              __typename
```
orta: I'm not actually using these

src/lib/utils/convertCityToGeoJSON.ts
```diff
@@ -21,31 +21,26 @@ import { MapGeoFeatureCollection } from "lib/Scenes/Map/types"
...
+  edges.map(node => {
     return {
-      node: {
-        ...node,
-        icon: "pin",
-      },
+      ...node,
+      icon: "pin",
     }
   })
```
orta: de-nodifying

l2succes:
Nice one @orta

PR #1434, opened 2019-03-11
Map better loading

l2succes:


Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -223,6 +230,18 @@ - (void)updateDrawerPosition:(NSString *)positionString
...
 }
 
+- (void)updateSafeAreaInsets
+{
+    UIEdgeInsets safeAreaInsets = self.view.safeAreaInsets;
+    [self.mapVC setProperty:@{
+                              @"top": [NSNumber numberWithFloat:safeAreaInsets.top],
+                              @"bottom": [NSNumber numberWithFloat:safeAreaInsets.bottom],
+                              @"left": [NSNumber numberWithFloat:safeAreaInsets.left],
+                              @"right": [NSNumber numberWithFloat:safeAreaInsets.right]
```
ashfurrow: Not a big deal, but we can use ObjC object literals here:

```diff
- @"right": [NSNumber numberWithFloat:safeAreaInsets.right]
+ @"right": @(safeAreaInsets.right)
```

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -338,14 +352,21 @@ export class GlobalMap extends React.Component<Props, State> {
     const { activeShows } = this.state
     const hasShows = activeShows.length > 0
 
+    // Check if it's an iPhone with ears (iPhone X, Xr, Xs, etc...)
+    const iPhoneHasEars = this.props.safeAreaInsets.top === 44
```
ashfurrow: Should we `=== 44` or do something like `> 20`? What if the ears get bigger, for example?

l2succes: yeah > 20 makes sense. Because it'll be `=== 20` on everything else

Pod/Classes/ViewControllers/ARMapContainerViewController.m
l2succes: True, forgot about that

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -338,14 +350,21 @@ export class GlobalMap extends React.Component<Props, State> {
     const { activeShows } = this.state
     const hasShows = activeShows.length > 0
 
+    // Check if it's an iPhone with ears (iPhone X, Xr, Xs, etc...)
+    const iPhoneHasEars = this.props.safeAreaInsets.top > 20
```
alloy: Top-notch joke 👌

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -54,11 +62,15 @@ interface Props {
...
+  relay?: RelayProp
   /** Tracking */
-  tracking: any
+  tracking?: any
+  /** city slug */
   citySlug: string
+  /** Whether the bottom sheet drawer is opened */
   isDrawerOpen?: boolean
+  /**  */
+  safeAreaInsets: SafeAreaInsets
```
alloy: These docs… 🙈

src/lib/Scenes/Map/GlobalMap.tsx
orta: :groan:

src/lib/Scenes/Map/GlobalMap.tsx
l2succes: 😅 Updated!

l2succes:
merge on green

PR #1435, opened 2019-03-11
Select pin ld67

ashleyjelks:
This PR adds UI changes to indicate when a user has selected a single show or cluster point in the City view.

[Links to LD-67](https://artsyproduct.atlassian.net/browse/LD-67)

Kapture 2019-03-11 at 1 04 54

### DIFF
--- a/src/lib/Scenes/Map/GlobalMap.tsx
+++ b/src/lib/Scenes/Map/GlobalMap.tsx
@@ -1,4 +1,4 @@
-import { Box, Flex, Theme } from "@artsy/palette"
+import { Box, color, Flex, Sans, Theme } from "@artsy/palette"
 import Mapbox from "@mapbox/react-native-mapbox-gl"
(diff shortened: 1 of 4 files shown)


src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -148,7 +152,7 @@ export class GlobalMap extends React.Component<Props, State> {
     singleShow: {
       iconImage: Mapbox.StyleSheet.identity("icon"),
       iconSize: 0.8,
-      iconOffset: [0, -21],
+      // iconOffset: [0, -21], @TODO: This property causes with the selected shows icon. Does this need to be here?
```
l2succes: We noticed the pins were off a little bit (not actually pointing to the right location) that's why this was added. we might have to offset the PointAnnotations by that value as well. Since it's a constant it shouldn't be too hard

PR #1436, opened 2019-03-11
[City] Use data from authoritative upstream source

alloy:
Part of https://artsyproduct.atlassian.net/browse/LD-335

Other side of https://github.com/artsy/metaphysics/pull/1566

#trivial


package.json
```diff
@@ -147,11 +148,10 @@
     "stylelint-config-styled-components": "^0.1.1",
     "stylelint-processor-styled-components": "^1.5.1",
     "ts-jest": "^22.0.2",
-    "ts-node": "^8.0.2",
     "tslint": "^5.11.0",
     "tslint-config-prettier": "^1.6.0",
     "tslint-plugin-relay": "^0.0.2",
-    "typescript": "^3.0.3",
+    "typescript": "^3.1.5",
```
alloy: This bump was needed to support `resolveJsonModule` when targeting `es2015` output: https://github.com/Microsoft/TypeScript/pull/26825

package.json
```diff
@@ -147,11 +148,10 @@
     "stylelint-config-styled-components": "^0.1.1",
     "stylelint-processor-styled-components": "^1.5.1",
     "ts-jest": "^22.0.2",
-    "ts-node": "^8.0.2",
```
alloy: No longer needed now that we just load JSON.

alloy:
Merge on green

orta:
You got some TypeScript errors - likely from the version bump improving inferrence

src/lib/Components/Bidding/Screens/BidResult.tsx
```diff
@@ -49,7 +49,7 @@ export class BidResult extends React.Component<BidResultProps> {
     }
 
     // fetch the latest increments for the select max bid screen
-    if (this.props.refreshSaleArtwork()) {
+    if (this.props.refreshSaleArtwork) {
```
alloy: The TS upgrade led to identifying this little bug fix, although I didn’t find any errors in Sentry, so presumably this was never undefined. However, the callback would get invoked twice, which may or may not have been an issue.

PR #1442, opened 2019-03-11
Fixes missing safe insets #trivial

ashfurrow:
Fixes the following error on non-iOS 11+ devices:

Screen Shot 2019-03-11 at 4 02 48 PM

#native_no_changes.

### DIFF
--- a/Pod/Classes/ViewControllers/ARMapComponentViewController.m
+++ b/Pod/Classes/ViewControllers/ARMapComponentViewController.m
@@ -4,7 +4,14 @@
 
 - (instancetype)init
 {
-    return [super initWithEmission:nil moduleName:@"Map" initialProperties:nil];
+    return [super initWithEmission:nil moduleName:@"Map" initialProperties:@{
+        @"safeAreaInsets": @{
+            @"top": @(0),
+            @"left": @(0),
+            @"bottom": @(0),
+            @"right": @(0)
+        }
+    }];
 }
 
 - (UIStatusBarStyle)preferredStatusBarStyle


Pod/Classes/ViewControllers/ARMapComponentViewController.m
```diff
@@ -4,7 +4,14 @@ @implementation ARMapComponentViewController
...
 {
-    return [super initWithEmission:nil moduleName:@"Map" initialProperties:nil];
+    return [super initWithEmission:nil moduleName:@"Map" initialProperties:@{
+        @"safeAreaInsets": @{
+            @"top": @(0),
+            @"left": @(0),
+            @"bottom": @(0),
+            @"right": @(0)
+        }
+    }];
```
orta: Maybe this can live inside the VC - inside an else on this func: https://github.com/artsy/emission/pull/1434/files#diff-a6663cfe9d0e55272a71b62f28e85090R234

ashfurrow: I'd prefer to have it from the get-go. The `Props` define it as a required property and this decouples the React component's requirements from our ObjC code (what if `updateSafeAreaInsets` isn't called by the time the first render occurs?).

orta: 👍

ashfurrow: But I can see it either way, since `updateSafeAreaInsets` looks pretty critical to the UI anyway. Yeah, let's do that so our support for older devices is explicit 👍

PR #1445, opened 2019-03-11
LD-381 - Updates to bucket queries and sorts

kierangillen:
- ( https://artsyproduct.atlassian.net/browse/LD-381 )
- Worked with @anandaroop on this PR
- Removes PartnerShows from queries
- Adds correct sort to city and museum tabs/lists/bricks
- Adds upcomingShows to globalMap query
- Shows upcoming saved shows on map


src/lib/relay/QueryRenderers.tsx
```diff
@@ -492,22 +492,29 @@ export const CitySectionListRenderer: React.SFC<CitySectionListProps> = ({ rende
...
     case "closing":
       variables.status = "CLOSING_SOON"
-      variables.dayThreshold = 7
+      variables.sort = "END_AT_ASC"
+      variables.dayThreshold = 30
       break
     case "opening":
       variables.status = "UPCOMING"
-      variables.dayThreshold = 7
+      variables.sort = "START_AT_ASC"
```
ashfurrow: I don't think this `sort` parameter does anything, since there's no `sort` parameter defined in the GraphQL query. We'll need to add it there for this to work.

kierangillen: hey yeah these were all just quick notes essentially on what we need to do. I added it in another commit. About to push it up, but still working on some other small details.

ashfurrow: Still no sort param on `CitySectionList_city`?

kierangillen: hey @ashfurrow it's been added now

ashfurrow:
Resolve these conflicts locally and pushed.

kierangillen:
🎉

PR #1447, opened 2019-03-11
Add ability to pre-heat GraphQL cache and include cached City data

alloy:
Closes https://artsyproduct.atlassian.net/browse/LD-286


Pod/Classes/GraphQL/ARGraphQLQueryCache.m
```diff
@@ -61,7 +107,13 @@ - (instancetype)init;
...
 static NSString *
 CacheKey(NSString *queryID, NSDictionary *variables) {
-    NSData *variablesJSONData = [NSJSONSerialization dataWithJSONObject:variables options:0 error:nil];
+    // THe variables dictionary may come from anywhere and that may influence the order in many ways,
+    // as such we apply some sorting to generate a stable key.
+    NSArray *sortedKeys = [[variables allKeys] sortedArrayUsingSelector:@selector(compare:)];
+    NSArray *sortedValues = [variables objectsForKeys:sortedKeys notFoundMarker:@""];
+    NSArray *sortedVariables = [sortedKeys arrayByAddingObjectsFromArray:sortedValues];
+
+    NSData *variablesJSONData = [NSJSONSerialization dataWithJSONObject:sortedVariables options:0 error:nil];
```
alloy: This drove me mad for a good 15 minutes 😄

scripts/generate-cities-cache.js
```diff
@@ -0,0 +1,76 @@
...
+const TTL = 3600 // 1 hour
+
+const MAX_GRAPHQL_INT = 2147483647
+
+const assetsDir = path.resolve(__dirname, "../Pod/Assets/PreHeatedGraphQLCache")
+const queryData = queryMap()[QUERY_NAME]
+
+cities.forEach(city => {
+  const queryParams = {
+    // We don't use the persisted ID here because the queries may not actually have been persisted yet.
```
alloy: This means we can safely run it _just_ after generating the query map, i.e. before pushing the queries to MP.

Pod/Classes/GraphQL/ARGraphQLQueryCache.m
ashfurrow: I bet!

Pod/Classes/GraphQL/ARGraphQLQueryCache.m
```diff
@@ -88,6 +140,14 @@ - (instancetype)init;
     return [validUntil compare:[NSDate date]] == NSOrderedDescending;
 }
 
+static NSDate *
```
l2succes: @alloy is there an advantage to using static methods vs instance ones?

ashfurrow: These methods are ultimately getting called from `+load`, which is a class method that's invoked when the Objective-C runtime is loaded. There is no instance to call these on (or, if you prefer, the class itself _is_ the instance).

alloy: Yeah besides it not really being possible in this case, as @ashfurrow said, there’s also the issue that these functions don’t really rely on any instance state (they are stateless). In those case I prefer to move a method out of a class into a function.

Finally, static functions will likely be inlined by the compiler whereas methods are dynamic and will never be inlined. Not that that optimization is really needed here, just one extra pro for static functions when all else is equal.

l2succes: Got it, so it's essentially another way to implement the singleton pattern. Thanks @ashfurrow

ashfurrow: Yes exactly – you can think of Objective-C classes / class methods as singleton objects. Swift takes this a step further: every instance method is _literally_ the return value of an identically-named class method that takes a `self` argument and returns a curried function (the instance method).

orta: This wa a good thread, thanks for asking @l2succes

PR #1448, opened 2019-03-11
Skips native CI when not files have been changed #trivial

ashfurrow:
We discussed this at some point, @orta, but I can't remember when/where. This PR adds a new script (that does _not_ have any dependencies, so it can be run before `yarn install`) and invokes it during native CI runs. It uses bash short-circuiting to skip running tests on builds that have no native changes.


orta:
Yep - nice thinking @ashfurrow  - merge on green

ashfurrow:
Some weird bash syntax stuff, but I think it'll work this time!

ashfurrow:
Finished in `1:02`, not bad!

PR #1455, opened 2019-03-12
[GraphQL] Add freshness to pre-heated cache.

alloy:
Part of https://artsyproduct.atlassian.net/browse/LD-286.

Adds freshness to GraphQL pre-heated cache after which it won’t be used anymore–which is 1 month, from the time of cutting the release, for City.

#native_no_changes


scripts/preheatGraphQLCache.js
```diff
@@ -0,0 +1,75 @@
...
+    })
+  })
+
+  req.on("error", error => {
+    throw error
+  })
+
+  req.write(requestPostData)
+  req.end()
+}
```
alloy: Made this generic so it can be used for other data in the future.

scripts/generate-cities-cache.js
```diff
@@ -1,76 +1,33 @@
...
 const cities = require("../data/cityDataSortedByDisplayPreference.json")
 
 const QUERY_NAME = "MapRendererQuery"
-const TTL = 3600 // 1 hour
-
 const MAX_GRAPHQL_INT = 2147483647
+const TTL = 3600 // 1 hour
+const FRESHNESS = moment()
+  .add(1, "month")
+  .toDate()
```
alloy: For City data we set the freshness to be 1 month [from the time we cut the release].

alloy:
Remind me when I should bump the native code version? There’s no change in external API, the freshness will just be ignored by older builds.

ashfurrow:
@alloy we only need a bump when the native interface that eigen uses changes. @orta and I were discussing today, and the Danger check could be modified to only look for header file changes.

scripts/preheatGraphQLCache.js
orta: cool

PR #1456, opened 2019-03-12
Move set state to callback

ashleyjelks:
About this PR:

To handle showing and removing the checkmark icon from a city name on the City, we save a selectedCities field to state.  

This state is updated twice when a user interacts with the modal: once to show the icon on the selected city, and then again after the touch event to clear the icon and remove the modal. 

Since this.setState is an asynchronous function, I am passing the second call to it as a callback to be called after the first call completes. 

#trivial

### DIFF
--- a/src/lib/Scenes/City/CityPicker.tsx
+++ b/src/lib/Scenes/City/CityPicker.tsx
@@ -32,11 +32,14 @@ export class CityPicker extends Component<Props, State> {
     }
   }
 
-  selectCity(city: string, index: number) {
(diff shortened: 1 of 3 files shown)


orta:
Does this fix the need to double tap on a show>

ashleyjelks:
> Does this fix the need to double tap on a show?

Not sure I follow...you mean you have to double tap a map pin to get the Show card and purple icon map pin to render??

orta:
Yeah that's the bug, which I guess this doesn't fix that then - no problem

PR #1458, opened 2019-03-12
Fix the changelog

orta:
So we get back to having versions


orta:
forcing the merge early because I keep needing to rebase

PR #1460, opened 2019-03-12
Removes user location button when user hasn't supplied location 

ashfurrow:
Fixes [LD-411](https://artsyproduct.atlassian.net/browse/LD-411). This doesn't actually need the CoreLocation API, since we can rely on `this.state.userLocation` being null if the map hasn't provided us with the location (because it can't).


src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -535,15 +535,17 @@ export class GlobalMap extends React.Component<Props, State> {
...
-              <Box style={{ marginLeft: "auto" }}>
-                <UserPositionButton
-                  highlight={this.state.userLocation === this.state.currentLocation}
-                  onPress={() => {
-                    const { lat, lng } = this.state.userLocation
-                    this.map.moveTo([lng, lat], 500)
-                  }}
-                />
-              </Box>
+              {this.state.userLocation && (
```
kierangillen: Maybe add a default `null` state for `userLocation`?

ashfurrow: It's an optional in `State`, I think `undefined` makes sense as a default?

PR #1461, opened 2019-03-12
Adds header text

kierangillen:
- Styles city show list views and adds header text

Screen Shot 2019-03-12 at 2 55 51 PM

#trivial


src/lib/Scenes/City/CitySavedList.tsx
```diff
@@ -47,18 +47,16 @@ class CitySavedList extends React.Component<Props, State> {
...
+          header="Saved shows"
+          key={name + "save"}
+          cityName={name}
+          bucket={shows.edges.map(e => e.node) as any}
+          type={"saved"}
+          relay={relay}
+          onScroll={isCloseToBottom(this.fetchData)}
+          fetchingNextPage={fetchingNextPage}
+        />
       </Theme>
```
orta: This may not need the theme?

orta: Ah, the one below adds it, so it probably does

src/lib/Scenes/City/Components/AllEvents.tsx
```diff
@@ -92,7 +92,6 @@ export class AllEvents extends React.Component<Props, State> {
     if (buckets.museums && buckets.museums.length) {
       sections.push({
         type: "bmw",
-        data: buckets.museums,
```
ashfurrow: Ah, good catch.

kierangillen:
Something strange is happening on this branch where the components are not rendering, creating a new PR

PR #1462, opened 2019-03-12
Error screen

ashfurrow:
This is just a first stab, lots to clean up. 

Fixes LD-318.


src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -73,6 +73,11 @@ interface Props {
   isDrawerOpen?: boolean
   /** Reflects the area not covered by navigation bars, tab bars, toolbars, and other ancestors  */
   safeAreaInsets: SafeAreaInsets
+  /** Error from Relay (MapRenderer.tsx). Needed here to send over the EventEmitter. */
+  relayError?: {
+    error: Error
+    retry: () => void
+  }
```
alloy: I'm sure you were already planning to do so, but just in case; create a named interface for this and use that in all places you are typing this out.

ashfurrow: Yeah definitely – was going to put it in `types.ts`.

ashfurrow:
This is still a WIP (the React error and missing tabs) but it's really coming together:



I test this by pointing Emission to `localhost` for MP and applying this diff:

```diff
diff --git a/src/lib/fetchPersistedQuery.ts b/src/lib/fetchPersistedQuery.ts
index 4460a3d8..652cb10b 100644
--- a/src/lib/fetchPersistedQuery.ts
+++ b/src/lib/fetchPersistedQuery.ts
@@ -4,6 +4,17 @@ import { RequestHandler } from "express"
 
 export const fetchPersistedQuery: RequestHandler = (req, res, next) => {
   const { documentID } = req.body
+
+  return setTimeout(() => {
+    console.log("yeah!")
+    res
+      .send({
+        data: null,
+        errors: [{ error: "this is an error for testing" }],
+      })
+      .end()
+  }, 1500)
+
   if (documentID) {
     const query = queryMap[documentID]
     if (query) {
```

src/lib/Scenes/Map/MapRenderer.tsx
```diff
@@ -34,14 +35,35 @@ export const MapRenderer: React.SFC<{
...
+                isRetrying,
+              }}
+            />
+          )
+        } else if (isRetrying) {
+          isRetrying = false
+          return <GlobalMap {...computedProps} viewer={null} relayErrorState={{ isRetrying: true }} />
+        } else if (mapProps || props.initialCoordinates) {
+          return <GlobalMap {...computedProps} />
+        } else {
```
ashfurrow: I need to come back and comment some of this – I had to learn a lot about how query renderers work and would like to capture that knowledge.

ashfurrow:
Okay, this is done:

2019-03-13 17-42-41 gifcask 2019-03-13 17_43_20

The button is doing this weird thing where it stays purple (after the spinner goes away). But it's the standard button we use everything, so maybe someone has an idea of what's up.

ashfurrow:
@l2succes This deviates from the designs slightly, where we don't render the tab labels. Those are dependent on rendering the actual tabs, which we can't without data. Figured it'd be fine but just a head's up.

ashfurrow:
Oh, actually this isn't rendering tabs at all now 😢 
Simulator Screen Shot - iPhone X - 2019-03-13 at 17 49 21

src/lib/Scenes/City/City.tsx
```diff
@@ -206,3 +219,18 @@ const Handle = styled.View`
   border-radius: 2.5px;
   background-color: ${color("black30")};
 `
+
+const ErrorScreen: React.SFC<{ relayErrorState: RelayErrorState }> = ({ relayErrorState: { retry, isRetrying } }) => {
+  return (
+    <Box py={2}>
+      <Sans size="3t" textAlign="center" mx={2}>
+        We are having trouble loading content right now, please try again later.
```
orta: I dunno the spec, but "the content" feels a bit meh - "the city guide" feels a bit better

src/lib/Scenes/Map/Components/CitySwitcherButton.tsx
```diff
@@ -72,6 +72,6 @@ export class CitySwitcherButton extends Component<Props> {
           )}
         </Background>
       </TouchableWithoutFeedback>
-    )
+    ) : null
```
orta: Aside: This could be a 2nd `||` instead of the ternary, false won't render in a JSX tree

src/lib/Scenes/Map/EventEmitter.ts
```diff
@@ -1,6 +1,6 @@
 type Callback = (data: any) => void
 
-type Event = "map:change" | "filters:change"
+type Event = "map:change" | "map:error" | "filters:change"
```
orta: nice, yah

orta:
Gonna handle those conflicts and merge manually into master

orta:
Nah, I'll verify on CI - merge on green

src/lib/Scenes/Map/Components/CitySwitcherButton.tsx
ashfurrow: Ah yeah, that makes sense. So you could get a kind of null-coalescing component tree, like `??` in Swift.

src/lib/Scenes/City/City.tsx
ashfurrow: Yeah, let's get feedback on that in peace room tomorrow.

alloy:
Nice one, @ashfurrow 👌

PR #1463, opened 2019-03-12
Make sure the query happens so we show 3 artists instead of 2 in an entity list for fairs/shows

orta:
Before:
Screen Shot 2019-03-12 at 3 19 42 PM

After:
Screen Shot 2019-03-12 at 3 27 26 PM

We had code for this, but not the query changes to ask for 3. #trivial


ashleyjelks:
Looks good!

PR #1464, opened 2019-03-12
Styles show lists

kierangillen:
- Styles city show list

Screen Shot 2019-03-12 at 2 55 51 PM

#trivial


l2succes:
@kierangillen where does this back button come from?
@ashfurrow I forget where we left off with that

ashfurrow:
We punted on the back button, so once integrated, this view will show Eigen’s standard back button.

PR #1465, opened 2019-03-12
removes save button from Show Card component

ashleyjelks:
Screen Shot 2019-03-12 at 4 35 17 PM

Screen Shot 2019-03-12 at 4 33 11 PM


orta:
This is likely to also affect the lists of shows I think too

ashleyjelks:
> This is likely to also affect the lists of shows I think too

Yep...

ashleyjelks:
> This is likely to also affect the lists of shows I think too

This is addressed here #1479  @orta

PR #1466, opened 2019-03-12
 [Map] Make user indicator purple.

alloy:
Fixes https://artsyproduct.atlassian.net/browse/LD-439

Simulator Screen Shot - iPhone 8 - 2019-03-12 at 21 52 23

#native_no_changes


Pod/Classes/ViewControllers/ARMapComponentViewController.m
```diff
@@ -1,7 +1,13 @@
 #import "ARMapComponentViewController.h"
+#import <react-native-mapbox-gl/RCTMGLMapView.h>
 
 @implementation ARMapComponentViewController
 
++ (void)initialize {
+    UIColor *purple100 = [UIColor colorWithRed:110.0f/255.0f green:30.0f/255.0f blue:255.0f/255.0f alpha:1.0f];
+    [[RCTMGLMapView appearanceWhenContainedInInstancesOfClasses:@[self]] setTintColor:purple100];
```
l2succes: Ah! it's `UIAppearance` compatible, nice

alloy:
Ah, hah, I had already added the danger ignore rule. This code change doesn’t really require Eigen to change in any way so technically no need to restrict PRs that people can try out from the beta.

alloy:
(I’m good either way, if you don’t think existing testers will suffer from that.)

l2succes:
Yup that's true

PR #1468, opened 2019-03-12
Hides fair sections if there are no partners, artists, or artworks

kierangillen:
- Hides fair sections if there are no partners, artists, or artworks

#trivial


ashfurrow:
I'll merge locally to avoid another CI run.

PR #1472, opened 2019-03-13
Adds BMW Links

ashleyjelks:
LD-420

Kapture 2019-03-13 at 11 02 33

#native_no_changes


src/lib/Scenes/City/Components/SavedEventSection/index.tsx
```diff
@@ -17,6 +17,10 @@ export class SavedEventSection extends Component<any> {
     SwitchBoard.presentNavigationViewController(this, `/city-save/${this.props.citySlug}`)
   }
 
+  navigateToBMWArtGuide() {
+    SwitchBoard.presentNavigationViewController(this, "https://www.bmw-arts-design.com/bmw_art_guide")
```
anandaroop: This url is available from MP as:

```graphql
{
  city(slug: "your-slug-here") {
    name
    sponsoredContent {
      artGuideUrl
    }
  }
}
```

src/lib/Scenes/City/CityPicker.tsx
```diff
@@ -63,6 +64,10 @@ export class CityPicker extends Component<Props, State> {
     )
   }
 
+  navigateToBMWArtGuide() {
+    SwitchBoard.presentNavigationViewController(this, "https://www.bmw-arts-design.com/bmw_art_guide")
```
anandaroop: See other comment

src/lib/Scenes/City/Components/SavedEventSection/index.tsx
ashfurrow: We do already include this in the `GlobalMap` query, so it should be accessible there. The CityPicker is trickier because of how we present our components. @ashleyjelks can you ping me and we'll pair on it?

src/lib/Scenes/City/Components/SavedEventSection/index.tsx
```diff
@@ -4,19 +4,28 @@ import PinSavedOff from "lib/Icons/PinSavedOff"
...
 export class SavedEventSection extends Component<any> {
   handleTap = () => {
     SwitchBoard.presentNavigationViewController(this, `/city-save/${this.props.citySlug}`)
   }
 
+  navigateToBMWArtGuide() {
+    const { sponsoredContentUrl } = this.props
+    SwitchBoard.presentNavigationViewController(
+      this,
+      sponsoredContentUrl || "https://www.bmw-arts-design.com/bmw_art_guide"
```
ashfurrow: Since the `SavedEventSection` _always_ gets its data, we can drop the `|| "https:..."` here because `sponsoredContentUrl` will never be `nil`.

I like checking for `nil`, though! Good to guard against error cases. Can't wait to enable some more of those TypeScript flags... /cc @alloy

orta:
👍

PR #1473, opened 2019-03-13
Only ever show one city picker at a time

ashfurrow:
Fixes [LD-443](https://artsyproduct.atlassian.net/browse/LD-443). It also disables user interaction on the city picker until its presentation animation is complete. #native_no_changes.

### DIFF
--- a/Pod/Classes/ViewControllers/ARMapContainerViewController.m
+++ b/Pod/Classes/ViewControllers/ARMapContainerViewController.m
@@ -154,6 +154,10 @@ Since this controller already has to do the above logic, having it handle the Ci
 
 - (void)showCityPicker
 {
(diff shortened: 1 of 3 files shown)


Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -154,13 +154,18 @@ - (void)userSuppliedLocation:(CLLocation *)userLocation
 
 - (void)showCityPicker
 {
+    if (self.cityPickerContainerView) {
```
l2succes: cityPickerContainerView is set to `nil` here
https://github.com/artsy/emission/pull/1473/files#diff-a6663cfe9d0e55272a71b62f28e85090R219

PR #1474, opened 2019-03-13
Switch to fork of react-native-mapbox-gl 

l2succes:
We noticed a bug with tap events being captured every other tap in `RCTMGLMapViewManager` with the following method. 

https://github.com/nitaliano/react-native-mapbox-gl/blob/master/ios/RCTMGL/RCTMGLMapViewManager.m#L38-L66

The fix was to remove an unused `UITapGestureRecognizer` meant to recognize double tap events.
I created a fork of the library to fix this and will create a PR upstream as it's affecting all consumers of the library.


orta:
Cool, yep, merge on green

PR #1477, opened 2019-03-13
Update scrollable tab bar

l2succes:
This updates the styling of the tab bar and begins addressing animation performance issues on the ScrollableTabView. 

TODO:
- [x] fix horizontal scrolling

Before:

After:

### DIFF
--- a/Pod/Classes/ViewControllers/ARMapContainerViewController.m
+++ b/Pod/Classes/ViewControllers/ARMapContainerViewController.m
@@ -34,7 +34,12 @@ FindFirstScrollView(UIView *view)
 {
     for (UIView *subview in view.subviews) {
         if ([subview isKindOfClass:UIScrollView.class]) {
-            return (UIScrollView *)subview;
+            CGSize size = [(UIScrollView *)subview contentSize];
+             // The tab bar on the City guide is a scrollview, so we need to make sure we hit
+            //  the main scrollview instead.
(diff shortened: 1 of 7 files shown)


orta:
Looks good, gonna merge manually to avoid the changelog merge.

Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -34,7 +34,12 @@ @interface ARMapContainerViewController () <PulleyDelegate, PulleyDrawerViewCont
 {
     for (UIView *subview in view.subviews) {
         if ([subview isKindOfClass:UIScrollView.class]) {
-            return (UIScrollView *)subview;
+            CGSize size = [(UIScrollView *)subview contentSize];
+             // The tab bar on the City guide is a scrollview, so we need to make sure we hit
+            //  the main scrollview instead.
+            if (size.height > size.width ) {
+                return (UIScrollView *)subview;
+            }
```
alloy: The name of this function doesn’t reflect the change, it should probably be called `FindMapScrollView` or something. I would also prefer to look for known classes to bail on, e.g. the main map view class, rather than this ambiguous check.

l2succes: 👍 I'll update the naming in a subsequent PR

PR #1478, opened 2019-03-13
Switch to opening hours for fairs and shows

ashkan18:
Made it WIP since i haven't tried it yet on the device. But the idea is to switch to [newly added](https://github.com/artsy/metaphysics/pull/1562)


src/lib/Scenes/Fair/__fixtures__/index.ts
```diff
@@ -49,12 +49,14 @@ export const fairFixture = {
     cached: null,
     day_schedules: null,
     day_schedule_text: null,
-    displayDaySchedules: [
```
ashkan18: as far as i can tell, `displayDaySchedules` is not even offered on `Fair` 🤷‍♂️

src/lib/Scenes/Fair/Screens/FairDetail.tsx
```diff
@@ -258,10 +258,6 @@ export const FairDetailContainer = createPaginationContainer(
         hours
         location {
           ...LocationMap_location
-          coordinates {
```
ashkan18: this is already part of `LocationMap_location`

src/lib/Scenes/Show/Screens/Detail.tsx
```diff
@@ -198,15 +198,6 @@ export const DetailContainer = createFragmentContainer(
       description
       city
       is_local_discovery
-      location {
```
ashkan18: removed since we already had https://github.com/artsy/emission/pull/1478/files#diff-bc10a37d56d39bee898f2d52abfcbcdfR209 few lines further down, correct me if im wrong but this seemed duplicate.

alloy: Not entirely. When using proper Relay containers you should _not_ worry about other fragments fetching the same data, as doing so would couple your components. Relay goes out of its way to enforce this through ‘data masking’ and will dedupe those selections in the final compiled query.

Required reading https://facebook.github.io/relay/docs/en/thinking-in-relay.html#data-masking

alloy: Most importantly, this file’s component seems to only care about the existence of a location, not the actual contents. For the existence check to work, only the stuff you left in is needed.

```graphql
      location {
        ...LocationMap_location
      }
```

alloy:
LGTM 👍

src/lib/Scenes/Show/Screens/Detail.tsx
```diff
@@ -69,11 +69,11 @@ export class Detail extends React.Component<Props, State> {
       },
     })
 
-    if (show.location && !isEmpty(show.location.displayDaySchedules)) {
+    if (show.location && !isEmpty(show.location)) {
```
ashleyjelks: Why isn't the truthy check for show.location sufficient here/do we need the isEmpty() call? Or does the LocationMap_location sometimes return an empty object?

alloy: Good question!

ashkan18: good point, i removed the `displayDaySchedule` part and i think this doesn't make much sense anymore, i'll do a small follow up PR and merge on green it. 

Thanks both for looking

PR #1479, opened 2019-03-13
Revert remove save shows

ashleyjelks:
This PR reverts a change where the Saved button was removed from all instances of the Show Item Row component.

Now the button is only removed from this component on the Show Card in the Global Map view.

Kapture 2019-03-13 at 13 36 56


kierangillen:
nice!

orta:
Ace 👍

PR #1480, opened 2019-03-13
Make sure mapbox attribution is visible

l2succes:


Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -206,6 +226,14 @@ - (void)showCityPicker:(NSString *)sponsoredContentUrl
     }];
 }
 
+- (void)displayLicensingViews
+{
+    RCTMGLMapView *mapView = FindMapView(self.mapVC.view);
+    
+    [mapView.attributionButton alignBottomEdgeWithView:self.mapVC.view predicate:@"-50"];
+    [mapView.logoView alignBottomEdgeWithView:self.mapVC.view predicate:@"-50"];
```
ashfurrow: We should put a guard in so we only make these constraints once.

PR #1482, opened 2019-03-13
Map analytics v2.0

kierangillen:
- Commented out code for the Session for now, should work to get it working for launch 1.1
- Adds all analytic from spreadsheet
- Fixes no image state for artist lists

## Artists lists with no image

Screen Shot 2019-03-13 at 3 18 13 PM


src/lib/Scenes/City/City.tsx
```diff
@@ -95,18 +96,27 @@ export class CityView extends Component<Props, State> {
...
+  // })
+  // trackSession() {
+  //   console.log("duration", sessionStart - Date.now())
+  //   return null
+  // }
+
+  // componentDidMount() {
+  //   sessionStart = Date.now()
+  // }
+
```
orta: Extra comments ^

orta: ah noticed you said this was fine, OK

src/lib/Scenes/City/CityPicker.tsx
```diff
@@ -65,6 +63,12 @@ export class CityPicker extends Component<Props, State> {
     )
   }
 
+  @track(() => {
+    return {
+      action_name: Schema.ActionNames.BMWLogo,
+      action_type: Schema.ActionTypes.Tap,
+    } as any
```
orta: this feels a bit weird, we should revisit ( /cc @alloy )

src/lib/Scenes/City/Components/Event/index.tsx
```diff
@@ -94,8 +102,31 @@ export class Event extends React.Component<Props, State> {
...
+    const actionName = args[0]
+    const slug = args[1]
+    const id = args[2]
+    return {
+      action_name: actionName,
+      action_type: Schema.ActionTypes.Tap,
+      owner_type: Schema.OwnerEntityTypes.Show,
+      owner_id: id,
+      owner_slug: slug,
+    } as any
```
orta: this too

orta:
OK, one or two contentious `as any`s, but not blockers

src/lib/Scenes/City/Components/Event/index.tsx
kierangillen: Yeah i looked at this with @l2succes and it was acting weird with anything but `any`, interested in revisiting it - we're using it all over the analytics

alloy: Yeah this isn’t good. Can you file a ticket that this should be cleaned up post launch and include the TS error it throws?

PR #1484, opened 2019-03-13
Hide Artists title when there are zero artists

orta:
Fixes LD-422 by checking the artists count before showing the Artists section

#trivial


orta:
I needed this PR to build off, so I carried on and also did LD-403

Screen Shot 2019-03-13 at 6 32 26 PM

src/lib/Components/Show/ShowArtistsPreview.tsx
```diff
@@ -71,12 +71,22 @@ export const ShowArtistsPreviewContainer = createFragmentContainer(
...
+
+      # Comes from CMS
       artists {
         _id
         id
         href
         ...ArtistListItem_artist
       }
+
+      # Comes from stubbed data
```
ashfurrow: Nice commenting 👍

PR #1487, opened 2019-03-14
Fixes drawer UX

ashfurrow:
This fixes the drawer interaction, hopefully once and for all:

The was broken by our switch to `ScrollableTabView` and #1476. The solution is complicated. But basically, the new tab view implementation has _a bunch_ of vertical scroll views (the old one replace the contents each time). This means that our `ARLocalDiscoveryCityGotScrollView` notification can't look for a reliable vertical scroll view (since the one it finds might be offscreen later). It also allows vertical bouncing, which is the behaviour we're seeing on the current beta. 

The solution is to find the city content scroll view, then walk up the hierarchy to find its containing `UIScrollView`. This is the horizontally-scrolling (vertically-bouncing) `ScrollableTabView`. We could just disable scrolling, but remember that we need to disable all the vertical scroll views inside it, too, which is a mess. So I just *disable all user interaction* under the tab bar while the drawer isn't open. Tab buttons will still work. This is a UX change I want to flag for you @l2succes. But I think this is the best way. 

Fixes [LD-437](https://artsyproduct.atlassian.net/browse/LD-437). #native_no_changes.

### DIFF
--- a/Pod/Classes/ViewControllers/ARMapContainerViewController.m
+++ b/Pod/Classes/ViewControllers/ARMapContainerViewController.m
@@ -24,7 +24,8 @@ NSString * const __nonnull SelectedCityNameKey = @"SelectedCityName";
 @property (nonatomic, strong) ARCityPickerComponentViewController *cityPickerController;
 @property (nonatomic, strong) UIView *cityPickerContainerView; // Need a view to have its own shadow.
 
-@property (nonatomic, weak) UIScrollView *rnScrollView;
+/// ScrollableTabView contains all the other scroll views.
+@property (nonatomic, weak) UIScrollView *scrollableTabView;
 
 @property (nonatomic, assign) BOOL initialDataIsLoaded;
 @property (nonatomic, assign) BOOL attributionViewsConstraintsAdded;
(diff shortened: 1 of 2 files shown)


Pod/Classes/ViewControllers/ARMapContainerViewController.m
```diff
@@ -309,9 +321,10 @@ - (void)drawerPositionDidChangeWithDrawer:(PulleyViewController *)drawer bottomS
 
     BOOL isDrawerOpen = [drawer.drawerPosition isEqualToPosition:PulleyPosition.open];
     [self.cityVC setProperty:@(isDrawerOpen) forKey:@"isDrawerOpen"];
-    self.rnScrollView.scrollEnabled = isDrawerOpen;
+    self.scrollableTabView.userInteractionEnabled = isDrawerOpen;
     if (!isDrawerOpen) {
-        [self.rnScrollView setContentOffset:CGPointZero animated:YES];
+        UIScrollView *cityScrollView = FindCityScrollView(self.cityVC.view);
```
ashfurrow: Note that this change is _closer_ to what we'd like (all the vertical scroll views scroll to the top) but currently grabs whichever one we find first (the AllEvents one). Since it has the title, I figured this makes sense for now but I'll open a ticket to follow-up.

ashfurrow: https://artsyproduct.atlassian.net/browse/LD-451

PR #1488, opened 2019-03-14
Fix fair hours

ashkan18:
Fix using `openingHours` on `Fair`, while `fair` has `Location` which can have `openingHours`, looks like we are using `hours` field of a fair which is a markdown text.

This was caused by #1478


ashkan18:
merging so it gets to next build

PR #1489, opened 2019-03-14
 Reset zoom level after switching city or filter

l2succes:
This resets the zoom level when a user switches cities or filters. We also adjust the minimum zoom level which was too restricted.


src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -510,10 +530,10 @@ export class GlobalMap extends React.Component<Props, State> {
       styleURL: ArtsyMapStyleURL,
       userTrackingMode: Mapbox.UserTrackingModes.Follow,
       centerCoordinate: [centerLng, centerLat],
-      zoomLevel: 13,
-      minZoomLevel: 11,
-      logoEnabled: false,
-      attributionEnabled: true,
+      zoomLevel: DefauftZoomLevel,
```
orta: typo

orta:
Yep, alright, looking good 👍

PR #1492, opened 2019-03-14
Adds RUNNING_AND_UPCOMING to CitySavedList

kierangillen:
- Needs updated relay schema once MP is promoted


ashkan18:
this should be good to go @ashfurrow  no more WIP

PR #1499, opened 2019-03-14
Fixes zoom outs when switching cities and fix cluster crash

l2succes:
Fixes the following crash: https://sentry.io/organizations/artsynet/issues/933653125/
Related: https://artsyproduct.atlassian.net/browse/LD-407

We were getting inconsistent zoom outs as the user was switching cities because of a race condition.

Two different zoom levels were being applied
1. when a user switches filters
2. when a user switches cities via the picker

Here we remove the filter switch zoom out as it was causing issues as they were being applied almost at the same time when switching cities. Also speeds up the zoom out animation a bit.

A better long term fix will be to set bounds to contain all pins on the map to make sure there's always something visible on the map.

#trivial

### DIFF
--- a/src/lib/Scenes/Map/Components/ShowCard.tsx
+++ b/src/lib/Scenes/Map/Components/ShowCard.tsx
@@ -58,7 +58,11 @@ export class ShowCard extends Component<ShowCardProps, ShowCardState> {
     const equal = isEqual(previousIds, currentIds)
 
     if (!this.state.isSaving && !equal && this.list) {
-      setTimeout(() => this.list.scrollToOffset({ offset: 0, animated: true }), 500)
+      setTimeout(() => {
+        if (this.list) {
+          this.list.scrollToOffset({ offset: 0, animated: true })
+        }
+      }, 500)
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/lib/Scenes/Map/Components/ShowCard.tsx
```diff
@@ -58,7 +58,11 @@ export class ShowCard extends Component<ShowCardProps, ShowCardState> {
     const equal = isEqual(previousIds, currentIds)
 
     if (!this.state.isSaving && !equal && this.list) {
-      setTimeout(() => this.list.scrollToOffset({ offset: 0, animated: true }), 500)
+      setTimeout(() => {
+        if (this.list) {
```
ashfurrow: Classic!

orta: hahah - yep

orta:
Yep, cool, merge on green

src/lib/Scenes/Map/Components/ShowCard.tsx
l2succes: lol yup, that's the JS version of `[weak self]` bug

ashfurrow:
I'll merge locally.

PR #1500, opened 2019-03-14
Fixes missing "view all shows" link in sponsored content sections

ashfurrow:
We reproduced this in the simulator by setting the `dev` query string param to `false`: 

https://github.com/artsy/emission/blob/d053d5335fe9b335759b6bb9d9f11e605b1b8e74/Example/Emission/AppSetup.m#L76

The `CaretButton` was wrapped  in a `Sans` component, but it actually contains a `Sans` component, so that double-`Sans` was causing no text to display.

Fixes [LD-418](https://artsyproduct.atlassian.net/browse/LD-418).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Fixes C compiler error - ash
 - Removes gallery context from fair page - ash
 - Only shows shows with images in all bricks - kieran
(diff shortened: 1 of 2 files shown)


ashleyjelks:
Merge on green!

ashleyjelks:
Tricky bug, nice work @ashfurrow!

PR #1502, opened 2019-03-14
[relay] Make sure Map/City/Show/Favorites views all use the same environment and store.

alloy:
Fixes https://artsyproduct.atlassian.net/browse/LD-415


alloy:
I want to test this PR in the latest Eigen beta first.

PR #1505, opened 2019-03-14
Noscroll padding

ashleyjelks:
Adds margins to View when there are no saved shows, fairs, or galleries to render on the City card.

#trivial


l2succes:
merging manually

alloy:
Yup, looks good, thanks!

Merge on green

PR #1508, opened 2019-03-15
Optimize tab switching in city view #trivial

l2succes:
Just a small change to improve tab switching performance a bit. A better solution to come


l2succes:
I'll go ahead and merge this one

PR #1510, opened 2019-03-15
[DO NOT MERGE] Shared performance issue hunt branch

alloy:
This branch is meant to find the cause of the tab switching performance issue.

### DIFF
--- a/src/lib/Scenes/Favorites/index.tsx
+++ b/src/lib/Scenes/Favorites/index.tsx
@@ -28,6 +28,10 @@ import ShowsRenderer from "./Components/Shows/Relay/FavoriteShows"
 
 import { gravityURL } from "lib/relay/config"
 
+import TabBar, { Tab } from "lib/Components/TabBar"
+
+const TabBarContainer = styled.View``
+
 const Title = styled.Text`
   font-family: ${Fonts.GaramondRegular};
   font-size: 30px;
@@ -60,10 +64,9 @@ class Favorites extends React.Component<Props, null> {
         <ScrollableTabView
           onChangeTab={selectedTab => this.fireTabSelectionAnalytics(selectedTab)}
           renderTabBar={props => (
-            <View>
-              <Title>Saves &amp; Follows</Title>
-              <ScrollableTabBar {...props} />
-            </View>
+            <TabBarContainer>
+              <TabBar {...props} />
+            </TabBarContainer>
           )}
         >
           <ScrollableTab tabLabel="Works">
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,4 @@
 {
-  "plugins": [["relay", { "artifactDirectory": "./src/__generated__" }]],
+  "plugins": [["relay", { "artifactDirectory": "./src/__generated__" }], "transform-remove-console"],
   "presets": ["react-native"]
 }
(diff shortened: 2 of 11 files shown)


alloy:
Ok, thus far I’ve disabled the following things and that still has _not_ gotten rid of the lag:
* bucketing more than once (initially)
* got rid of all console.log calls
* any map rendering
* any eventlist rendering
* any scroll regular events being sent from the native side to js

alloy:
When I log the traffic over the RN bridge, it _feels_ like the react reconciler updates from js to native takes longer than any other message, so I'll do some native code profiling to see what happens there on the native side

See how that first `WebSocketModule` `send` event in each transition seems to take the longest? That's React JS posting the required view changes to native: https://artsy.slack.com/files/U03HAB39U/FGXJSNWSC/log-rn-bridge-traffic.mov

alloy:
Next things to try:

* the tab titles are still there and skimming the message over the bridge does show some text/SC stuff showing up
* then see if you can disable animations on the tabs entirely

alloy:
Nothing really on the native side seems to pop out, most time on tab switching is spent in a JS call

Screenshot 2019-03-15 at 12 44 02

src/lib/Scenes/Favorites/index.tsx
```diff
@@ -60,10 +64,9 @@ class Favorites extends React.Component<Props, null> {
         <ScrollableTabView
           onChangeTab={selectedTab => this.fireTabSelectionAnalytics(selectedTab)}
           renderTabBar={props => (
-            <View>
-              <Title>Saves &amp; Follows</Title>
-              <ScrollableTabBar {...props} />
-            </View>
+            <TabBarContainer>
+              <TabBar {...props} />
+            </TabBarContainer>
```
alloy: Ok, so this ended up being the major issue with the tab bar. Why using a scrollable tab bar is not performant isn’t clear yet, but we can work around that for now, as it doesn’t seem like we really need it to be scrollable. (This same scrollable tab bar is also used in Favourites and has the same type of bad performance, so we should switch it out there too.)

Aside from this issue, the issue is likely compounded by a few different CPU bound operations, which we'll try to reduce too.

ashleyjelks: We added the Scrollable tab on the Favorites menu b/c there used to be a Saved Fairs options and we needed the scrolling to support the extra content. It can be removed for now of course, but I think the original reason for leaving it there was that we eventually intended to support saving fairs. (I think @l2succes  or @mbilokonsky can speak more to why we couldn't support saving fairs for the time being. But also not critical so this can wait) In the meantime I'll revert original tab in Favorites!

alloy: @ashleyjelks Can you copy that message to https://artsyproduct.atlassian.net/browse/LD-476 ?

mbilokonsky: I weighed in on the Jira ticket but will respond here too - as I understand it we scrapped the saved fairs feature because we found some subtle ambiguities in terms of what various fair references refer to. A little fuzzy, this was before my break, but if I remember correctly we have a lot of code that seems to treat `fair` and `fairOrganizer` interchangeably when they are very much not.

I think @anandaroop can speak more to this if you're curious, and I think Jun made the call to drop the feature.

alloy: Thanks for the context 👌

l2succes:
These issues are fixed with the following PR: https://github.com/artsy/emission/pull/1511. To recap steps taken:

We noticed very slow map and city view performance when a user switches tabs.
This PR does the following.

- Create separate supercluster instance for each filter type
- Ensures map clustering data gets computed only initial or when show data is mutated
- Cap tab list data to a fixed number (set to 25) which was the heaviest slowdown

PR #1511, opened 2019-03-15
Passes featurecollection as a group

kierangillen:
- Map clustering data now doesn't recreate geo data on tab change
- FIXME: Needs to pass clustering engine to each filter object


orta:
Might wanna rebase these commits a bit :D

src/lib/utils/convertCityToGeoJSON.ts
```diff
@@ -61,5 +61,5 @@ export const convertCityToGeoJSON = data => {
           },
         }
       }),
-  } as MapGeoFeatureCollection
+  } as any
```
alloy: Why the need to cast to `any`?

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -213,26 +212,42 @@ export class GlobalMap extends React.Component<Props, State> {
...
   }
 
   componentWillUnmount() {
     EventEmitter.unsubscribe("filters:change", this.handleFilterChange)
   }
 
+  componentDidUpdate(_, prevState) {
+    // Update the clusterMap if new bucket results
+    if (this.state.bucketResults) {
+      const shouldUpdate = ["saved"].some(key => {
```
alloy: Did you copy this pattern because you expect to add more keys?

src/lib/Scenes/Map/Components/PinsShapeLayer.tsx
```diff
@@ -52,7 +58,7 @@ export class ShapeLayer extends Component<Props, any> {
   }
 
   componentWillReceiveProps(newProps: Props) {
-    if (!isEqual(this.props.featureCollection.features, newProps.featureCollection.features)) {
+    if (!isEqual(this.props.featureCollections[0].shapes.features, newProps.featureCollections[0].shapes.features)) {
```
l2succes: I'm not sure about  `featureCollections` being an array, I would say an object where the key is the filter type makes more sense. Lookup times are faster

src/lib/Scenes/City/City.tsx
```diff
@@ -179,7 +179,6 @@ export class CityView extends Component<Props, State> {
             <ScrollableTabView
               initialPage={this.props.initialTab || AllCityMetaTab}
               onChangeTab={this.setSelectedTab}
-              prerenderingSiblingsNumber={2}
```
ashfurrow: Won't this remove the optimization of loading the next tabs for the user?

l2succes: Yes it will but now that we drastically reduced rendering tab of list view (capped to 30 items), it's not needed anymore as they load fast enough for the user not to notice

ashfurrow: Pretty good reason to remove it!

l2succes:
Related: https://artsyproduct.atlassian.net/browse/LD-338

PR #1512, opened 2019-03-15
Deploy script syncs local data with Metaphysics

ashfurrow:
Makes sure we always have the latest. I also ran the script with the changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 ### Master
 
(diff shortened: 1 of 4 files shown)


ashfurrow:
Here's what Hong Kong looks like now:

Screen Shot 2019-03-15 at 11 02 55 AM

orta:
Yeah, this makes a bunch of sense

PR #1513, opened 2019-03-15
Updates drawer sub-scrollview enabled state when tabs change 📱

ashfurrow:
The problem was occurring when we the tabs changed and a new scroll view was added to the view hierarchy. I tested this on device to make sure it worked.


orta:
Ah, yeah! This makes sense!

Merge on green

PR #1515, opened 2019-03-15
adds 10px of space between tappable areas to prevent event collision #trivial

ashleyjelks:
[Links to LD-479](https://artsyproduct.atlassian.net/browse/LD-479)

Adds a bit of space between 2 tappable areas to prevent firing both events if a user taps in both areas at the same time. The red border is only here to show the amount of space that has been added between the components. It is not in the final markup.

Screen Shot 2019-03-15 at 12 50 28 PM

#trivial

### DIFF
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -135,11 +135,7 @@ type Artist implements Node & Searchable {
     before: String
     last: Int
   ): ArticleConnection
-  articles(
-    sort: ArticleSorts
-    limit: Int
-    in_editorial_feed: Boolean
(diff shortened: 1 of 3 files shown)


ashfurrow:
The rebase left this commit out of the git history, but I merged here: https://github.com/artsy/emission/commit/c342f37f0ccbb67c66257d2144f243d092fe6d4f 

The commit from this PR was rewritten as: https://github.com/artsy/emission/commit/b06fe7d40ea483da2534458c9b891f55d8170a57

PR #1516, opened 2019-03-15
Fewer city render passes

alloy:
Will comment inline about reduced city render pass. Otherwise this PR replaces inline function props to be stable, just on the odd chance that we're re-rendering trees more than necessary as these functions will never be the same across render passes.

### DIFF
--- a/src/lib/Scenes/City/City.tsx
+++ b/src/lib/Scenes/City/City.tsx
@@ -29,7 +29,6 @@ interface State {
   relay: RelayProp
   cityName: string
   citySlug: string
-  selectedTab: number
   sponsoredContent: { introText: string; artGuideUrl: string }
   relayErrorState?: RelayErrorState
 }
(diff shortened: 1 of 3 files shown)


src/lib/Scenes/City/City.tsx
```diff
@@ -104,7 +102,6 @@ export class CityView extends Component<Props, State> {
   }
 
   setSelectedTab(selectedTab) {
-    this.setState({ selectedTab: selectedTab.i })
```
alloy: This state doesn't actually get used by the component, but does trigger an extra render pass.

https://github.com/artsy/emission/pull/1406/files#r266063792

src/lib/Scenes/City/City.tsx
```diff
@@ -147,6 +144,23 @@ export class CityView extends Component<Props, State> {
...
+    return (
+      <View>
+        <TabBar {...props} spaceEvenly={false} />
+      </View>
+    )
+  }
+
+  // TODO: Is it correct that we have these two similar ones?
+  onScrollableTabViewLayout = layout => {
+    this.scrollViewVerticalStart = layout.nativeEvent.layout.y
```
l2succes: Yup, not sure if this is needed so we should revisit this post v1

l2succes:
merge on green

PR #1519, opened 2019-03-16
Adds a hitslop to the button component, and uses it in show list items

orta:
Re: LD-471

Also adds some comments to the props of the button, and removes unused props


l2succes:
merge on green

PR #1429, opened 2019-03-10
Adds more metadata to the Podspec

orta:
This JSON can be picked up in Eigen, and used to provide better visibility into the Emission integration in Eigen.

```
~/d/p/a/i/emission  $ pod ipc spec Emission.podspec
{
  "name": "Emission",
  "version": "1.8.14",
  "summary": "React Native Components used by Eigen.",
  "homepage": "https://github.com/artsy/emission",
  "license": "MIT",
  "authors": {
    "Artsy Mobile": "mobile@artsy.net"
  },
  "source": {
    "git": "https://github.com/artsy/emission.git",
    "tag": "v1.8.14"
  },
  "platforms": {
    "ios": "8.0"
  },
  "source_files": "Pod/Classes/**/*.{h,m}",
  "preserve_paths": "Pod/Classes/**/*.generated.objc",
  "resources": "Pod/Assets/{Emission.js,assets}",
  "dependencies": {
    "Artsy+UIColors": [

    ],
    "Artsy+UIFonts": [
      ">= 3.0.0"
    ],
    "Extraction": [
      ">= 1.2.1"
    ],
    "Pulley": [

    ],
    "SDWebImage": [
      ">= 3.7.2",
      "< 4"
    ],
    "React/Core": [
      "0.54.4"
    ],
    "React/CxxBridge": [
      "0.54.4"
    ],
    "React/RCTAnimation": [
      "0.54.4"
    ],
    "React/RCTCameraRoll": [
      "0.54.4"
    ],
    "React/RCTImage": [
      "0.54.4"
    ],
    "React/RCTLinkingIOS": [
      "0.54.4"
    ],
    "React/RCTNetwork": [
      "0.54.4"
    ],
    "React/RCTText": [
      "0.54.4"
    ],
    "React/RCTGeolocation": [
      "0.54.4"
    ],
    "React/RCTActionSheet": [
      "0.54.4"
    ],
    "yoga": [
      "0.54.4.React"
    ],
    "DoubleConversion": [
      "1.1.5"
    ],
    "Folly": [
      "2016.09.26.00"
    ],
    "glog": [
      "0.3.4"
    ],
    "tipsi-stripe": [
      "5.2.4"
    ],
    "react-native-mapbox-gl": [
      "6.1.3"
    ],
    "SentryReactNative": [
      "0.30.3"
    ],
    "RNSVG": [
      "9.0.4"
    ]
  },
  "native_version": 23,
  "release_date": "Mar 10, 2019",
[...]

PR #1521, opened 2019-03-18
Only includes Podspec metadata on releases #trivial

ashfurrow:
Follow-up from #1429. Fixes LD-485.

This PR only includes the extra metadata when we generate the [JSON version](https://github.com/artsy/Specs/blob/master/Emission/1.8.26/Emission.podspec.json) of the Podspec. This will keep the podspec used by `pod install` during development deterministic. 

Not sure why `Example/Pods/Pods.xcodeproj/project.pbxproj` has so many changes, but they seem benign so I'm going to attribute them to Xcode's arcane inner-workings.


orta:
Yep, looks 👍

PR #1524, opened 2019-03-18
Fix UI issues in Saves and Follows on non iPhone X devices

ashleyjelks:
This PR corrects text overflow issues that happened in the Favorites view on smaller sized iPhone devices. It also changes the order of the tabs that appear in the favorites view.
Kapture 2019-03-18 at 14 03 34
Kapture 2019-03-18 at 14 05 08

[Links to LD-481](https://artsyproduct.atlassian.net/browse/LD-481)

### DIFF
--- a/src/lib/Components/ScrollableTabBar.tsx
+++ b/src/lib/Components/ScrollableTabBar.tsx
@@ -78,6 +78,10 @@ export default class ScrollableTabBar extends React.Component<ScrollableTabBarPr
         accessible={true}
         accessibilityLabel={name}
         accessibilityTraits="button"
+        onLayout={e => {
+          const layout = e.nativeEvent.layout
+          this.els[page] = layout
+        }}
         onPress={() => {
           this.setState({
             activeTab: page,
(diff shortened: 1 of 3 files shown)


src/lib/Components/ScrollableTabBar.tsx
```diff
@@ -26,19 +26,18 @@ const Tabs = styled.ScrollView`
...
-  padding-left: 20px;
-  padding-right: 20px;
   flex-grow: 1;
   ${p =>
     p.active &&
     `
     border-color: ${color("black100")};
     border-bottom-width: 2px;
   `};
+  ${p => (p.isNonXPhone ? `padding-left:15px; padding-right:15px;` : `padding-left:25px; padding-right:20px;`)};
```
ashfurrow: This kind of 'is non-X iPhone' boolean tightly couples our `TabButton` component to the device it's running on (or, more specifically, whether or not its container is at least a certain width).

The bug really is that when you tap on a button that's partially offscreen, it doesn't get scrolled on screen. This PR solves that bug by making sure everything is on screen for this specific view, but a more long-term solution would be to modify the `ScrollableTabBar` component to scroll itself when a button gets tapped. Would you like to pair?

alloy:
I thought we had removed the scrollable tab bar for now? (Due to performance issues)

ashfurrow:
@alloy this is on the global saves & follows tab (favourites tab in Eigen).

PR #1526, opened 2019-03-18
Fixes LD saving events inconsistencies

ashfurrow:
This is a follow-up from #1502. 

The query renderers in `QueryRenderers.tsx` all use the same environment, but it is not the `defaultEnvironment`:

https://github.com/artsy/emission/blob/150ed9f7eefbf3494fcd81e11c5fccce7f5d5cf6/src/lib/relay/QueryRenderers.tsx#L57

But that environment is different from the environment used in our map component: 

https://github.com/artsy/emission/blob/150ed9f7eefbf3494fcd81e11c5fccce7f5d5cf6/src/lib/Scenes/Map/MapRenderer.tsx#L25

Leading to [LD-488](https://artsyproduct.atlassian.net/browse/LD-488). 

There seem to be only two production Relay environments left, and this PR will unite them into one. I've added a comment to the `createEnvironment.ts` file for now, and have added tooling discussion to prevent this in the future to our list of Front-End iOS Practice items.

### DIFF
--- a/src/lib/relay/createEnvironment.ts
+++ b/src/lib/relay/createEnvironment.ts
@@ -8,6 +8,8 @@ import { metaphysicsExtensionsLoggerMiddleware } from "./middlewares/metaphysics
 
 const Emission = NativeModules.Emission || {}
 
+/// WARNING: Creates a whole new, separate Relay environment. Useful for testing and in Storybooks.
+/// Use `defaultEnvironment` for production code.
 export default function createEnvironment() {
   const network = new RelayNetworkLayer([
     cacheMiddleware(),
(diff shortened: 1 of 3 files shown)


src/lib/relay/createEnvironment.ts
```diff
@@ -8,6 +8,8 @@ import { metaphysicsExtensionsLoggerMiddleware } from "./middlewares/metaphysics
 
 const Emission = NativeModules.Emission || {}
 
+/// WARNING: Creates a whole new, separate Relay environment. Useful for testing and in Storybooks.
+/// Use `defaultEnvironment` for production code.
```
alloy: 👍

ashfurrow:
I'm still going through the other views (there are quite a few in that file!). I'll merge when finished and green.

PR #1528, opened 2019-03-19
Uses new isActive field

kierangillen:
- Uses `isActive` instead of `is_active` on fair 

#trivial


src/lib/Scenes/Fair/Screens/FairDetail.tsx
```diff
@@ -91,7 +91,7 @@ export class FairDetail extends React.Component<Props, State> {
 
     let boothCount = 0
 
-    if (is_active) {
+    if (isActive) {
```
anandaroop: 👍🏽

PR #1529, opened 2019-03-19
Relay no longer caches GraphQL responses which contain errors field

ashfurrow:
This PR changes the caching middleware on Emission to no longer store GraphQL responses that contain `errors` in the on-device cache. I have tested this locally against metaphysics.

Fixes [LD-502](https://artsyproduct.atlassian.net/browse/LD-502).

### DIFF
--- a/src/lib/relay/middlewares/cacheMiddleware.ts
+++ b/src/lib/relay/middlewares/cacheMiddleware.ts
@@ -34,19 +34,28 @@ export const cacheMiddleware = () => {
 
     const response: RelayResponsePayload = await next(req)
 
-    if (response.status >= 200 && response.status < 300) {
-      if (isQuery) {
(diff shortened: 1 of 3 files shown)


src/lib/relay/middlewares/cacheMiddleware.ts
```diff
@@ -34,19 +34,28 @@ export const cacheMiddleware = () => {
...
+    }
+
     if (response.status >= 200 && response.status < 300) {
       if (isQuery) {
-        cache.set(queryID, req.variables, JSON.stringify(response.json), req.cacheConfig.emissionCacheTTL)
+        // Don't cache responses with errors in them (GraphQL responses are always 200, even if they contain errors).
+        if (response.json.errors === undefined) {
+          cache.set(queryID, req.variables, JSON.stringify(response.json), req.cacheConfig.emissionCacheTTL)
+        } else {
+          clearCacheAndThrowError()
```
anandaroop: Looks reasonable to me 👍

PR #1530, opened 2019-03-19
LD-483 - Adds museums to all shows

kierangillen:
- Museums now appear in all shows on map
- Fixes cluster size for clusters of 5 and less shows to be the same size
- https://artsyproduct.atlassian.net/browse/LD-483

#trivial


l2succes:
merge on green

PR #1532, opened 2019-03-19
Replaces start/end date range with "Ongoing" to Shows running for 2 or more years

ashleyjelks:
This PR updates all the places in the UI where we display Show date information:

-if the show is running for 2+ years we now show "Ongoing" instead of the dates.

The screenshots below reflect what this change will look like in all the possible places this change could appear.

[Links to LD-467](https://artsyproduct.atlassian.net/browse/LD-467)

Screen Shot 2019-03-19 at 1 17 30 PM
Screen Shot 2019-03-19 at 1 20 11 PM
Screen Shot 2019-03-19 at 1 25 02 PM
Screen Shot 2019-03-19 at 1 29 06 PM
Screen Shot 2019-03-19 at 1 29 59 PM
Screen Shot 2019-03-19 at 1 30 08 PM


ashleyjelks:
@ashfurrow would it be better to make these changes in MP now instead of doing this refactor?

ashfurrow:
@ashleyjelks if you have the time! I know shipping v1.1 is important, too.

kierangillen:
@ashleyjelks This looks good! We may want to move the `getExhibitionPeriod` function into an export in its own file and then just import it where needed instead of having it in each file.

kierangillen:
Oh I missed @ashfurrow comment my bad for repeating his thought!

l2succes:
merge on green

kierangillen:
@ashfurrow and I looked into adding this to metaphysics but decided against it. The MP exhibitionPeriod function is being used in too many places to make the change not be overly impactful.

PR #1535, opened 2019-03-19
Fixes Misplaced Scrollbar on Saved/Follows shows

ashleyjelks:
This PR fixes a regression from a refactor that componentized Saved Shows.

[Links to LD-473](https://artsyproduct.atlassian.net/browse/LD-473)

Kapture 2019-03-19 at 15 12 43

#trivial


kierangillen:
I think @alloy made the point that we don't want to be adding margin to these components, since a component shouldn't be rendering anything outside of its viewable content. The parent should be rendering any margin.

l2succes:
merge on green

PR #1536, opened 2019-03-19
LD-510 - Adds correct formatting to time in show event section

kierangillen:
- https://artsyproduct.atlassian.net/browse/LD-510
- Adds correct formatting to time in show event section
- Adds `openingReceptionText` for stubshows
- Renames show `EventSection` to `ShowEventSection` since there's another city component with the same name.

Screen Shot 2019-03-19 at 3 29 14 PM


src/lib/Scenes/Show/Screens/MoreInfo.tsx
```diff
@@ -47,6 +47,22 @@ export class MoreInfo extends React.Component<Props, State> {
       })
     })
 
+    if (show.openingReceptionText) {
+      if (show.press_release) {
```
anandaroop: Is the nesting of these `if`s intentional? I'm not seeing how these are dependent on one another.

kierangillen: Good call this was an error.

PR #1537, opened 2019-03-19
Configure a maxZoom for map component

anandaroop:
Implements: https://artsyproduct.atlassian.net/browse/LD-489

Workaround/fix for: https://artsyproduct.atlassian.net/browse/LD-438

(Update: Note that Jun and I have also adjusted min zoom and default zoom in this PR as well.)

This adds a maxZoom to the map. Before this you could zoom in so far that you were looking at nothing but a pin and a vast gray featureless expanse (see LD-489). Now you will be limited at a max zoom of 17.5, which basically ensures that you will see some street grid.

(This also appears to have a side effect of fixing LD-438.)

maxzoom

#trivial

### DIFF
--- a/src/lib/Scenes/Map/GlobalMap.tsx
+++ b/src/lib/Scenes/Map/GlobalMap.tsx
@@ -115,7 +115,9 @@ interface State {
 
 export const ArtsyMapStyleURL = "mapbox://styles/artsyit/cjrb59mjb2tsq2tqxl17pfoak"
 
-const DefaultZoomLevel = 12
+const DefaultZoomLevel = 11
+const MinZoomLevel = 9
+const MaxZoomLevel = 17.5
 
 const ButtonAnimation = {
   yDelta: -200,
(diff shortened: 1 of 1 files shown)


src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -116,6 +116,8 @@ interface State {
 export const ArtsyMapStyleURL = "mapbox://styles/artsyit/cjrb59mjb2tsq2tqxl17pfoak"
 
 const DefaultZoomLevel = 12
+const MinZoomLevel = 10
+const MaxZoomLevel = 17.5
```
anandaroop: With a lot of trial and error in the Simulator this is what I found:

- 17.0 = over-clustering: many clustered-yet-distinct addresses do not expand into distinct pins
- 17.5 = best compromise: most but not all clustered-yet-distinct addresses expand to distinct pins; no disappearing shows
- 18.0 = buggy: mismatch between cluster count and pins results in "disappearing" shows as you approach max zoom

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -210,6 +212,8 @@ export class GlobalMap extends React.Component<Props, State> {
 
     this.clusterEngine = new Supercluster({
       radius: 50,
+      minZoom: Math.floor(MinZoomLevel),
```
l2succes: 👍

PR #1541, opened 2019-03-20
Reset zoom, pitch and heading  upon city switch

anandaroop:
Iteration on https://artsyproduct.atlassian.net/browse/LD-407

When interacting with the map, it's possible to get into a permutation of custom **zoom**, **heading** (rotating away from north), and **pitch** (tilting away from camera-pointing down)

For example, facing down Canal St towards the Manhattan Bridge.

canal

Currently when we switch to another city, we reset the zoom, but leave the heading and pitch unchanged, which leads to an unexpected initial experience:

lon

We can instead reset all three — zoom, pitch and heading in order to restore a neutral initial vantage point into the new city.

Furthermore, by taking advantage of an easing animation with a generous duration, we can also mitigate the perceived delays associated with the initial fetch of a new city's data.

Putting that all together looks something like this: 

anim

Duration is currently set to 2 seconds, but that's up for grabs of course.

I would love to see this bit of finish added to v1.1 _**IF**_ it is reasonable to do so with solid Q/A. But if not, I don't want to rush things. Wondering what the best way to test this is outside of the beta. I looked into side-loading onto my own device but hit all kinds of walls with that. Would love some help if there's time tomorrow.

### DIFF
--- a/src/lib/Scenes/Map/GlobalMap.tsx
+++ b/src/lib/Scenes/Map/GlobalMap.tsx
@@ -109,6 +109,7 @@ export const ArtsyMapStyleURL = "mapbox://styles/artsyit/cjrb59mjb2tsq2tqxl17pfo
 const DefaultZoomLevel = 11
 const MinZoomLevel = 9
 const MaxZoomLevel = 17.5
+const DefaultCameraMode = 1 // https://github.com/nitaliano/react-native-mapbox-gl/blob/master/ios/RCTMGL/CameraMode.m
 
 const ButtonAnimation = {
   yDelta: -200,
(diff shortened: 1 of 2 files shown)


orta:
Can you interact with the screen during this animation? (If you already had most of the same values then that 2 seconds is gonna feel like it's unresponsive)

anandaroop:
Good point —  I was testing against the worst case, which benefits from a longer duration. Should test it with more typical cases and shorter durations too.

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -226,6 +227,16 @@ export class GlobalMap extends React.Component<Props, State> {
     this.setState({ activeIndex, activePin: null, activeShows: [] }, () => this.emitFilteredBucketResults())
   }
 
+  resetZoomAndCamera = () => {
+    this.map.setCamera({
```
l2succes: I've had some issues with the `setCamera` call in the past but I was only setting `zoom` and `duration`, we should test it out a bit when the pitch and heading are already at 0.

anandaroop:
Example when _only_ the zoom needs to change. Imho this 2s transition actually works well in conjunction with the spinner, but I'm open whatev duration people think is reasonable. 

only zoom

anandaroop:
@l2succes and I tested on-device with with various permutations of city/zoom/pitch/heading.

We've decided to go with 1000ms for the transition duration.

Also noting that the final zoom level that results from this seems to be not fully deterministic, which is true of the _current_ implementation as well.

PR #1551, opened 2019-03-21
React Native 0.57 - rebase and merge branch

orta:
Already converted #1210 into a single commit, then will be re-basing that to master

PR #1553, opened 2019-03-21
React Native Upgrade for 0.59

orta:
OK, now that #1551 is in.

Notable upgrade points:

- Jest 0.23 -> 0.24
  - Config changes

- React Native 0.57 - 0.59
  - Removal of NavigationiOS - See https://github.com/michaelknoch/react-native-navigator-ios/pull/1 
https://github.com/michaelknoch/react-native-navigator-ios/pull/2
  - https://github.com/react-native-community/react-native-async-storage is next up on the chopping block, but not this release

- Types
  - Seems to be some pretty fundamental type upgrades, because that's a lot of errors

### Blocked on

 - [x] `yarn type-check`
 - [x] `yarn jest`
 - [x] Running all screens
   - [x] Shows
   - [x] Fair
   - [x] Map
   - [x] Artist
   - [x] Homepage
   - [x] Gene
   - [x] Gene Refined
   - [x] My Profile
   - [x] Start Consignment Flow
   - [x] Inbox
   - [x] Inquiry
   - [x] Favorites
   - [x] Bid Flow

#skip_new_tests


orta:
Looks like since we started piping the TS output into `tee` - then fails don't matter, normally not a problem because Danger would run on it - but while this PR is a draft, Travis doesn't make pr-based CI runs, so no danger output it seems.

Example/Gemfile
```diff
@@ -9,8 +9,5 @@ gem 'cocoapods-check'
 # To manage our secret keys
 gem "cocoapods-keys"
 
-# To apply fixes to the React Native code
-gem 'cocoapods-fix-react-native'
```
ashfurrow: 😻

orta:
@alloy  - I believe you have commits laying around for this

alloy:
Yup, continuing on this today 👍

orta:
I made sure that TypeScript can fail CI now 👍

alloy:
Thanks!

alloy:
This PR updates tipsi-stripe from 5.2.4 to 7.4.0 because of RN API changes that were [fixed in 7.2.0](https://github.com/tipsi/tipsi-stripe/pull/431). Here’s the CHANGELOG https://github.com/tipsi/tipsi-stripe/blob/master/CHANGELOG.md#530---2018-07-23

We can probably fix the issue with a patch-package patch, if this update is deemed too dangerous.

Who is the person that knows most about how this _should_ all work?

alloy:
There’s a few deprecation warnings about libraries that have moved out of core RN, but I didn’t want to pile that on in this PR, we should follow up with that.

orta:
I just realized, we should be bumping major versions with these react native updates

alloy:
Should we? Are we changing Emission’s API?

alloy:
Seeing the following error in some of the Consignment Flow forms.

Simulator Screen Shot - iPhone 8 - 2019-04-02 at 19 16 14

alloy:
@orta Ok, I believe you and @sweir27 are testing the tipsi-stripe upgrade, but otherwise I think this is good to go.

orta:
I mean, yeah, it's a full breaking change the emission API right? You can't use this with an existing client at all

patches/graphql+0.13.2.patch
```diff
@@ -10,7 +10,7 @@ patch-package
 -  }
 +  // if (name.length > 1 && name[0] === '_' && name[1] === '_') {
 +  //   return new _GraphQLError.GraphQLError('Name "' + name + '" must not begin with "__", which is reserved by ' + 'GraphQL introspection.', node);
-+  // }
++  // }x
```
orta: ^

alloy: Hah

alloy:
I’m not sure I follow. Eigen doesn’t need to change any API usage (of Emission), right?

orta:
OK, I've given the stripe code a runtime audit and everything looks to be working just fine. Merging!

orta:
Thanks for putting the time in @alloy !

PR #1556, opened 2019-03-22
LD-531 - Adds correct margin/padding to last carousel image

kierangillen:
- https://artsyproduct.atlassian.net/browse/LD-531

### DIFF
--- a/src/lib/Scenes/Show/Components/ShowHeader/Components/Carousel.tsx
+++ b/src/lib/Scenes/Show/Components/ShowHeader/Components/Carousel.tsx
@@ -40,7 +40,12 @@ export class Carousel extends React.Component<Props> {
     const { sources } = this.props
 
     return sources.map((source, i) => (
-      <Flex key={this.keyForSource(source)} mr={1} alignItems="flex-start">
+      <Flex
+        key={this.keyForSource(source)}
+        mr={i === sources.length - 1 ? 2 : 1}
+        pr={i === sources.length - 1 ? 2 : 0}
+        alignItems="flex-start"
+      >
         <ImageView {...source} isFirst={i === 0} />
       </Flex>
     ))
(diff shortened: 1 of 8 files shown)


src/lib/Scenes/Show/Components/ShowHeader/Components/Carousel.tsx
```diff
@@ -40,7 +40,12 @@ export class Carousel extends React.Component<Props> {
     const { sources } = this.props
 
     return sources.map((source, i) => (
-      <Flex key={this.keyForSource(source)} mr={1} alignItems="flex-start">
+      <Flex
+        key={this.keyForSource(source)}
+        mr={i === sources.length - 1 ? 2 : 1}
+        pr={i === sources.length - 1 ? 2 : 0}
```
ashfurrow: We repeat this logic twice: `i === sources.length - 1`. Maybe we could put it in a local var? It'd make it easier to read.

storybook/storyLoader.js
```diff
@@ -1,99 +1,102 @@
...
+  '../src/lib/Scenes/Settings/__stories__/Settings.story',
+  '../src/lib/Scenes/Show/__stories__/Show.story',
+  
+];
 
 module.exports = {
   loadStories,
   stories,
-}
+};
```
ashfurrow: I'm seeing these changes locally to, let's commit  them here 👍

src/lib/Scenes/Show/Components/ShowHeader/Components/Carousel.tsx
kierangillen: Makes sense, just updated

ashfurrow:
Our TypeScript error-parsing is broken (see @orta's comment on #1553), but here's the problem:

> src/lib/Scenes/Show/Components/ShowHeader/Components/Carousel.tsx(55,3): error TS1005: ',' expected.

PR #1559, opened 2019-03-26
Fixes all strings instances to boolean for RN

kierangillen:
#trivial #skip_new_tests


ashfurrow:
These files are strewn about the codebase, so I've added `#skip_new_tests` to the PR description.

PR #1561, opened 2019-03-27
Fixes City not re-rendering after a MP request is successfully retried

ashfurrow:
Fixes [LD-533](https://artsyproduct.atlassian.net/browse/LD-533).

The problem was that we were only ever sending the error (and never sending once the error was completed). I added a check where we only send if the values are different, and I want to take a look at using a similar approach with calculating+emitting bucket results (marked by a TODO right now).

This also needs a test, hoping to pair with @kierangillen on the `EventEmitter` mocking stuff tomorrow morning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 ### Master
 
+- Fixes City not re-rendering after a MP request is successfully retried - ash
+
 ### 1.9.2
 
 - Fixes scrollable tab bar on global saves & follows - ash & ashley
--- a/src/lib/Scenes/Map/GlobalMap.tsx
+++ b/src/lib/Scenes/Map/GlobalMap.tsx
@@ -246,7 +246,9 @@ export class GlobalMap extends React.Component<Props, State> {
       setTimeout(this.resetZoomAndCamera, 500)
     }
 
+    // If there is a new city, emity it and update our map.
     if (nextProps.viewer) {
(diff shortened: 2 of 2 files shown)


ashfurrow:
Oh, I forgot the GIF:



I did this by pointing the app to `localhost` for metaphysics, removing `cache.set` from `cacheMiddleware.ts`, and applying the following diff to metaphysics to fail every other request:

```diff
diff --git a/src/lib/fetchPersistedQuery.ts b/src/lib/fetchPersistedQuery.ts
index 4460a3d8..4a698757 100644
--- a/src/lib/fetchPersistedQuery.ts
+++ b/src/lib/fetchPersistedQuery.ts
@@ -2,8 +2,26 @@ import queryMap from "../data/complete.queryMap.json"
 import { info, error } from "./loggers"
 import { RequestHandler } from "express"
 
+let succeed = true
+
 export const fetchPersistedQuery: RequestHandler = (req, res, next) => {
   const { documentID } = req.body
+
+  succeed = !succeed
+
+  if (!succeed) {
+    return setTimeout(
+      () =>
+        res
+          .send({
+            data: null,
+            errors: [{ error: "this is an error for testing" }],
+          })
+          .end(),
+      500
+    )
+  }
+
   if (documentID) {
     const query = queryMap[documentID]
     if (query) {
```

ashfurrow:
This is blocked by #1562 and I want to get it into the next beta, so I'm going to merge.

PR #1562, opened 2019-03-27
WIP: Fair tests

kierangillen:
- Fixes error handling in `MockRelayRenderer` and properly passes error message
- Mirrors `MockRelayRenderer` used in Reaction

#trivial


ashfurrow:
Took a look and we're getting errors like this in our tests:

> Query.mockData defined in resolvers, but not in schema

Happy to continue working on this together – we can also ask @ds300 or @alloy for assistance.

kierangillen:
Just paired with @ds300 on this and we fixed the mockData issue.

src/lib/Scenes/Fair/__tests__/Fair-tests.tsx
```diff
@@ -25,7 +25,7 @@ describe("Fair", () => {
           }
         `}
         mockData={{
-          Fair: () => fairFixture,
+          fair: fairFixture,
```
kierangillen: Here is an example of a test that is now working properly with the updated `mockRelayRenderer`

src/lib/Scenes/Fair/Components/FairBoothPreview/__tests__/index-tests.tsx
```diff
@@ -8,7 +8,7 @@ import { FairBoothPreviewContainer } from "../index"
 jest.unmock("react-relay")
 
 describe("FairBoothPreview", () => {
-  it("renders properly", async () => {
+  xit("renders properly", async () => {
```
alloy: Why are these (and below) being disabled?

src/lib/tests/createMockNetworkLayer/__tests__/createMockNetworkLayer-tests.tsx
```diff
@@ -30,29 +30,38 @@ describe("createMockNetworkLayer", () => {
   describe("preserves the upstream behaviour", () => {
     it("returns the data if present", async () => {
       const data = await fetchQueryWithResolvers({
```
alloy: Rename these either now or in a next PR to be about “data” instead of “resolvers”.

src/lib/utils/renderWithLoadProgress.tsx
```diff
@@ -13,8 +13,10 @@ export default function<P>(
   let retrying = false
   return ({ error, props, retry }) => {
     if (error) {
+      if (typeof jest !== "undefined") {
```
alloy: Add the comment from Reaction too (which explains why we do this).

src/lib/Scenes/Fair/Components/FairBoothPreview/__tests__/index-tests.tsx
kierangillen: Some tests need to be refactored after this is updated. This test has the wrong `mockData` fixture.

src/lib/tests/createMockNetworkLayer/__tests__/createMockNetworkLayer-tests.tsx
kierangillen: Updating this, this should also be updated in Reaction, as this was copied from there.

kierangillen:
@alloy Updated with your comments.

alloy:
Merge on green

PR #1566, opened 2019-03-28
Spot missing tests

ashleyjelks:
This PR adds todo notes to indicate where new tests need to be written and adds placeholder tests where we are missing test coverage for new features added from Local Discovery.

[Links to LD-529](https://artsyproduct.atlassian.net/browse/LD-529)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -424,7 +424,8 @@
 - Adds MoreInfo screen to show detail view - javamonn
 - Clean up Jest log output, update relay-mock-network-layer - javamonn
 - Fair performance enhancements - ashley
-- Add start and end properties to Favorites Works and Shows page info queries -ashley
+- Add start and end properties to Favorites Works and Shows page info queries - ashley
+- Identifies areas missing test coverage and adds files to support missing tests - ashley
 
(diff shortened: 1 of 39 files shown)


ashfurrow:
The Danger failures are because the files are named `FavoriteArtists/index.tsx` but the tests are `FavoriteArtists-tests.tsx`. I've added a note to address this to the FE iOS Practice, but I'd like to get this PR green without adding `skip new tests`. Can we rename the test files `index-tests.tsx`? We follow that pattern elsewhere in Emission, but if we should avoid it, let me know 👍

ashleyjelks:
@ashfurrow makes sense. Since there are so many tests that are missing and I brought this up during standup to see if its a good use of time to test all these smaller components. I think @l2succes said we should really focus on testing the larger screens and the map bounds. But I'm happy to make tickets to track all these tests.

ashfurrow:
@ashleyjelks cool, thanks! Like I said, we can have a single ticket for "fair tests" that covers the `Fair` and `FairHeader` components, etc. Then we we write the fair tests, we can decide what level of testing is appropriate for each TODO comment with that ticket number.

PR #1569, opened 2019-03-29
Remove assertion for createdAt file

l2succes:
Fixes assertion exiting the app


alloy:
Besides removing the assertion, I think you should explicitly handle `nil` being returned as a signal to skip precache installation here https://github.com/artsy/emission/blob/1be93986b653b45865885652e099b4b6d18722bc/Pod/Classes/GraphQL/ARGraphQLQueryCache.m#L156

l2succes:
merge on green

PR #1571, opened 2019-04-02
Fixes button overflow on smaller devices

ashleyjelks:
This PR fixes the UI bug causing the Save button on smaller screen iPhones to extend beyond their containers.

Screen Shot 2019-04-02 at 12 53 54 PM

[Links to LD-546](https://artsyproduct.atlassian.net/browse/LD-546)

[Here's a good post to explain my rationale for implementing the fix this way](https://gedd.ski/post/the-difference-between-width-and-flex-basis/)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 ### Master
 
+- Fixes overflowing Save button on small screen devices - ashley
+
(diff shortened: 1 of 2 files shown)


alloy:
Thanks for the great link! (Perhaps share it in #front-end?)

PR #1573, opened 2019-04-02
Missing tests

ashleyjelks:
This PR revisits a closed PR #1566 to track parts of the City Guide app that are missing tests.

Jira tickets have been created to track the Screens that need additional testing and the URLs to them have been added in this PR.

[Links to LD-529](https://artsyproduct.atlassian.net/browse/LD-529)


alloy:
Not sure why this was self-assigned, but it seems done and it will require changes in #1553 which I want to get started on right away so I’ll take this 👍

PR #1574, opened 2019-04-03
[metaphysics] Pass local timezone to MP

alloy:
So MP can use that as the default timezone for datetimes that should be rendered in the user’s local time (for instance the start time of a live auction).

Most importantly, this will allow us to fix _some_ issues outside of an app release, in the future.

This mimics Reaction https://github.com/artsy/reaction/blob/0aba9cc92bc99464bf1e4209301bd78eba7f39be/src/Artsy/Relay/createRelaySSREnvironment.ts#L70

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 ### Master
 
+- Inlcude local timezone in each MP request so MP can decide to render datetimes in the user’s local time - alloy
 - Fixes overflowing Save button on small screen devices - ashley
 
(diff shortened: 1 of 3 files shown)


alloy:
@sweir27 @ashfurrow Not sure I entirely follow the requirements of https://github.com/artsy/emission/pull/1162, but I _think_ that TZ faffing could have been avoided by letting MP format the time. If that’s the case, you may want to clean that up after this change gets merged.

PR #1575, opened 2019-04-04
Map tests

ashleyjelks:
Enables/updates tests in the Global Map component.

#trivial

### DIFF
--- a/src/lib/Scenes/Fair/Screens/__tests__/FairBooth-tests.tsx
+++ b/src/lib/Scenes/Fair/Screens/__tests__/FairBooth-tests.tsx
@@ -8,7 +8,8 @@ import { FairBoothContainer as FairBooth } from "../FairBooth"
 
 jest.unmock("react-relay")
 
-it("renders properly", async () => {
+// This test is intentionally disabled as of RN 0.59 upgrade
+xit("renders properly", async () => {
   const tree = await renderUntil(
     wrapper => wrapper.find(FairBoothHeader).length > 0,
     <MockRelayRenderer
--- a/src/lib/Scenes/Fair/Screens/__tests__/FairBooth-tests.tsx
+++ b/src/lib/Scenes/Fair/Screens/__tests__/FairBooth-tests.tsx
@@ -8,7 +8,7 @@ import { FairBoothContainer as FairBooth } from "../FairBooth"
 
 jest.unmock("react-relay")
 
-xit("renders properly", async () => {
+it("renders properly", async () => {
   const tree = await renderUntil(
     wrapper => wrapper.find(FairBoothHeader).length > 0,
     <MockRelayRenderer
--- a/src/lib/Scenes/Map/GlobalMap.tsx
+++ b/src/lib/Scenes/Map/GlobalMap.tsx
@@ -604,7 +604,6 @@ export class GlobalMap extends React.Component<Props, State> {
           resizeMode="cover"
           style={{ ...this.backgroundImageSize }}
         />
-        {console.log("TCL: render -> this.props", this.props)}
         <TopButtonsContainer style={{ top: this.props.safeAreaInsets.top }}>
           <Animated.View style={this.moveButtons && { transform: [{ translateY: this.moveButtons }] }}>
             <Flex flexDirection="row" justifyContent="flex-start" alignContent="flex-start" px={3} pt={1}>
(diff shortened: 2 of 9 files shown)


src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -604,6 +604,7 @@ export class GlobalMap extends React.Component<Props, State> {
           resizeMode="cover"
           style={{ ...this.backgroundImageSize }}
         />
+        {console.log("TCL: render -> this.props", this.props)}
```
kierangillen: Not sure if this log is intentionally left

kierangillen:
Looks good other than the loose log!

src/lib/Scenes/Fair/Screens/__tests__/FairBooth-tests.tsx
```diff
@@ -8,7 +8,7 @@ import { FairBoothContainer as FairBooth } from "../FairBooth"
 
 jest.unmock("react-relay")
 
-xit("renders properly", async () => {
+it("renders properly", async () => {
```
alloy: This one is really flaky. I disabled it in the RN 0.59 upgrade PR.

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -246,7 +246,7 @@ export class GlobalMap extends React.Component<Props, State> {
       setTimeout(this.resetZoomAndCamera, 500)
     }
 
-    // If there is a new city, emity it and update our map.
+    // If there is a new city, enity it and update our map.
```
alloy: ?

```suggestion
    // If there is a new city, emit it and update our map.
```

src/lib/Scenes/Map/__tests__/GlobalMap-tests.tsx
```diff
@@ -27,6 +30,9 @@ describe("GlobalMap", () => {
         }
       `,
       mockData: {
+        safeAreaInsets: {
+          top: 1,
+        },
```
alloy: This is a regular prop, not GraphQL data. What you want to do is update the MockRelayRenderer and renderRelayTree utils to the latest versions in Reaction and use the component props https://github.com/artsy/reaction/blob/2308594ff1d34da2a089b816413bed6985928445/src/DevTools/MockRelayRenderer.tsx#L25

src/lib/__fixtures__/CityFixture.ts
```diff
@@ -1,15 +1,113 @@
 export const CityFixture = {
-  name: "Paris",
+  safeAreaInsets: {
+    top: 1,
+    bottom: 1,
+    left: 1,
+    right: 1,
+  },
```
alloy: As per above, this shouldn't be mixed in with the GraphQL data.

src/lib/Scenes/Map/__tests__/GlobalMap-tests.tsx
ashleyjelks: Super helpful, thanks!

ashleyjelks:
> Do you know what’s up with all the Xcode project changes?

Not sure. These were generated after running `bundle exec pod install` I'll uncommit these file changes.

src/lib/Scenes/Map/__tests__/GlobalMap-tests.tsx
```diff
@@ -1,24 +1,32 @@
...
   setAccessToken: jest.fn(),
+  UserTrackingModes: jest.fn(),
 }))
 
 describe("GlobalMap", () => {
-  xit("renders correctly", async () => {
+  it("renders correctly", async () => {
     await renderRelayTree({
-      Component: GlobalMap,
+      componentProps: {
```
kierangillen: We'll want to mock the `viewer` object in this `mockData`. Generally, I think the mockData object should match the top level of the corresponding query. So ```mockData: {
        viewer: viewerFixture,
      }```

ashleyjelks: Not sure I follow what you mean. Would you please elaborate? The `mockData` shape maps this query mockData > viewer > city > cityFixture. 

The props passed to `componentProps` don't go in the mockData object as @alloy pointed out in a previous code review so they are being passed here. The `safeAreaInsets` prop comes Mapbox (not the Graphql data) so it has to be mocked in the `componentProps` field.

kierangillen: Ah yes you're right! Usually viewer is used when wrapping two or more queries so I thought that was the case here.

kierangillen:
Merged!

PR #1578, opened 2019-04-10
[Refactor] Update deprecated Relay `graphql` assignments

damassi:
Runs a jscodemod transform over all of our Relay code to deal with implicit prop assignment deprecation. See https://mobile.twitter.com/kassens/status/1116018336192548865 for more info. 

Related: https://github.com/artsy/reaction/pull/2289. 

**Details:**

Relay has deprecated implicit property assignment in GraphQL calls in favor of explicitness; e.g.,: 

```
graphql`
  fragment Foo_artist on Artist { ... }
`
```
Would automatically attach the response to an `artist` prop. Post codemod, this now looks like this: 

```
artist: graphql`
  fragment Foo_artist on Artist { ... }
`
```

#skip_new_tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Fixes overflowing Save button on small screen devices - ashley
 - Moves fair open hours display to metaphysics - kieran
 - Upgrade to React Native 0.59.2 - orta/alloy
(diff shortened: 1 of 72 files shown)


alloy:
Sweet, thanks!

PR #1581, opened 2019-04-19
Improve Shows Performance

ashleyjelks:
This PR refactors the show artists, show artworks, and show more information query renderers so the data for these components is not requested until they are needed to populate their respective views. 

[Links to LD-557](https://artsyproduct.atlassian.net/browse/LD-557)

Kapture 2019-04-19 at 18 15 11

#skip_new_tests


alloy:
Good catch!

The code looks great, but I did notice that [a test is outputting this error](https://travis-ci.org/artsy/emission/jobs/522322744#L594) (yet not failing /cc @ds300), so that should probably be addressed.

src/lib/Scenes/Show/Screens/MoreInfo.tsx
```diff
@@ -16,6 +19,7 @@ const ListHeaderText = styled(Serif)`
 
 interface Props extends ViewProperties {
   show: MoreInfo_show
+  showID: string
```
l2succes: I don't think this `showID` isn't being used since you have it defined in `ShowMoreInfoRenderer`

src/lib/Scenes/Show/Screens/ShowArtists.tsx
```diff
@@ -1,16 +1,20 @@
...
 import React from "react"
 import { ViewProperties } from "react-native"
-import { createFragmentContainer, graphql } from "react-relay"
+import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
+import { defaultEnvironment } from "../../../relay/createEnvironment"
+import renderWithLoadProgress from "../../../utils/renderWithLoadProgress"
 
 interface Props extends ViewProperties {
   show: ShowArtists_show
+  showID: string
```
l2succes: same ^

PR #1582, opened 2019-04-23
Use MP v2 schema and upgrade to Relay v4

alloy:
#skip_new_tests

Relies on https://github.com/artsy/metaphysics/pull/1678

- [x] Show
- [x] Fair
  - [x] Initial view
  - [x] Pagination
  - Sub-VCs:
    - [x] Artists
    - [x] Exhibitors
    - [x] Works
- [x] Map
- [x] Artwork
- [x] Artist
- [x] Homepage
- [x] Gene
- [x] Gene Refined
- [x] My Profile
- [ ] Start Consignment Flow
- [x] Inbox
- [x] Inquiry
- [x] Favorites
- [x] Bid Flow


alloy:
Came across this when trying to load 2 year old conversations. Could be because of staging data issues _or_ legit ID issues.

Simulator Screen Shot - iPhone 8 - 2019-05-29 at 18 25 25

alloy:
The consignment flow breaks when trying to enter an artist. You can reproduce the failing query with:

```graphql
{
  match_artist(term: "Banksy") {
    id: internalID
    name
    image {
      url
    }
  }
}
```

```json
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field Artist.internalID.",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "match_artist",
        0,
        "id"
      ]
    },
    {
      "message": "Cannot return null for non-nullable field Artist.internalID.",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "match_artist",
        1,
        "id"
      ]
    },
    {
      "message": "Cannot return null for non-nullable field Artist.internalID.",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "match_artist",
        2,
        "id"
      ]
    },
    {
      "message": "Cannot return null for non-nullable field Artist.internalID.",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "match_artist",
        3,
        "id"
      ]
    },
    {
      "message": "Cannot return null for non-nullable field Artist.internalID.",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "match_artist",
        4,
        "id"
      ]
    }
  ],
  "data": {
    "match_artist": [
      null,
      null,
      null,
      null,
      null
    ]
  },
  "extensions": {
    "requests": {
      "gravity": {
        "requests": {
          "match/artists?term=Banksy": {
            "time": "0s 0.927ms",
            "cache": true,
            "length": "N/A"
          }
        }
      }
    },
    "requestID": "5fd925f0-824d-11e9-b503-df30dfba6450"
  }
}
```

zephraph:
I _really_ suspect this is an issue with the query rewriting not quite working with aliases. 

So this query fails (like you pointed out)

```graphql
{
  match_artist(term: "Banksy") {
    id: internalID
 	internalID
    name
    image {
      url
    }
  }
}
```

But this succeeds

```diff
{
  match_artist(term: "Banksy") {
    id: internalID
+	internalID
    name
    image {
      url
    }
  }
}
```

Oddly, this also works (which I think is the more semantically correct here)

```graphql
{
  match_artist(term: "Banksy") {
    id: gravityID
    name
    image {
      url
    }
  }
}
```

alloy:
Right right, aliasing isn’t implemented yet (we have a jira ticket for it). What I meant more is, do we really need the alias there?

zephraph:
Well that query is being used for `ArtistSearchResponse` which corresponds to [SearchResult](https://github.com/artsy/emission/blob/9d7724226704cb9d48038858fb4cdc72641c3fda/src/lib/Components/Consignments/index.tsx#L21)

Changing the ID in that type from `id` to `internalID` causes 13 ts errors. A lot of them related to tests and stories. 

I already pushed an update that changed the alias to use `gravityID` (which I think is the most backwards compatible change anyway) which fixes this issue for now. I think removing the alias is probably more trouble than it's worth.

alloy:
I’m not sure we want to change that to `gravityID`. It could well be that convection requires database IDs, as some other services/endpoints do too. I think we can just get rid of the alias entirely, but first I’ll make sure we have a green test suite as well.

zephraph:
Maybe, but that's used as search results and there are definitely tests/stories that are providing slugs. `id` here was the id from gravity in v1, right? 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

alloy:
No it was previously using `_id`, which is the database ID https://github.com/artsy/emission/commit/7e9253b3847e03dba49ece1f709b2fd252d45588. I’ve made the changes and updated the code/tests 👍

alloy:
There is one remaining problem with consignments, which is when trying to select a result from the ‘Location’ search.

Screenshot 2019-05-30 at 15 46 52

package.json
```diff
@@ -152,10 +149,9 @@
     "react-storybooks-relay-container": "1.2.2",
     "react-test-renderer": "16.8.3",
     "recursive-readdir-sync": "1.0.6",
-    "relay-compiler": "https://github.com/artsy/relay/releases/download/v1.5.0-artsy.5/relay-compiler-1.5.0-artsy.5.tgz",
-    "relay-compiler-language-typescript": "1.1.1",
+    "relay-compiler": "4.0.0",
```
damassi: 🎈

PR #1583, opened 2019-04-30
Removes yellow boxes, adds pull-to-refresh controls, pre-renders sibling tabs for Home+Favs

ashfurrow:
Related to [DISCO-978](https://artsyproduct.atlassian.net/browse/DISCO-978). Here's what's up:

- I removed the yellow boxes from our dev experience, with links back to why we're ignoring those warnings.
- I added pull-to-refresh controls to the favs tab components (Works, Artists, Shows, etc). The control doesn't work on empty data sets, so there's still room for improvement, but having a pull-to-refresh on the fav artworks page is an important step forward.
- ~I specified the Home and Favs tabs to pre-render sibling components on load. This means that, when the app launches on the home tab (artists sub-tab), the WFY and Auctions sub-tabs will load in the background, making the app feel more responsive.~

That last one is the most contentious – I'm hoping to get a 👍 from @orta to make sure we're not doing it for a reason. Otherwise, it seems like a low-effort, high-impact change. 

The components are really similar – I've [opened a ticket](https://artsyproduct.atlassian.net/browse/DISCO-983) to return to them to refactor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 
 ### Master
 
+- Removes yellow box warnings from node_modules - ash
+
 ### 1.10.0
 
 ### 1.10.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### Master
 
 - Removes yellow box warnings from node_modules - ash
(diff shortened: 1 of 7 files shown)


orta:
Yep, looks good 👍

PR #1590, opened 2019-05-16
[AUCT-382] Fixes consignment photo selection

ashfurrow:
Our recent upgrade of React Native caused a problem where photo selection in the consignments screen wasn't working. We still use `CameraRoll` (part of RN) as opposed to [`react-native-cameraroll`](https://github.com/react-native-community/react-native-cameraroll), but they had a similar problem. It seems RN recently switched fro AVFoundation for asset-querying to the new Photos framework. [The fix mentioned here](https://github.com/react-native-community/react-native-cameraroll/issues/18#issuecomment-478175321) fixes things in the simulator, and I think it _should_ work. I'll test the beta, of course.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### 1.11.0
 
 - Bootstrap Artwork component view controller - alloy & david
+- Fixes consignment photo selection - ash
 
 ### 1.10.1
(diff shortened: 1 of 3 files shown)


alloy:
LGTM 👍

PR #1591, opened 2019-05-17
[Lib] Upgrade styled components@4.1.3

damassi:
Part 1 of a two-part update to Emission. This PR upgrades Emission to styled-components 4.1.3, the same version as seen in Palette. (Part 2 is upgrading Palette itself.) 

#### _However_...

Along the way I ran into a (seemingly insurmountable) issue involving our tests that would seem common, but after a lot of googling I couldn't find anything. It involves:

- styled-components@4.x, which enables React's [strict mode](https://reactjs.org/docs/strict-mode.html) under the hood
- styled-system
- react-test-renderer
- jest

Styled System allows us to hook into our theme, and define custom keys for values, such as colors. When upgrading from styled-components 3 to 4, suddenly all of our tests would fail if they involved styled-system due to styled-components passing raw inline props up to react native's style validator. For example: 

```
borderColor='black10'
```

Would fail along the lines of: 

Screen Shot 2019-05-16 at 4 34 01 PM

Digging in to the stack trace, validation seemed to happen before we even got to styled-system's util functions, where the custom key/value pairs would be transformed into valid inputs. 

After trying various things I remembered a link I saw the other day from the creator of styled-system, demoing a new babel plugin that allows one to attach styled-system props to native html elements and have everything compile out to proper values. This seemed like a good fix for the issue we're seeing here, and after testing found it worked well. It seems like a hack, but not sure how else to approach ATM. 

Thoughts / Suggestions? 

#### Update:

Tests were all green locally, but after `rm -rf` and reinstall, no tests pass. I'm not sure what's up but will have to return to this.


alloy:
> Tests were all green locally, but after `rm -rf` and reinstall, no tests pass. I'm not sure what's up but will have to return to this.

Hmm, perhaps a build cache issue? Have you tried `jest --no-cache`?

damassi:
> Hmm, perhaps a build cache issue? Have you tried jest --no-cache

Yup, no change. Really strange! (Will return to this next week.)

damassi:
Closing in favor of https://github.com/artsy/emission/pull/1623

PR #1592, opened 2019-05-17
Camera roll dependency change

ashfurrow:
Follow-up from #1590. This PR moves us from the `CameraRoll` defined in `react-native` to the one defined in `@react-native-community/cameraroll`. This community module is much more well-maintained and has fixed a bug (I haven't found the exact commit or code problem) related to the `getPhotos()` function.

Since this PR doesn't _add_ a new dependency, it just changes where the dependency is coming from, I'm bypassing the usual RFC process around adding new deps.

I was getting a TypeScript error on the included typings – I didn't recognize the error but I followed its suggestion to create a patch. If this looks good, I can PR the change back to them. This was my first time making a patch, it was a very smooth experience – thank you, @ds300 👍


alloy:
All looks good 👍

PR #1593, opened 2019-05-17
PURCHASE-1047: Adds (non-interactive) ArtworkActions to Artwork view

sweir27:
Small change here to give us a baseline for the artwork actions.

I also added a placeholder for the image carousel so we can start adding components to the page and have them fall roughly into place. 😄

Note that this currently uses dummy `SearchIcon`s as we are blocked on incorporating the new icons from palette until it's upgraded. (cc @damassi @alloy )

image


src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -13,8 +15,11 @@ interface Props {
 export class Artwork extends React.Component<Props> {
   render() {
     return (
-      <View style={{ backgroundColor: "red" }}>
-        <Text>{this.props.artwork.title}</Text>
+      <View>
+        <Flex width="100%" style={{ backgroundColor: "gray" }} height={340} />
```
damassi: `bg='black60'`? (`bg` prop should propgate through Flex)

src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
```diff
@@ -0,0 +1,56 @@
...
+import React from "react"
+import { View } from "react-native"
+import { createFragmentContainer, graphql } from "react-relay"
+import styled from "styled-components/native"
+
+interface ArtworkActionsProps {
+  artwork: ArtworkActions_artwork
+}
+
+export class ArtworkActions extends React.Component<ArtworkActionsProps> {
```
damassi: `export const ArtworkActions = React.FC`

src/lib/Scenes/Artwork/Artwork.tsx
sweir27: Oh, yeah. This is completely for dummy-purposes so just did the quickest thing.

PR #1595, opened 2019-05-20
[Beta Deploy] Update color from red to green

l2succes:
We're doing a beta demo

#skip_new_tests #trivial


l2succes:
Merging this

PR #1597, opened 2019-05-20
PURCHASE-1106: Add ArtworkAvailability component and add it to Artwork screen

ashkan18:
# Change
This PR adds new component for artwork availability. For now it just adds it to main section and its not properly styled 🙃  

image


src/lib/Scenes/Artwork/Components/__tests__/ArtworkAvailability-tests.tsx
```diff
@@ -0,0 +1,26 @@
...
+      component
+        .find(Sans)
+        .at(0)
+        .render()
+        .text()
+    ).toMatchInlineSnapshot(`"On Loan"`)
+  })
+})
+
+const artworkActionsArtwork = {
```
sweir27: Do you need `__id` and `_id` here? Doesn't look like you're querying them from graphQL.

You can also rename this fixture as it refers to another component! Or... extract the fixture somewhere and share it between the two.

sweir27:
One small comment but overall looks great!

src/lib/Scenes/Artwork/Components/__tests__/ArtworkAvailability-tests.tsx
ashkan18: let me check! i saw it complain about missing `__id` somewhere but maybe not here?

PR #1600, opened 2019-05-21
Removes extraneous guards, improves Jest setup #trivial

ashfurrow:
Follow-up from #1599. This removes some unneeded ` || {}` and adds a stub value for `NativeModules.Emission` in the Jest setup.


alloy:
Excellent 👌

PR #1601, opened 2019-05-21
Allow .tests to also be picked up by jest

ashkan18:
# Problem
Omakase by default [creates](https://github.com/omakase-js/omakase/blob/master/packages/cli/src/generator/templates/Component.test.tsx.ejs) tests with `.test.tsx` format and emission is currently only picks up `-tests` files.

# Solution
pick all `*tests.(ts|tsx|js)` files.


alloy:
👌

PR #1602, opened 2019-05-21
PURCHASE-1050: As any user I can see an artwork's tombstone

lilyfromseattle:
__User story__

That tombstone section includes: artist name, artwork title, year, medium, dimensions, attribution class.

__Spec:__ https://zpl.io/aX3O198

__Acceptance Criteria:__
- Artist name
- Artwork title, year
- Medium
- Dimensions (use in or cm based on user preference)

__With one artist:__
Screen Shot 2019-05-21 at 7 00 37 PM

__With two artists:__
Screen Shot 2019-05-21 at 7 00 49 PM

Screen Shot 2019-05-21 at 7 00 58 PM


src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
```diff
@@ -0,0 +1,164 @@
...
+      </>
+    )
+  }
+
+  renderMultipleArtists = () => {
+    const {
+      artwork: { artists },
+    } = this.props
+    return artists.map((artist, index) => {
+      const isHidden = index > 2 && !this.state.showingMoreArtists
```
ds300: I wonder whether it would be better if we were to use `this.state.showingMoreArtists` to slice the first three elements out of the `artists` array rather than mapping over the whole thing every time and dealing with hiding extraneous artists in here?

src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
```diff
@@ -0,0 +1,164 @@
...
+      const artistNameWithComma = index !== artists.length - 1 ? artist.name + ", " : artist.name
+      return (
+        <React.Fragment key={artist.__id}>
```
ds300: Could also use the `href` as a key, since it's guaranteed to be unique. Then we don't need to fetch the `__id` too.

src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
```diff
@@ -0,0 +1,164 @@
...
+          {!isHidden && <>{this.renderArtistName(artistNameWithComma, artist.href)}</>}
+        </React.Fragment>
+      )
+    })
+  }
+
+  renderTruncatedArtists = () => {
```
ds300: This name threw me off a little, since this method doesn't actually do the truncation. I wonder whether it would be cleaner to combine this with `renderMultipleArtists`?

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -9,20 +9,23 @@ import { View } from "react-native"
...
 export class Artwork extends React.Component<Props> {
   render() {
+    const { artwork } = this.props
     return (
       <View>
         <Flex width="100%" style={{ backgroundColor: "gray" }} height={340} />
         <Flex alignItems="center" mt={2}>
-          <ArtworkActions artwork={this.props.artwork} />
+          <ArtworkActions artwork={artwork} />
+          <ArtworkTombstone artwork={artwork} />
```
ds300: We need to revise this structure to avoid centering the tombstone and availability

image

Should be justified left with padding=2 (i think, from eyeballing the spec)

src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  render() {
+    const { artwork } = this.props
+    return (
+      <Box textAlign="left">
+        <Flex flexDirection="row" flexWrap="wrap">
+          {artwork.artists.length === 1 ? this.renderSingleArtist(artwork.artists[0]) : this.renderTruncatedArtists()}
+          {artwork.artists.length === 0 && artwork.cultural_maker && this.renderCulturalMaker(artwork.cultural_maker)}
+        </Flex>
+        <Serif color={color("black60")} size="3" m="0" p="0">
```
ds300: The line spacing on these serifs looks too large to me. Maybe it's supposed to be `"3t"`?

src/lib/Scenes/Artwork/Components/__tests__/ArtworkTombstone-tests.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { NativeModules, TouchableWithoutFeedback } from "react-native"
+import { ArtworkTombstone } from "../ArtworkTombstone"
+
+jest.mock("lib/NativeModules/SwitchBoard", () => ({
+  presentNavigationViewController: jest.fn(),
+}))
+
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+
+describe("ArtworkTombstone", () => {
```
ds300: Great tests 🙌 💆

src/lib/Scenes/Artwork/Artwork.tsx
sweir27: I'm OK with following up on this-- once we finish up this entire top-section maybe we can revisit the layout?

ds300: Yeah there's no rush :+1:

src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
```diff
@@ -0,0 +1,164 @@
...
+          </Sans>
+          <Sans color={color("black60")} size="4">
+            Follow
+          </Sans>
+        </Flex>
+      </React.Fragment>
+    )
+  }
+
+  renderArtistName(artistName, href) {
```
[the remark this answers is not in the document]
lilyfromseattle: If I do that will it be expecting me to pass in an Artist? Because I'm formatting the name before I pass it in

src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
lilyfromseattle: I did it like this because I didn't know how to combine them without the "X more" being repeated because the return is inside a loop but I agree the name isn't the best

Example/Podfile.lock
```diff
@@ -289,17 +289,17 @@ SPEC CHECKSUMS:
   ORStackView: b9507271cb41fb9e0b3eecc6414d831201e7cf7c
   Pulley: 2679a2b5a9714c5eb760b1fbb535d15c60e50fb4
   React: 9d063e2f356c8cd2f54dd550d4507740037cbabe
-  react-native-cameraroll: 4c5ca0eab943949e7c39fa549d8f442dae9d3776
+  react-native-cameraroll: b1faef9f2ea27b07bdf818b8043909a93f797eca
```
sweir27: I'm not sure if this will have unintended effects on other installs-- cc @ashfurrow is it OK for this file to change?

ashfurrow: Hmm! It _should_ be okay. What I'd want to look out for is churn in this file (lots of PRs that all change these lines). These values are hashes of the podspecs (similar to gemspecs) and so constantly churning values would indicate that someone's local machine is for some reason generating different hashes.

tl;dr this is totally fine to merge, let's just keep an eye out for further churn in this file 👍

sweir27:
Closed in lieu of: https://github.com/artsy/emission/pull/1609

PR #1607, opened 2019-05-22
Adds entry to example project to fix warning #trivial

ashfurrow:
This adds an entry to fix the warning we get on emission-nebula beta deploys:

Screen Shot 2019-05-22 at 4 15 20 PM


ashfurrow:
Peril pending is a known thing, and the License compliance has been fixed but isn't being reflected here. I'm merging as-is.

PR #1609, opened 2019-05-22
 PURCHASE-1050: As any user I can see an artwork's tombstone

lilyfromseattle:
__NOTE:__ This is a duplicate PR of the same changes because I made the other one from my fork by accident

__User story__

That tombstone section includes: artist name, artwork title, year, medium, dimensions, attribution class.

__Spec:__ https://zpl.io/aX3O198

__Acceptance Criteria:__
- Artist name
- Artwork title, year
- Medium
- Dimensions (use in or cm based on user preference)

__With one artist:__
Screen Shot 2019-05-21 at 7 00 37 PM

__With two artists:__
Screen Shot 2019-05-21 at 7 00 49 PM

Screen Shot 2019-05-21 at 7 00 58 PM


src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  render() {
+    const { artwork } = this.props
+    return (
+      <Box textAlign="left">
+        <Flex flexDirection="row" flexWrap="wrap">
+          {artwork.artists.length === 1 ? this.renderSingleArtist(artwork.artists[0]) : this.renderMultipleArtists()}
+          {artwork.artists.length === 0 && artwork.cultural_maker && this.renderCulturalMaker(artwork.cultural_maker)}
+        </Flex>
+        <Serif color={color("black60")} size="3t" m="0" p="0">
```
ds300: The line spacing / font sizes here are still a bit off. I made a quick mockup to compare the spec with the current implementation, using the x-height of the artist name to calibrate the scaling:

image

And then again using the x-height of the detail text to calibrate:

image

So we can see that the font sizes are too far apart, and the line spacing is still slightly too large on the detail text, although that might be an issue with palette 😬 so maybe we can add another ticket to our backlog to investigate palette text sizes on mobile web vs react native to make things consistent @damassi do you know whether that work has been thought about/attempted before?

sweir27: @ds300 good find, and let's definitely investigate in a separate ticket as it seems outside the scope of this PR.

damassi: I'm curious how this shook out on the local discovery side during their build out /cc @l2succes.

src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
```diff
@@ -0,0 +1,151 @@
...
+        <Serif color={color("black60")} size="3t">
+          {Constants.CurrentLocale === "en_US" ? artwork.dimensions.in : artwork.dimensions.cm}
+        </Serif>
+        {artwork.edition_of && (
+          <Serif color={color("black60")} size="3t">
+            {artwork.edition_of}
+          </Serif>
+        )}
+        {artwork.attribution_class && (
+          <Serif color={color("black60")} size="3t">
```
damassi: Shouldn't need to use the `color` helper everywhere (all theme values should be accessible as strings / numbers). At the top of the react tree can wrap with 

```tsx
import { Theme } from '@artsy/palette'
...
<Theme>
  <App />
</Theme>
```
Which then injects the styled-system context from which palette pulls its values from.

lilyfromseattle: I tried that but it doesn't seem to be working for me

lilyfromseattle: ```Unhandled JS Exception: Invariant Violation: Invalid prop `color` supplied to `StyleSheet generated`: black60
Valid color formats are
  - '#f0f' (#rgb)
  - '#f0fc' (#rgba)
  - '#ff00ff' (#rrggbb)
  - '#ff00ff00' (#rrggbbaa)
  - 'rgb(255, 255, 255)'
  - 'rgba(255, 255, 255, 1.0)'
  - 'hsl(360, 100%, 100%)'
  - 'hsla(360, 100%, 100%, 1.0)'
  - 'transparent'
  - 'red'
  - 0xff00ff00 (0xrrggbbaa)```

damassi: Ah, interesting -- I'm facing something similar in a separate issue, which I thought was related to a styled-components upgrade. It's validating values :(

lilyfromseattle: Nevermind! Just needed to rebase

lilyfromseattle: It actually works ok its just the storybook for the component that gets the error because it doesn't have access to Theme. The storybook for the whole screen works fine.

damassi: Can you try adding this to the top of the story really quick?

```js
storiesOf(...).addDecorator(storyFn => {
    return <Theme>{storyFn()}</Theme>
  })
.add(...) 
```

(the add decorator bit)

Does it still error out without the `color` function?

damassi: (As an aside, how very exciting that this sometimes works -- it means that the styled-components upgrade on my side isn't as catastrophic as it seems...)

lilyfromseattle: I ended up just enclosing them all in Theme but I'll try that instead

lilyfromseattle: It works!

damassi: In reaction we use our own [custom `storiesOf` function](https://github.com/artsy/reaction/blob/master/src/__stories__/storiesOf.js). It's specific to Reaction but something similar could be devised for Emission that provides the `<Theme>`. 

Would suggest setting that up in Emission at some point (in a separate PR) so that y'all don't need to use `addDecorator` everywhere.  cc @ds300

ds300: pretty sure storybook decorators can be configured globally. See e.g. https://storybook.js.org/docs/basics/writing-stories/#using-decorators

ds300: Can also leave that for another ticket though.

sweir27:
I'm not sure what's up with Peril, but everything else is 🍏 here so I'm gonna go ahead and merge! We can ignore the License Compliance check.

src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
damassi: Oh wow, wasn't aware of the global option 👍

Example/Emission/index.storybooks.js
```diff
@@ -1,10 +1,12 @@
...
-import { getStorybookUI, configure } from "@storybook/react-native"
-
+import { getStorybookUI, configure, addDecorator } from "@storybook/react-native"
+import React from "react"
+import { Theme } from "@artsy/palette"
 import { loadStories } from "../../storybook/storyLoader"
 
 // import your stories
 configure(() => {
+  addDecorator(storyFn => <Theme>{storyFn()}</Theme>)
```
damassi: 👍 Now if only there was something like this for enzyme

PR #1619, opened 2019-05-29
Fixes a bug where artist search results in consignments overlap with search bar

yuki24:
fixes https://artsyproduct.atlassian.net/browse/AUCT-318

This fixes an issue where the artist search view get squeezed and search results overlap with the search bar. The issue was that for some reason the parent view no longer defaults to "match parent" behaviour (might be related to RN upgrade) and some styles need to be added to make sure the view covers the entire viewport.

I wonder if both portrait and landscape should be covered, but let me know if they should be.

## Before:

IMG_3957

## After:

ezgif com-optimize


yuki24:
CI is for people like me who breaks builds without noticing. fixing it now.

sweir27:
I think you're fine not worrying about landscape! But given the nature of this ticket I would verify on iPad if you haven't already!

Otherwise, looks good! 🎉

yuki24:
blocked by https://github.com/artsy/emission/pull/1582

yuki24:
Good call on the iPad QA, here is a screenshot:

Screen Shot 2019-05-30 at 10 18 55 AM

dleve123:
@yuki24 I think that @ashfurrow is out at a conference for the rest of the week, so I'm going to re-assign to you so that you can merge considering that @sweir27 reviewed!

zephraph:
This is unblocked now btw.

PR #1622, opened 2019-05-29
[Lib] Point relay fork to Artsy

damassi:
Just noticed installs fail locally after running `yarn cache clean`. Looking around I see that other repos recently updated this as well.


damassi:
Ah, this has already been completed in master. Closing

PR #1625, opened 2019-05-30
PURCHASE-1066: Add new OtherWorks component for showing related works for an Artwork

ashkan18:
# Change
Add  "Other Works by <artist>" section to show 8 artworks by the same artist of the artwork.

fixes: https://artsyproduct.atlassian.net/browse/PURCHASE-1066

# Solution
Added a new `OtherWorks` component which uses `artwork.context.__typename` to determine what type of context we need to show. By default it will show `Other Works By Artist` which is defined in newly added `ArtworkContextArtist` to get artist's artwork and use `GenericGrid` to show the related works.

# Selfie

## Other Works By Artist
image

## Related Works
image

#skip_new_tests : file names for tests travis is complaining about are not `index-test.tsx`


src/lib/Scenes/Artwork/Components/__tests__/OtherWorks-tests.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { ArtworkFixture } from "lib/__fixtures__/ArtworkFixture"
+import React from "react"
+import { ArtworkContextArtistFragmentContainer as ArtworkContextArtist } from "../OtherWorks/ArtworkContexts/ArtworkContextArtist"
+import { OtherWorksFragmentContainer as OtherWorks } from "../OtherWorks/index"
+
+describe("OtherWorks", () => {
+  it("renders ArtistArtworkGrid with correct components", () => {
+    const regularArtwork = {
+      ...ArtworkFixture,
+      context: null,
```
ds300: would be good to add another success test case for `ArtworkContextArtist`

src/lib/Scenes/Artwork/Components/OtherWorks/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+      }
+      default: {
+        return <ArtworkContextArtist artwork={props.artwork} />
+      }
+    }
+  },
+  {
+    artwork: graphql`
+      fragment OtherWorks_artwork on Artwork {
+        id
```
ds300: looks like this `id` is not being used

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/__tests__/__snapshots__/ArtworkContextArtist-tests.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`AllArtists renders properly 1`] = `"<view flexdirection=\\"column\\" alignitems=\\"center\\"><text font-family=\\"AGaramondPro-Regular\\" font-size=\\"22px,32px\\" lineheight=\\"28px,38px\\" class=\\"sc-bwzfXH uTxCW\\" mb=\\"2\\" textalign=\\"center\\">Other works by Abbas Kiarostami</text></view>"`;
```
ds300: did you figure out why this is not rendering the grid?

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkGrids/ArtistArtworkGrid.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export const ArtistArtworkGridFragmentContainer = createFragmentContainer(ArtistArtworkGrid, {
+  artwork: graphql`
+    fragment ArtistArtworkGrid_artwork on Artwork {
+      id
+      artist {
+        name
+        artworks_connection(first: 8, sort: PUBLISHED_AT_DESC, exclude: $excludeArtworkIds)
+          @connection(key: "GenericGrid_artworks_connection") {
+          pageInfo {
```
ds300: looks like `pageInfo` is not being used. Also you don't need the @connection directive since this is not being used in a pagination container

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkContextArtist.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { ArtistArtworkGrid } from "./ArtworkGrids"
+
+export const ArtworkContextArtistFragmentContainer = createFragmentContainer<{
+  artwork: ArtworkContextArtist_artwork
+}>(
+  props => {
+    return (
+      <Join separator={<Spacer my={2} />}>
+        <ArtistArtworkGrid artwork={props.artwork} />
+      </Join>
```
ds300: will the `Join` have more children soon? Because `Join` with one child is a no-op.

Example/Pods/Pods.xcodeproj/project.pbxproj
```diff
@@ -2323,7 +2324,7 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXFileReference section */
-		0000000002D0 /* Podfile */ = {isa = PBXFileReference; explicitFileType = text.script.ruby; includeInIndex = 1; indentWidth = 2; lastKnownFileType = text; name = Podfile; path = ../Podfile; sourceTree = SOURCE_ROOT; tabWidth = 2; xcLanguageSpecificationIdentifier = xcode.lang.ruby; };
+		0000000002D0 /* Podfile */ = {isa = PBXFileReference; explicitFileType = text.script.ruby; includeInIndex = 1; indentWidth = 2; name = Podfile; path = ../Podfile; sourceTree = SOURCE_ROOT; tabWidth = 2; xcLanguageSpecificationIdentifier = xcode.lang.ruby; };
```
ds300: I wouldn't expect to see changes in the file if you didn't change any native deps. @ashfurrow @alloy can either of you tell if these changes are kosher?

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkGrids/ArtistArtworkGrid.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const {
+      artwork: { artist },
+    } = this.props
+
+    if (!artist) {
+      return null
+    }
+    const artworks = artist.artworks_connection.edges.map(({ node }) => node)
+
+    return (
```
ds300: this should probably also return null if artworks has length 0?

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/__tests__/__snapshots__/ArtworkContextArtist-tests.tsx.snap
ashkan18: not really, maybe because i stop at `Header`?  do you think i should just switch to a normal test and ignore this snapshot magic?

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkContextArtist.tsx
ashkan18: Just talked to @sweir27 going to add two more grids.

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/__tests__/__snapshots__/ArtworkContextArtist-tests.tsx.snap
ds300: Yeah. And I think the test should make sure that the grid is there and has the expected number of cards in it.

sweir27:
WOW! So exciting!!! This looks great, and thanks for setting the pattern for us to add all the rest of the grids 😊.

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/__tests__/ArtworkContextArtist-tests.tsx
```diff
@@ -0,0 +1,16 @@
+import { mount } from "enzyme"
+import { ArtworkFixture } from "lib/__fixtures__/ArtworkFixture"
+import React from "react"
+import { ArtworkContextArtistFragmentContainer as ArtworkContextArtist } from "../ArtworkContextArtist"
+import { ArtistArtworkGrid } from "../ArtworkGrids/ArtistArtworkGrid"
+import { PartnerArtworkGrid } from "../ArtworkGrids/PartnerArtworkGrid"
+import { RelatedArtworkGrid } from "../ArtworkGrids/RelatedArtworkGrid"
+
+describe("PartnerArtworkGrid", () => {
+  it("renders PartnerArtworkGrid with correct components", () => {
```
ds300: `PartnerArtworkGrid => ArtworkContextArtist`?

src/lib/Scenes/Artwork/Components/__tests__/OtherWorks-tests.tsx
ds300: i.e. with

```
context: {
   __typename: "ArtworkContextArtist",
},
```

rather than

```
context: null,
```

Example/Pods/Pods.xcodeproj/project.pbxproj
alloy: Hmm, unsure what would cause these. They don’t look problematic, all seem related to non-critical metadata that Xcode.app updates 🤔 In short, I’m ok with checking these in, but let’s keep an eye on if this keeps changing with each PR.

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkGrids/ArtistArtworkGrid.tsx
```diff
@@ -0,0 +1,51 @@
...
+    } = this.props
+
+    if (!artist) {
+      return null
+    }
+
+    const artworks = artist.artworks_connection.edges.map(({ node }) => node)
+    if (!artworks || _.isEmpty(artworks)) {
+      return null
+    }
```
alloy: FYI `isEmpty` also does a null check https://github.com/lodash/lodash/blob/4.17.11/lodash.js#L11481-L11483

src/lib/Scenes/Artwork/Components/__tests__/OtherWorks-tests.tsx
ashkan18: ended up adding a separate test for that context, i saw `null` context in real data so going to keep that test around too 🙌

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkGrids/ArtistArtworkGrid.tsx
ashkan18: nice! makes sense!

ashkan18:
thanks all! going to merge on green

PR #1627, opened 2019-05-31
Fix consignments failure after v2 update

zephraph:
Consignments started failing after we moved emission over to v2. 

See the [slack thread](https://artsy.slack.com/archives/C02BAQ5K7/p1559320357036400) for a more context overview. 

Ultimately the `SearchResult` type was being overloaded for data from two different sources. They just so happened to have the same api, but when we updated emission to v2 it changed the api for one but not the other... since we were relying on typescript to help us fix issues, it all sorta broke.


yuki24:
So it seems like now Cloudflare is blocking requests to MP? I got the following error:

Screen Shot 2019-05-31 at 3 44 32 PM

zephraph:
This is ready other than the cloudflare issue.

alloy:
Aha, I see, that makes sense then. Fix looks good, thanks!

PR #1631, opened 2019-06-03
PURCHASE-1063: Add about this artist section to Artwork page

ashkan18:
# Change
Add artist card to Artwork page:
https://artsyproduct.atlassian.net/browse/PURCHASE-1063

# Selfie
image


src/lib/Scenes/Artwork/Components/AboutArtist.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export class AboutArtist extends React.Component<AboutArtistProps> {
+  render() {
+    return (
+      <Flex alignItems="flex-start" mt={2}>
+        <Sans size="4t" weight="medium" mb={2}>
+          About the artist
+        </Sans>
+        <ArtistListItem artist={this.props.artwork.artist} />
```
sweir27: We talked about this in slack but putting a note that this should iterate through all of the artwork's artists and render a card for each!

ashkan18: 👍 this should now support multiple artists.

src/lib/Scenes/Artwork/Components/AboutArtist.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { AboutArtist_artwork } from "__generated__/AboutArtist_artwork.graphql"
+import { ArtistListItemContainer as ArtistListItem } from "lib/Components/ArtistListItem"
+import React from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+
+interface AboutArtistProps {
+  artwork: AboutArtist_artwork
+}
+
+export class AboutArtist extends React.Component<AboutArtistProps> {
```
damassi: @ashkan18 - We should avoid using classes where possible and stick to functional components. (Since the release of hooks classes have been effectively deprecated.)

PR #1632, opened 2019-06-03
PURCHASE-1051: As any user I can follow an artist from an artwork's tombstone

lilyfromseattle:
__User story__

And in those tombstone details I can see the artist name
And next to the artist name I can tap “follow” to follow the artist
When I tap “follow” the text changes to “following”
If I am already following, the text should already be “following”
If i tap when it is “following”, I unfollow the artist.

__Spec__

https://zpl.io/aX3O198

Screen Shot 2019-06-03 at 4 32 02 PM
Screen Shot 2019-06-03 at 4 32 07 PM

#skip_new_tests


src/lib/Scenes/Artwork/Components/__stories__/ArtworkTombstone.story.tsx
```diff
@@ -1,148 +1,44 @@
...
+    },
+  ],
+  cultural_maker: null,
+  dimensions: { in: "15 × 20 in", cm: "38.1 × 50.8 cm" },
+  edition_of: "Edition 100/200",
+  attribution_class: { short_description: "This is an edition of something" },
+  " $refType": null,
+}
+storiesOf("Artwork/Components").add("New Component", () => {
+  return <MockArtworkTombstone artwork={artworkTombstoneArtwork} />
```
sweir27: @ds300 when you get in tomorrow... it would be ideal if we could use `MockRelayRenderer` here, but importing `MockRelayRenderer` seems to cause storybooks to not compile 😮. Wondering if you have any ideas here/would mind taking a look?

image

src/lib/Scenes/Artwork/Components/__tests__/ArtworkTombstone-tests.tsx
```diff
@@ -180,22 +180,50 @@ const artworkTombstoneArtwork = {
...
+      " $refType": null,
+      " $fragmentRefs": null,
+    },
+    {
+      name: "Banksy",
+      href: "/artist/banksy",
+      id: "3468",
+      gravityID: "banksy",
+      is_followed: false,
+      " $refType": null,
```
sweir27: I don't _think_ you should need the ` $refType`s here

src/lib/Scenes/Artwork/Components/FollowArtistButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+  artist: FollowArtistButton_artist
+  relay: RelayProp
+}
+
+interface State {
+  isFollowedChanging: boolean
+}
+
+export class FollowArtistButton extends React.Component<Props, State> {
+  state = { isFollowedChanging: false }
```
ds300: Seems like `isFollowedChanging` isn't being read anywhere, only written to. Can we delete it?

sweir27:
@lilyfromseattle if you add `#skip_new_tests` to your PR description, we should be able to re-run that job and get a 🍏.

PR #1637, opened 2019-06-04
PURCHASE-1044 ImageCarousel Part 1

ds300:
This is just the version you see on the artwork page, no fullscreen mode yet, as discussed on slack.

## ImageCarousel
carousel

## With slow loading images
carousel-slow-loading

I simplified the pagination dots since I showed them at the knowledge share. That implementation was quite complicated and not noticably better than doing the simplest thing so I did the simplest thing instead.

#skip_new_tests


src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { createFragmentContainer, graphql } from "react-relay"
+import { findClosestIndex, getMeasurements } from "./geometry"
+import { ImageWithLoadingState } from "./ImageWithLoadingState"
+
+export interface ImageCarouselProps {
+  images: ImageCarousel_images
+}
+
+const windowWidth = Dimensions.get("window").width
+const cardHeight = windowWidth >= 375 ? 340 : 290
```
sweir27: Mind adding a comment to explain where these values come from?

src/lib/Scenes/Artwork/Components/ImageCarousel/__tests__/ImageCarousel-tests.tsx
```diff
@@ -0,0 +1,170 @@
...
+      variables: {
+        screenWidth: 234,
+      },
+      mockData: {
+        artwork,
+      },
+    })
+  }
+  describe("with five images", () => {
+    it("renders a flat list with eight entries", async () => {
```
sweir27: `five entries`?

src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
```diff
@@ -0,0 +1,111 @@
...
+export interface ImageCarouselProps {
+  images: ImageCarousel_images
+}
+
+const windowWidth = Dimensions.get("window").width
+// The logic for cardHeight comes from the zeplin spec https://zpl.io/25JLX0Q
+const cardHeight = windowWidth >= 375 ? 340 : 290
+export const cardBoundingBox = { width: windowWidth, height: cardHeight }
+
+export const ImageCarousel: React.FC<ImageCarouselProps> = ({ images }) => {
```
sweir27: Per the convo in slack, mind adding a comment here with a short description of what a future developer should keep in mind if they need to extract this for other use-cases?

PR #1641, opened 2019-06-05
Xcode Project Housekeeping #trivial

ashfurrow:
This PR is to narrow down possible causes of our automated build failures, related to [this bug report](https://github.com/fastlane/fastlane/issues/14798). Here's what I've done:

- Update deployment target of Emission app to iOS 10.
- Marks Xcode project as having skipped the Swift 5 conversion (we have no Swift, not sure why the warning was present).
- Skips a recommended change to add whole module optimizations (let's not rock the boat until we get deploys fixed).
- Updates our bundle, namely CocoaPods 1.7.1.
- Uses a CocoaPods `post_install` hook to mark Pulley and FBSnapshotTestCase as also having skipped the Swift 5 conversion.

This means that the only warning in Xcode is a linker warning about duplicated method names in categories from React Native and our camera roll manager dependency:

Screen Shot 2019-06-05 at 5 16 27 PM

I'm betting we can remove those, too, and get a clean build again 🎉


Example/Podfile
```diff
@@ -15,10 +15,7 @@ source 'https://github.com/artsy/Specs.git'
 source 'https://github.com/CocoaPods/Specs.git'
 
 inhibit_all_warnings!
-platform :ios, '9.0'
-
-# FIXME: Figure out why we are getting hit by this issue.
-install! 'cocoapods', deterministic_uuids: false
```
ashfurrow: This didn't seem necessary anymore – others have reported not needing it, but I wanted to highlight this change /cc @alloy

alloy:
LGTM 👍

PR #1644, opened 2019-06-06
[WIP] [tsc] Enable noImplicitAny

alloy:
Got to a part where I got stuck because of styled-components/Palette, for which I then [started an upgrade](https://github.com/artsy/palette/pull/498) but also hit a wall. Perhaps that can be avoided altogether, which I will take a stab at next time.


damassi:
@alloy - now that the upgrade is in can this work continue?

alloy:
Unsure what upgrade you mean, but if you’re interested in continuing on this then by all means please do!

damassi:
@alloy meaning 

> Got to a part where I got stuck because of styled-components/Palette, for which I then started an upgrade but also hit a wall.

Now that we're using the latest styled-components and palette was wondering if things might be unstuck.

alloy:
Ah I see, didn’t know that happened. Yeah I’ll try to poke it again soon 👍

ds300:
We've added a ticket to our backlog to get this done once and for all, but it will happen after merging emission into eigen so I'm going to close this and pilfer the approach in a few weeks 🙏

PR #1654, opened 2019-06-10
PURCHASE-1139: ReadMore component based on Markdown renderer

sweir27:
Most of the work by @ds300 ! I'm just the one writing this up. 😄 

This PR:
- Refactors the existing `Markdown` component and extracts the renderer into a shared location.
- Adds another utility `plainTextFromTree` which pulls out all of the text nodes of an arbitrary react tree.
- Adds a `ReadMore` component. It's so tied to emission's own Markdown renderer, that it didn't make sense to extract this into palette at this time.

image

#skip_new_tests


package.json
```diff
@@ -136,7 +136,7 @@
     "danger": "6.1.1",
     "danger-plugin-jest": "1.1.0",
     "enzyme": "3.9.0",
-    "enzyme-adapter-react-16": "1.11.2",
+    "enzyme-adapter-react-16": "1.13.0",
```
sweir27: Updated this because newer versions have better support for `React.memo`

src/lib/Components/ReadMore.tsx
```diff
@@ -0,0 +1,121 @@
...
+      )
+    },
+  },
+}
+
+export const ReadMore = React.memo(({ content, maxChars }: Props) => {
+  const [isExpanded, setIsExpanded] = useState(false)
+  const root = renderMarkdown(content, rules)
+
+  const plainTextVersion = plainTextFromTree(root)
```
sweir27: @ds300 I pulled this into a separate util because it could be used elsewhere, but also it didn't seem worth muddying up the existing `truncate` component.

Not tied to this implementation at all, though.

src/lib/utils/plainTextFromTree.ts
```diff
@@ -0,0 +1,24 @@
+import React from "react"
+
+export function plainTextFromTree(root: React.ReactNode): string | any {
+  function traverse(node: React.ReactNode, currentProgress: string): string {
```
ds300: does this need the `currentProgress` parameter?

src/lib/utils/__tests__/plainTextFromTree-tests.tsx
```diff
@@ -0,0 +1,34 @@
...
+          <Text>very </Text>
+          <Text>broken </Text>
+          <Text>up </Text>
+        </Flex>
+        <Text>but that's OK</Text>
+      </View>
+    )
+
+    expect(plainTextFromTree(component)).toEqual("Hi this is very broken up but that's OK")
+  })
```
ds300: Could we add a test that operates on the result of `renderMarkdown`?

src/lib/utils/renderMarkdown.tsx
```diff
@@ -0,0 +1,165 @@
...
+      return (
+        <Sans mb="1" key={state.key} size={size}>
+          {output(node.content, state)}
+        </Sans>
+      )
+    },
+  },
+}
+
+export function renderMarkdown(markdown: string, rules: any = defaultRules): React.ReactElement {
```
ds300: Do you think we should add some unit tests for this?

src/lib/utils/plainTextFromTree.ts
```diff
@@ -0,0 +1,24 @@
...
+      for (const child of node) {
+        result += traverse(child, currentProgress)
+      }
+      return result
+    }
+
+    if (React.isValidElement(node)) {
+      const children = React.Children.toArray((node.props as any).children)
+      return traverse(children, currentProgress)
+    }
```
ds300: it just occurred to me that we're not handling nulls, numbers, booleans, and other truthy objects properly. It should probably be:

- null/undefined/true/false - do not get rendered
- arrays - get traversed
- react elements - have their children traversed if present
- everything else - gets coerced to a string

src/lib/Components/ReadMore.tsx
```diff
@@ -0,0 +1,121 @@
...
+              </Sans>
+            </LinkText>,
+          ])
+        }
+        textDepth -= 1
+      }
+
+      return React.cloneElement(node, null, truncatedChildren)
+    }
+  }
```
ds300: it just occurred to me that we're not handling nulls, numbers, booleans, and other truthy objects properly. It should probably be:

- null/undefined/true/false - do not get rendered
- arrays - get traversed
- react elements - have their children traversed if present
- everything else - gets coerced to a string

src/lib/utils/plainTextFromTree.ts
sweir27: By "do not get rendered", would you expect it to throw an error? In this case it seems like it could just return `null` or `undefined`?

i.e.
```javascript
if (node === null || typeof node === "boolean" || typeof node === "undefined") {
  return null
}
```

src/lib/utils/renderMarkdown.tsx
sweir27: They probably already exist for the `Markdown` component but I'll add some here.

src/lib/Components/ReadMore.tsx
sweir27: (similar question as below). Should we throw an error if it's null/undefined/true/false? Or just return null/undefined?

src/lib/utils/plainTextFromTree.ts
ds300: No it wouldn't throw an error, we want to replicate what react would render if you put the value, e.g. in an empty div. So for null/undefined/true/false it would be the empty string

ds300:
@sweir27 This looks great! I noticed a few things playing around with it this morning. Added a couple more commits and fixed some warnings so have assigned it back to you, feel free to merge if you're OK with my changes! 🙏

sweir27:
Thank you, @ds300 !

PR #1655, opened 2019-06-11
PURCHASE-1060/PURCHASE-1059: As a user interested in a work, I can follow or unfollow the work's gallery

lilyfromseattle:
__NOTE:__ depends on Palette changes in https://github.com/artsy/palette/pull/509

Screen Shot 2019-06-11 at 11 49 20 AM
Screen Shot 2019-06-11 at 12 04 04 PM

__User story__
As a user, I can tap the “follow” button to follow, or “unfollow” button to unfollow the gallery/institution/partner.

__Spec__
https://zpl.io/2Egzy5y


src/lib/Scenes/Artwork/Components/PartnerCard.tsx
```diff
@@ -0,0 +1,147 @@
...
+    const {
+      gravityID: partnerSlug,
+      profile: { is_followed: partnerFollowed, internalID: profileID },
+    } = artwork.partner
+
+    this.setState(
+      {
+        isFollowedChanging: true,
+      },
+      () => {
```
alloy: Nit: no need to wait for React to re-render the UI until you send off the network request, these things can be done in parallel.

src/lib/Scenes/Artwork/Components/PartnerCard.tsx
```diff
@@ -0,0 +1,147 @@
...
+        internalID
+        id
+        href
+        initials
+        profile {
+          internalID
+          gravityID
+          is_followed
+          icon {
+            url(version: "square140")
```
alloy: Don’t request specific resolution URLs, the `OpaqueImage` component handles that based on the actual layout and screen density. Instead simply request the url without args and make sure your image component is laid out correctly.

lilyfromseattle: For some reason when I remove the resolution it breaks the circular crop of the image. I'm specifying a width and a height and it appears the same size its just not circular. Do you know why this might be happening?

alloy: Unsure off-hand, as I have no experience with that component. Perhaps @l2succes does?

l2succes: I took a quick look at the `Avatar` component with Lily yesterday, despite the fact that it's not using `OpaqueImage` I could find an obvious reason why this is happening. My suggestion was to track down where the `corner-radius` css prop was coming from first.

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -34,11 +36,13 @@ export class Artwork extends React.Component<Props> {
           </Flex>
           <Separator />
           <Join separator={<Spacer my={2} />}>
+            <ReadMore content={artwork.artist.biography_blurb.text} maxChars={140} />
```
sweir27: This already exists in the `AboutArtist` component, I'm not sure why it got added here but we should remove!

Looks like we probably don't need the query for these fields below either. That likely got kept in there when I moved that component! 😞

src/lib/Scenes/Artwork/Components/PartnerCard.tsx
alloy: Ok, if no solution can be found right away, can you file a ticket @l2succes with the context you have? As this is tech debt that will proliferate and definitely not something we should be doing.

PR #1672, opened 2019-06-17
 PURCHASE-1049 - Artwork share button

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1049
- Artwork share button now shares `"[Artwork Title] by [Artist Name] on Artsy http://artsy.net/artwork/artist-name-artwork-title"`


Example/Pods/Pods.xcodeproj/project.pbxproj
```diff
@@ -15,6 +15,7 @@
 			dependencies = (
 			);
 			name = "boost-for-react-native";
+			productName = "boost-for-react-native";
```
kierangillen: This line keeps getting added, I had checked with Orta in the past if it was an issue and he gave it the green light that it shouldn't effect anything.

src/lib/Scenes/Artwork/Components/__stories__/ArtworkAvailability.story.tsx
```diff
@@ -3,5 +3,12 @@ import React from "react"
 import { ArtworkAvailability } from "../ArtworkAvailability"
 
 storiesOf("Artwork/Components").add("Availability", () => {
-  return <ArtworkAvailability artwork={{ id: "Blah", _id: "HI", availability: "On Loan", " $refType": null }} />
+  return (
```
kierangillen: Was getting a TS error on this line because we're not actually importing any of the other fields on the relay query.

src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
```diff
@@ -32,11 +32,32 @@ export class ArtworkActions extends React.Component<ArtworkActionsProps> {
     })
   }
 
+  handleArtworkShare = async () => {
+    const { title, artist: name, id } = this.props.artwork
+
+    try {
+      const result = await Share.share({
+        message: `${title} by ${name} on Artsy http://artsy.net/artwork/${id}`,
```
sweir27: This will only work for artworks with a single artist! We should check with @samchieng for how she wants to treat this case.

src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
```diff
@@ -32,11 +32,32 @@ export class ArtworkActions extends React.Component<ArtworkActionsProps> {
...
+        if (result.activityType) {
+          // shared with activity type of result.activityType
+        } else {
+          // shared
+        }
+      } else if (result.action === Share.dismissedAction) {
+        // dismissed
+      }
+    } catch (error) {
+      console.warn("Error sharing artwork")
```
sweir27: Elsewhere we have a pattern for logging errors that's a little more informative-- would be good to use here:
```
console.error("ArtworkActions.tsx", error)
```

src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
```diff
@@ -32,11 +32,32 @@ export class ArtworkActions extends React.Component<ArtworkActionsProps> {
...
 
+  handleArtworkShare = async () => {
+    const { title, artist: name, id } = this.props.artwork
+
+    try {
+      const result = await Share.share({
+        message: `${title} by ${name} on Artsy http://artsy.net/artwork/${id}`,
+      })
+      // TODO: Leaving these options here if we're adding analyics
+      if (result.action === Share.sharedAction) {
```
sweir27: I'm not sold on leaving this dead code in here, but we probably will end up tracking it!

src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
```diff
@@ -32,11 +32,32 @@ export class ArtworkActions extends React.Component<ArtworkActionsProps> {
     })
   }
 
+  handleArtworkShare = async () => {
+    const { title, artist: name, id } = this.props.artwork
+
+    try {
+      const result = await Share.share({
```
sweir27: Wow, so using this will give us everything we want? 😱. So cool

src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
samchieng: Let's just show up to 3 artists max. I don't think it's as important for the user to see past that.

"[Artwork Title] by [Artist Name], [Artist Name], [Artist Name] on Artsy [URL]"

src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
```diff
@@ -32,11 +32,35 @@ export class ArtworkActions extends React.Component<ArtworkActionsProps> {
...
+    const { title, id, artists } = this.props.artwork
+    let message
+    if (artists && artists.length) {
+      const names = []
+      artists.forEach((artist, i) => {
+        if (i < 3) {
+          names.push(artist.name)
+        }
+      })
+      message = `${title} by ${names.join(", ")} on Artsy http://artsy.net/artwork/${id}`
```
sweir27: #minor but these should probably all be `https`.

Also, it seemed non-trivial to add tests for the share functionality but would be good to have some unit tests for the logic that forms this message!

sweir27: Related-- this `id` will be the global ID so we should either query for `href` or use the `gravityID` here.

sweir27:
Nice!! ✨

PR #1673, opened 2019-06-18
Update to relay v4 dt types

alloy:
~~This can be finished once https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36212 is in.~~ These are released now.

#skip_new_tests


alloy:
Actually, going to update this to the v5 version that's already underway https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36266

alloy:
Nope, scratch that, we haven’t upgraded Emission to Relay v5 yet.

src/lib/Scenes/City/CitySavedList.tsx
```diff
@@ -65,7 +65,7 @@ class CitySavedList extends React.Component<Props, State> {
           cityName={name}
           bucket={edges.map(e => e.node) as any}
           type="saved"
-          relay={relay}
+          relay={relay as RelayProp}
```
damassi: This is good to know

src/lib/tests/MockRelayRenderer.tsx
```diff
@@ -4,10 +4,9 @@ import { QueryRenderer } from "react-relay"
 import {
   Environment,
   GraphQLTaggedNode,
-  OperationBase,
-  OperationDefaults,
+  INetwork as RelayNetwork,
```
damassi: I wonder what their thinking was with this naming 1-off?

src/lib/Scenes/City/CitySavedList.tsx
alloy: I wouldn’t bother remembering this too much, the passing around of the relay prop that is being done here should really be replaced with a `SystemContext`, like in Reaction–also because in reality it probably only needs the env for a mutation.

damassi:
Looks good 👍

src/lib/tests/MockRelayRenderer.tsx
alloy: I know, right? Bothers me. way. too. much.

src/lib/Scenes/City/CitySavedList.tsx
l2succes: Right, that makes sense, we had talked about using the context API to pass relay down instead here.

PR #1674, opened 2019-06-18
Refactors Artwork view to use FlatList

kierangillen:
- Minor styling adjustments, adding padding, margin and font sizes
- Refactors Artwork page to use `FlatList` which will optimize page render to only render visible sections
- Removes relay fragment containers when unnecessary and passes props directly to components


src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -1,47 +1,126 @@
...
+  }>
+}
+
 export class Artwork extends React.Component<Props> {
-  render() {
+  state: State = {
+    sections: [],
+  }
+
+  componentDidMount() {
```
sweir27: @l2succes can you elaborate on why we need to append the sections in `componentDidMount` vs. in the render function?

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -1,47 +1,126 @@
...
+      case "header":
+        return <ArtworkHeader artwork={artwork} />
+      case "tombstone":
+        return <ArtworkTombstone artwork={artwork} />
+      case "availability":
+        return <ArtworkAvailability availability={data} />
+      case "sellerInfo":
+        return <SellerInfo name={data} />
+      case "aboutWork":
+        return <AboutWork additionalInformation={data.additionalInformation} description={data.description} />
```
sweir27: We talked about this over slack but putting here for posterity. 😄

For components that _are_ tied to the data they fetch, I'm not sure I see the need for passing in props here vs. leaving it as a fragment container. In general it seems best to utilize relay's tools, and in this case we're just adding overhead-- if someone were to delete the additional information component, they'd have to know to delete these specific fields from the query vs. simply removing the fragment.

I'll default to @alloy or @l2succes in case there are things I'm missing tho!

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -1,47 +1,126 @@
...
+
+  renderItem = ({ item: { data, type } }) => {
     const { artwork } = this.props
+    switch (type) {
+      case "header":
+        return <ArtworkHeader artwork={artwork} />
+      case "tombstone":
+        return <ArtworkTombstone artwork={artwork} />
+      case "availability":
+        return <ArtworkAvailability availability={data} />
```
sweir27: This component in particular is small but was explicitly built this way because we will be adding more logic/data to it as we expand the artwork page to handle commercial works. It'll make those additions much easier if we leave it as a fragment container!

sweir27:
(Oops, should have left my comments as a review!)

I have some questions, but it looks soooo pretty 😍!

src/lib/Scenes/Artwork/Artwork.tsx
alloy: I 100% agree with @sweir27.

Additionally this overhead is worse because `data` seems to be untyped, which of course can be fixed, but is one more piece of overhead.

alloy: I’m interested to learn why this change was made, though?

src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx
```diff
@@ -0,0 +1,24 @@
+import { Box, Flex, Spacer } from "@artsy/palette"
+import { Artwork_artwork } from "__generated__/Artwork_artwork.graphql"
+import React from "react"
+import { ArtworkActionsFragmentContainer as ArtworkActions } from "./ArtworkActions"
+import { ImageCarouselFragmentContainer as ImageCarousel } from "./ImageCarousel/ImageCarousel"
+
+interface ArtworkHeaderProps {
+  artwork: Artwork_artwork
```
alloy: You should try to avoid using types of a fragment container that is not used in this very module. As the aforementioned comments, simply add a fragment container here that fetched just what you need, which keeps everything very clear.

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -1,47 +1,126 @@
...
+    })
+
+    sections.push({
+      type: "otherWorks",
+    })
+
+    this.setState({ sections })
+  }
+
+  renderItem = ({ item: { data, type } }) => {
```
alloy: Add the appropriate ‘item’ typing from the RN types.

src/lib/Scenes/Artwork/Artwork.tsx
alloy: +1 especially because the `setState` call will trigger a new render pass immediately after the one that led to `componentDidMount`.

src/lib/Scenes/Artwork/Artwork.tsx
alloy: Btw, if you were to find the right data from your props here inline instead of first making a list of sections, then you also don’t need to type all of the `data` fields. Additionally the list of sections could just be a static (module level) constant array, no need to use state for that at all (which it isn’t really anyways, it’s just a different view on your props).

src/lib/Scenes/Artwork/Artwork.tsx
kierangillen: My understanding of the `FlatList` component is that it will create a render pass on `renderItem` for each index of the data array. Therefore we wouldn't be able to use `ItemSeparatorComponent` or similar props if the data array isn't the exact length of the sections we want to render. So I'm not sure why we would want to have a static data array. My thought was that we would want to pull the data out of the `Artwork` relay query that is relevant to which sections we want to render so the sections are the correct length. Would love to pair to clear this up.

src/lib/Scenes/Artwork/Artwork.tsx
alloy: Ah, my bad, when I reviewed this earlier from my phone the context got lost on me a bit. So seeing as there are some conditionals, the list of sections needs to indeed be determined based on the data, but the actual data for each section doesn’t need to be captured here. (Will elaborate in a pairing session.)

src/lib/Scenes/Artwork/Artwork.tsx
l2succes: I'll defer to @sweir27 for context on edge cases of the Artwork pages. My only suggestion is: if it's possible for the artwork view to a variable number of sections. For instance, in the case that there are no other works, @kierangillen's current implementation would be the most flexible, namely pushing sections onto an array. We found that to work best for the Show and Fair views.

But if you can count on always having the same number of sections and section types for all artworks, as @alloy and @sweir27 mentioned, using a static array and leaving all section wrapped in fragment containers is preferable.

alloy: Sections should be determined in the component that renders the list, as this implementation currently does and which I acknowledged [here](https://github.com/artsy/emission/pull/1674#discussion_r295334250), but that’s orthogonal to using fragment containers or not. Perhaps I’m missing something in your reasoning @l2succes?

src/lib/Scenes/Artwork/Components/SellerInfo.tsx
```diff
@@ -9,11 +9,9 @@ interface SellerInfoProps {
...
 export class SellerInfo extends React.Component<SellerInfoProps> {
   render() {
-    return (
-      <Box>
-        <Sans size="3">{this.props.artwork.partner.name}</Sans>
-      </Box>
-    )
+    const { partner } = this.props.artwork
+    const { name } = partner
+    return <Box>{name && <Sans size="3">{name}</Sans>}</Box>
```
sweir27: Instead of returning an empty Box when the name isn't there, can we check for `name` before rendering anything? Not sure if there's some hidden reason for doing it the current way.

kierangillen: I'm assuming there's going to be more rendered in this component and just checking for the partner name for now, the component shouldn't be rendered at all if the relevant data is missing so the box wont be rendered in that case either way.

alloy: Drive-by FYI: if at some point you need more than, let’s say, two fields to determine if a section should be shown and that can semantically accurately be captured as a single predicate field in MP, then moving it there may become preferable.

src/lib/Scenes/Artwork/Artwork.tsx
l2succes: no need to use `setState` here, it definitely makes more sense to have it in the render function.

src/lib/Scenes/Artwork/Artwork.tsx
l2succes: No I think we're on the same page! Sorry for the confusion, I was just answering @kierangillen's point about how to create the sections array based on relay data. 

re: fragment containers. I agree with @sweir27 as well

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -42,10 +45,15 @@ export class Artwork extends React.Component<Props> {
     }
 
     sections.push("details")
-
     sections.push("aboutArtist")
 
-    sections.push("otherWorks")
+    if (
```
sweir27: This logic in particular is only going to get more complex as we build out support for more contexts so I worry about having to add to this logic here.

What's the risk of always pushing on this section?

kierangillen: We don't want to push any section if there's no supporting data. I think we should keep adding checks here and see how things land, and we can always move logic to metaphysics if necessary.

sweir27: But what happens if we do? I'm curious what the impact will be on the added separator etc. since this is the last section.

kierangillen: There will be an extra separator rendered at the bottom of the page.

You can see it doesn't need to get too crazy: https://github.com/artsy/emission/blob/master/src/lib/Scenes/Show/Screens/Detail.tsx

PR #1680, opened 2019-06-21
[relay] Upgrade to v5 and move to use single relay config

alloy:
Yay, entirely up to date with Relay!

And I get to use my new `relay-config` package 🎉  https://github.com/facebook/relay/pull/2746


alloy:
I tested this against latest staging, which has a re-applied change that broke https://github.com/artsy/metaphysics/pull/1787 and seems to be working well.

kierangillen:
☀️

PR #1685, opened 2019-06-24
PURCHASE-1178 - Adds context card

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1178

- Adds `ContextCard`
- Fixes bug in `PartnerCard` when no icon 
- Updates buttons to use `longestText` prop

## Examples

Screen Shot 2019-07-11 at 2 19 33 PM

Screen Shot 2019-07-11 at 2 10 55 PM

Screen Shot 2019-07-11 at 1 55 41 PM


alloy:
As [mentioned in Slack](https://artsy.slack.com/archives/C9YNS4X32/p1562765587438200?thread_ts=1562695609.412000&cid=C9YNS4X32), I think you should use the existing `ArtworkContextPartnerShow` type for now.

src/lib/Scenes/Artwork/Components/PartnerCard.tsx
```diff
@@ -86,7 +86,7 @@ export class PartnerCard extends React.Component<Props, State> {
       return null
     }
     const { isFollowedChanging } = this.state
-    const imageUrl = partner.profile ? partner.profile.icon.url : null
+    const imageUrl = partner.profile && partner.profile.icon ? partner.profile.icon.url : null
```
sweir27: You can also use the `get` util here, I believe!

src/lib/Scenes/Artwork/Components/ContextCard.tsx
```diff
@@ -0,0 +1,238 @@
...
+          id
+          name
+          href
+          formattedStartDateTime
+          cover_image {
+            url
+          }
+        }
+      }
+      shows(size: 1) {
```
sweir27: Is there a reason we're not spreading on the types for `show` and `fair`? (`... on ArtworkContextFair` and `... on ArtworkContextPartnerShow`)?

src/lib/Scenes/Artwork/Components/PartnerCard.tsx
kierangillen: Personally would rather not add overhead but personally preference.

ashfurrow: `get()` also doesn't provide any kind of type safety 😞 We'll get `strictNullChecks` turned on here eventually 😅

src/lib/Scenes/Artwork/Components/ContextCard.tsx
kierangillen: The reason we're not for Show is the context doesn't have a `ArtworkContextPartnerShow` type yet. Can add it for `Fair`.

kierangillen: Curious why we need to spread it in the context though if we're already querying for the fair directly on the Artwork.

src/lib/Scenes/Artwork/Components/PartnerCard.tsx
sweir27: Ah, I didn't mean importing directly from lodash.

We built a [typesafe version](https://github.com/artsy/emission/blob/master/src/lib/utils/get.ts) of this function in reaction, and it was imported here in April. 😄

src/lib/Scenes/Artwork/Components/ContextCard.tsx
sweir27: Gotcha on the `show`.

The way it's written right now you will _always_ be querying for `fair` on the artwork, but by spreading on the union type, you'll only fetch for the resource when it's part of the context that's returned.

src/lib/Scenes/Artwork/Components/PartnerCard.tsx
ashfurrow: Oh, super-cool!

src/lib/Scenes/Artwork/Components/ContextCard.tsx
kierangillen: Cool just updated!

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -236,11 +245,20 @@ export const ArtworkContainer = createFragmentContainer(Artwork, {
       manufacturer
       image_rights
 
+      # For ContextCard
+      context {
+        __typename
+      }
+      shows(size: 1) {
```
sweir27: It doesn't seem like you are using this anymore in here.

PR #1687, opened 2019-06-25
[PURCHASE-1128] Font update to fix RN text rendering

ashfurrow:
This is a follow-up from https://github.com/artsy/Artsy-UIFonts/pull/12. It now correctly loads and uses the new fonts with @ds300's modifications. I also changed the typography storybook entry to add separators between each element, making it clear where the bounds of the font are:

Simulator Screen Shot - Unit Test Device - 2019-06-25 at 13 22 08

Some of them still look strange, but those are the ones David didn't add RN-specific entries for, so I think it should be fine. https://github.com/artsy/Artsy-UIFonts/pull/12 documents how I fixed the underlying problem.

#skip_new_tests


src/lib/Components/Text/__stories__/Typography.story.tsx
```diff
@@ -15,29 +15,30 @@ storiesOf("App Style/Typography")
...
-        <Separator />
+        {Object.entries(fonts).map(value => {
+          const [name, Component] = value
+          return (
+            <React.Fragment key={name}>
+              <Component>{name}</Component>
+              <Separator />
+            </React.Fragment>
+          )
+        })}
```
ashfurrow: Credit for inspiring this idea (a dubious honour) goes to @jonallured. I still don't know if I _like_ this change, but I'm open to feedback 👍

jonallured: Woo! Converting things to data can be such a satisfying refactor! I think this is a little awkward because the extracted `fonts` local is an object rather than an array - what about something more like this:

```ts
const fonts = [
  { name: "AGaramondPro Bold", component: GaramondBold },
  { name: "AGaramondPro BoldItalic", component: GaramondBoldItalic },
  { name: "AGaramondPro Italic", component: GaramondItalic },
  { name: "AGaramondPro Regular", component: GaramondRegular },
  { name: "AGaramondPro Semibold", component: GaramondSemibold },
  { name: "Avant Garde Gothic ITC", component: AvantGardeRegular },
  { name: "Unica77LL Italic", component: Unita77LLItalic },
  { name: "Unica77LL Medium", component: Unita77LLMedium },
  { name: "Unica77LL MediumItalic", component: Unita77LLMediumItalic },
  { name: "Unica77LL Regular", component: Unita77LLRegular },
]
```

And then I'd use a few more locals to convert these to components:

```ts
const fontComponents = fonts.map((font, index) => {
  const Component = font.component
  return <Component key={index}>{font.name}</Component>
})

const separatedComponents = [<Separator />].concat(fontComponents.join(<Separator />))

return <Container>{separatedComponents}</Container>
```

Disclaimer: I didn't actually typecheck or lint this code so it's certainly possible I have syntax errors, etc - mostly just trying to convey an idea!! ❤️

ashfurrow: Oh yeah, that's way nicer. I also managed to get rid of a lot of boilerplate in this commit: https://github.com/artsy/emission/pull/1687/commits/0e93e52e7fa48952b75364f27eb732d3e5b4d844

ashfurrow: Here's what I've ended up with (JavaScript's `.join()` only takes strings, and outputs a string, sadly):

```tsx
const fonts = [
  { name: "AGaramondPro Bold", family: Fonts.GaramondBold },
  { name: "AGaramondPro BoldItalic", family: Fonts.GaramondBoldItalic },
  { name: "AGaramondPro Italic", family: Fonts.GaramondItalic },
  { name: "AGaramondPro Regular", family: Fonts.GaramondRegular },
  { name: "AGaramondPro Semibold", family: Fonts.GaramondSemibold },
  { name: "Avant Garde Gothic ITC", family: Fonts.AvantGardeRegular },
  { name: "Unica77LL Italic", family: Fonts.Unica77LLItalic },
  { name: "Unica77LL Medium", family: Fonts.Unica77LLMedium },
  { name: "Unica77LL MediumItalic", family: Fonts.Unica77LLMediumItalic },
  { name: "Unica77LL Regular", family: Fonts.Unica77LLRegular },
]
const fontComponents = fonts.map(({ name, family }) => {
  const Component = styled.Text`
    font-family: "${family}";
    font-size: 30px;
    margin-bottom: 10px;
  `
  return <Component key={name}>{name}</Component>
})
const separatedComponents = fontComponents.reduce(
  (memo, component, index) => [memo, <Separator key={index} />, component] as any // TypeScript doesn't know that JSX Elements can be arrays.
)
return (
  <Container>
    <Separator />
    {separatedComponents}
  </Container>
)
```

We end up with a bunch of nested arrays (based on `[memo, ...]`) but that should be fine for storybooks and for a limited number of fonts.

jonallured: LOL typescript! Come on computer you know what we mean!!

l2succes:
@ashfurrow how does that affect our use of `<Sans>` and `<Serif>` from palette throughout emission?

l2succes:
Sounds like we'll need to update this file with the new font names as well

```typescript
import { FontFamily } from "./index"

/**
 * A map of the font families and their settings
 */
export const fontFamily: FontFamily = {
  sans: {
    regular: "Unica77LL-Regular",
    italic: "Unica77LL-Italic",
    medium: "Unica77LL-Medium",
    mediumItalic: "Unica77LL-MediumItalic",
  },
  serif: {
    regular: "AGaramondPro-Regular",
    italic: "AGaramondPro-Italic",
    semibold: "AGaramondPro-Semibold",
  },
  display: {
    regular: "AvantGardeGothicITC",
  },
}
```

https://github.com/artsy/palette/blob/master/packages/palette/src/platform/fonts/fontFamily.ios.ts

ashfurrow:
Ah, good catch! I will follow-up with another PR.

PR #1692, opened 2019-06-27
View In Room, episode 1

saolsen:
vir

Stubs out the method to call to view an image in room. Next step is implementing that function in Eigen.


src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
```diff
@@ -77,12 +78,14 @@ export class ArtworkActions extends React.Component<ArtworkActionsProps> {
             </UtilButton>
           </TouchableWithoutFeedback>
           {Constants.AREnabled && (
-            <UtilButton pr={3}>
-              <Box mr={0.5}>
-                <EyeOpenedIcon />
-              </Box>
-              <Sans size="3">View in Room</Sans>
-            </UtilButton>
+            <TouchableWithoutFeedback onPress={() => ApiModule.presentAugmentedRealityVIR("url", 1, 2, "slug", "id")}>
```
sweir27: Should these be passing in an actual `url`, `slug`, and `id` vs. strings?

sweir27:
@saolsen I _think_ that peril comment is right re: incrementing the native version as well.

src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
ashfurrow: Yeah good call. Maybe we could even modify the ObjC alert to show the data that's passed into the `presentAugmentedRealityVIR` function.

ashfurrow: That would be doable with the [`NSString` method `stringWithFormat:`](https://developer.apple.com/documentation/foundation/nsstring/1497275-stringwithformat). Something like:

```objc
...
message: [NSString stringWithFormat:@"slug %@, width %f, height %f", slug, width, height]
...
```

Basically a C `printf`-style format string, with `%@` calling ObjC's equivalent of `toString()`.

PR #1693, opened 2019-06-28
PURCHASE-1211 - Removes InvertedButton

kierangillen:
- Deletes `ArtworkRail` and `FlatWhite` which aren't used
- Note: Some of the following images were taken before the `small` button had an additional `10px` added to each `x` side.
- https://artsyproduct.atlassian.net/browse/PURCHASE-1211

## FormButton

- Before

closebutton_before

- After 

formbutton_after

## CloseButton

- Before

closebutton_before

- After

closebutton_after

## ArtistHeader

- Before

artistheader_before

- After

artistheader_after

## GeneHeader

- Before

geneheader_before

- After

geneheader_after

## InvoicePreview

- Before

InvoicePreview_before

- After

InvoicePreview_after

## ShowItemRow

- Before

showitemrow_before

- After

showitemarrow_after

## ArtworkCarouselHeader

- Before

artwork_carousel_header_before

- After

artwork_carousel_header_after

## Event

- Before

event_before

- After

event_after

## Retry Button

retry

## FairBoothHeader

- Before

fairbooth_header_before

- After

fairbooth_header_after

## FairBoothPreview

- Before

fairbooth_preview_before

- After

fairbooth_preview_after

## ShowHeader

- Before

show_header_before

- After

show_header_after

#skip_new_tests


src/lib/Scenes/Show/Components/ShowHeader/ShowHeader.tsx
```diff
@@ -185,17 +178,17 @@ export class ShowHeader extends React.Component<Props, State> {
...
-              text={is_followed ? "Show saved" : "Save show"}
-              selected={is_followed}
-              onPress={this.handleFollowShow}
-              grayBorder={true}
-            />
-            <Spacer m={1} />
-          </ButtonWrapper>
+          <Spacer mb={2} />
+          <Button
+            width="100%"
```
ds300: I don't think percentage values work on RN. Is this being converted somehow to a flex value or something else?

src/lib/Components/Consignments/Screens/Confirmation.tsx
```diff
@@ -131,8 +131,8 @@ export default class Confirmation extends React.Component<Props, State> {
           <Subtitle>Please try again.</Subtitle>
         </View>
         <ButtonView>
-          <BorderedBlackButton text="TRY AGAIN" onPress={this.restart} />
-          <Button text="QUIT" onPress={this.exitModal} />
+          <FormButton text="TRY AGAIN" onPress={this.restart} />
+          <FormButton text="QUIT" onPress={this.exitModal} />
```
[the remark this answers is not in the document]
kierangillen: I wanted to bring this up actually with design, it's the only time we're using this `<BorderedBlackButton>`, this is a style we're no longer supporting for buttons and considering this is only displayed in an error state I took the liberty to remove the button style. But perhaps it should be replaced with another style.

src/lib/Scenes/Show/Components/ShowHeader/ShowHeader.tsx
kierangillen: Our new Palette `<Button>` supports width as a prop as long as `block` is also passed and I was told to pass it this string with percentage, but I'll take a look at using a plain number value instead!

src/lib/Components/Lists/ShowItemRow.tsx
```diff
@@ -162,18 +161,9 @@ export class ShowItemRow extends React.Component<Props, State> {
...
-                  text={show.is_followed ? "Saved" : "Save"}
-                  selected={show.is_followed}
-                  onPress={() => this.handleSave()}
-                  noBackground={true}
-                  hitSlop={{ top: 10, bottom: 10, left: 0, right: 0 }}
-                />
-              </Box>
-            </Flex>
+            <Button variant="noOutline" onPress={() => this.handleSave()} loading={this.state.isFollowedSaving}>
+              {show.is_followed ? "Saved" : "Save"}
```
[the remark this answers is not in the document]
kierangillen: We don't currently have a prop for button text color in the new `Button`, but maybe is should be added. I think we want to keep the styling on these buttons to a minimum but I'll bring this up with Sam.

src/lib/Scenes/Show/Components/ShowHeader/ShowHeader.tsx
ds300: I checked, looks like styled-components is changing percentage values to flex values.

src/lib/Components/Consignments/Screens/Confirmation.tsx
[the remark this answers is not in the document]
ashfurrow: Does Palette have a convention for primary+secondary buttons? Or destructive buttons? I'd say "QUIT" should be destructive, or secondary, in this case.

src/lib/Components/Home/ArtistRails/ArtistCard.tsx
```diff
@@ -72,11 +71,15 @@ export class ArtistCard extends React.Component<Props, State> {
 
   setFollowStatus(status: boolean): ArtistFollowHandlerResult {
     return new Promise(resolve => {
-      this.setState({
-        following: status,
-        processingChange: false,
-        followStatusChanged: resolve,
```
ashfurrow: I think we can also remove `followStatusChanged` from `state` altogether, right?

PR #1695, opened 2019-07-01
PURCHASE-1229, PURCHASE-1236, PURCHASE-1231, PURCHASE-1225: iOS Artwork Page small changes and bug fixes

lilyfromseattle:
__NOTE: Depends on metaphysics changes that are merged but not yet released__

# PURCHASE-1229: Artwork title should be in italics

__Before:__
Screen Shot 2019-07-01 at 12 34 12 PM

__After:__
Screen Shot 2019-07-01 at 12 35 18 PM

# PURCHASE-1236: Under the "Artwork Details" section, "Medium" should show the category 
We should be showing category under the “Medium” section in artwork details, not the artwork’s “medium”.

__Before:__
Screen Shot 2019-07-01 at 12 32 22 PM

__After:__
Screen Shot 2019-07-01 at 12 09 46 PM

# PURCHASE-1231: In the Artwork details section, if COA and Framed are null don't show field at all
Currently, under the Certificate of Authenticity and Framed sections, if the field is null (i.e. was never set), we should not show the field at all.

If the field is unchecked in CMS, we should hide the field.

If the field is checked either yes or no in CMS, we should show the field along with either "Included" or "Not included".

Screen Shot 2019-07-01 at 12 08 57 PM

# PURCHASE-1225: In the artwork details section, if the image rights field is an empty string, we should not show the field at all

Currently if Image rights does not exist, we still show Image rights. We should not show the field title at all in this case.

__NOTE:__ I also fixed a bug I found where the "show more details" link was displaying even if no details were hidden (the button did nothing)
#trivial


src/lib/Scenes/Artwork/Components/ArtworkDetails.tsx
```diff
@@ -17,17 +17,19 @@ export class ArtworkDetails extends React.Component<ArtworkDetailsProps, Artwork
...
       { title: "Signature", value: this.props.artwork.signature },
       {
         title: "Certificate of Authenticity",
-        value: this.props.artwork.certificateOfAuthenticity ? "Included" : "Not Included",
+        value: this.props.artwork.certificateOfAuthenticity
+          ? this.props.artwork.certificateOfAuthenticity.details
+          : null,
       },
-      { title: "Frame", value: this.props.artwork.framed ? "Included" : "Not Included" },
+      { title: "Frame", value: this.props.artwork.framed ? this.props.artwork.framed.details : null },
```
ashfurrow: This could be further shortened to the following:

```suggestion
      { title: "Frame", value: this.props.artwork.framed && this.props.artwork.framed.details },
      { title: "Frame", value: this.props.artwork.framed ? this.props.artwork.framed.details : null },
```

Which evaluates to the same thing.

sweir27:
@lilyfromseattle other than the merge conflict, this looks good!

The metaphysics changes were verified/deployed this morning. 👍

PR #1696, opened 2019-07-01
Refactored ArtistListItem to use palette EntityHeader component and added padding to artist cards

lilyfromseattle:
Although the ticket I pulled [PURCHASE-1237] was just for adding spacing between the artist cards, I though it made sense to refactor the artist cards to use the EntityHeader component from palette so that it is consistent with the other "card" on the artwork page (partner card).

I refactored ArtistListItem to use the EntityHeader component. I also added padding to the artist cards. I also removed a duplicate AboutArtist component that was getting added to the Artwork page twice.

__Before:__
Screen Shot 2019-07-01 at 5 36 27 PM

__After:__
Screen Shot 2019-07-01 at 5 35 49 PM

[PURCHASE-1237]: https://artsyproduct.atlassian.net/browse/PURCHASE-1237

#trivial


sweir27:
This looks good! (And is also an item that Sam brought up during our QA session, so good catch).

Will be interested in @kierangillen's review, and in particular whether or not this will have an effect on other places this component is used (I think City Guide?).

src/lib/Components/ArtistListItem.tsx
```diff
@@ -152,29 +120,31 @@ export class ArtistListItem extends React.Component<Props, State> {
...
-              variant={is_followed ? "secondaryOutline" : "primaryBlack"}
-              onPress={this.handleFollowArtist}
-              size="small"
-              loading={isFollowedChanging}
-            >
-              {is_followed ? "Following" : "Follow"}
-            </Button>
-          </Flex>
-        </TouchableWithoutFeedback>
+        <Box width="100%" py={1}>
```
kierangillen: We don't want to add margin/padding inside of a component which extends outside of the component's limits. I.e. this `Box` with padding should wrap the `ArtistListItem` outside of the component not live inside the component. Therefore we can control the margin/padding in different use cases/styles and keeps the padding and margin consistent without duplicating it in some places.

kierangillen: Other than this this looks great!

src/lib/Scenes/Artwork/Components/AboutArtist.tsx
```diff
@@ -27,7 +27,9 @@ export class AboutArtist extends React.Component<AboutArtistProps> {
             {hasSingleArtist ? "About the artist" : "About the artists"}
           </Sans>
           {artists.map(artist => (
-            <ArtistListItem key={artist.id} artist={artist} />
+            <Box width="100%" py={1} key={artist.id}>
```
sweir27: cc @kierangillen this should be good for another review.

I noticed that the other usages of the `ArtistListItem` added their own padding, so I only stuck the custom behavior here.

kierangillen: Hey would it be possible to add an image to the PR showing this component didn't visually change?

kierangillen: The `<ArtistListItem>` didn't have any internal padding previously so we shouldn't need to add a wrapper.

sweir27: Not sure I understand-- this wrapper is so that when there are multiple, the components have some padding between them.

sweir27: image

sweir27: Oh oh, you're referring to whether or not we need the `Box`-- let me check.

src/lib/Scenes/Artwork/Components/AboutArtist.tsx
```diff
@@ -26,9 +26,11 @@ export class AboutArtist extends React.Component<AboutArtistProps> {
           <Sans size="3t" weight="medium" mb={2}>
             {hasSingleArtist ? "About the artist" : "About the artists"}
           </Sans>
-          {artists.map(artist => (
-            <ArtistListItem key={artist.id} artist={artist} />
-          ))}
+          <Join separator={<Spacer my={1} />}>
```
sweir27: One last try-- I think this makes the most sense! When there are multiple artists, we need some spacing between them.

kierangillen: looks good!!

PR #1697, opened 2019-07-02
Use native image view in Carousel

ds300:
This PR changes the ImageCarousel component to use our `OpaqueImageView` component rather than the plain RN `Image` component. This has a couple of benefits:

- It lets the image cache be available for the View In Room feature.
- It shifts the onus on generating the correctly-sized image URL on to the `OpaqueImageView` component rather than having metaphysics do it, which involved adding extra query variables and other architectural weirdness.

This PR does two other bits of cleanup:

- **Fixes the carousel alignment**
   The carousel is full-bleed, but when padding was recently added to the artwork page the carousel wasn't treated differently than the rest of the page content and was therefore given an undesirable left offset.
- **Fixes a styled-components warning**
   The `Header` component was using responsive font sizes (presumably copy-pasted from web), which makes `styled-system` add a `@media(...)` clause into the component's css, which `styled-components` logs a warning about since RN doesn't support anything like media queries.


src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
```diff
@@ -50,9 +50,15 @@ export const ImageCarousel: React.FC<ImageCarouselProps> = ({ images }) => {
         decelerationRate="fast"
         onScroll={onScroll}
         renderItem={({ item, index }) => {
-          const { url } = item.thumbnail || item
           const { cumulativeScrollOffset, ...styles } = measurements[index]
-          return <ImageWithLoadingState source={{ uri: devCacheBust(url) }} style={styles} />
+          return (
+            <ImageWithLoadingState
+              imageURL={devCacheBust(item.url)}
```
ashfurrow: Only just now found out about this `devCacheBust` function, neat!

PR #1699, opened 2019-07-03
Changes method parameter names to make the units for width+height in inches

ashfurrow:
This was already the case, and the changed parameter names have no effect on _this_ code, though it'll make it more obvious. See conversation here: https://github.com/artsy/eigen/pull/2856#discussion_r299645729


sweir27:
Thank you!!

merge on green 😄

sweir27:
The Fossa integration appears to be taking a long time. Going to merge.

PR #1700, opened 2019-07-03
PURCHASE-1249 Inject safe area insets into artwork page

ds300:
Hellooo :wave:

While working on the fullscreen image carousel for the new artwork page I encountered a problem. I was creating a 'zoom' shared-element transition between the in-page artwork carousel image and the version of the same image in the fullscreen modal. In order to do that you need to `measure` the images both on the page and in the fullscreen modal and calculate the difference between their bounding boxes (both size and position). I discovered that the bounding box for the full screen version was 20px offset from the version on the page, and that immediately told me that this was some kind of status bar issue :)

So it seems like that RN `measure` bounding box positions are not relative to the screen, but relative to the parent UIView, and that RN modals are implemented as separate UIViews which occupy the whole screen. Meanwhile, before this PR, emission's example app was embedding the RN root view inside a different 'nav' UIView which was constrained to avoid the status bar.

So the solution I came up with is to let the artwork page occupy the whole screen and do it's own status-bar-avoiding in RN by injecting the `safeAreaInsets`.

The City Guide map view controller was already getting `safeAreaInsets` injected so I just moved that logic up to the `ARComponentViewController` and put it behind an opt-in flag.

I also needed to remove the status bar offset in the Example app navigation view, so that these changes are visible in emission and not just eigen. I toyed with adding it back in at a place where it wouldn't affect the RN views, but because the menu items are already taking safeAreaInsets into account it actually doesn't make any usability difference as far as I can tell. 🤷‍♂️

#skip_new_tests


Pod/Classes/ViewControllers/ARComponentViewController.m
```diff
@@ -125,4 +130,38 @@ - (void)setProperty:(id)value forKey:(NSString *)key
...
+    if (@available(iOS 11.0, *)) {
+        safeAreaInsets = self.view.safeAreaInsets;
+
+        // Once we receive the correct top inset value it gets resetted to 0 after the VC gets hidden
+        // So let's not reset it afterwards
+        if (self.view.safeAreaInsets.top > 0) {
+            self.safeAreaInsetsAreSet = YES;
+        }
+    } else {
+        self.safeAreaInsetsAreSet = YES;
```
ashfurrow: This `if` statement makes sense, but let's make sure to test both Emission and Eigen in pre-iOS 11 simulators and devices. I have some here we can use 👍

package.json
```diff
@@ -142,7 +142,7 @@
     "jest-styled-components": "7.0.0-2",
     "lint-staged": "6.1.1",
     "metro-react-native-babel-preset": "0.49.0",
-    "patch-package": "5.1.2",
+    "patch-package": "^6.1.2",
```
ashfurrow: 🎉

src/lib/Scenes/Map/GlobalMap.tsx
```diff
@@ -525,6 +526,9 @@ export class GlobalMap extends React.Component<Props, State> {
   }
 
   onRegionIsChanging = async () => {
+    if (!this.map) {
```
ashfurrow: Curious about this change – seems like there's a story behind it?

ds300: Ahh yes. I had meant to add a comment here about it after submitting the pull request, but forgot about it after standup 😬 

I was seeing `unhandled promise rejection`s pointing to this line when panning around in the map. It seemed that the map ref was `undefined` at some points during panning, but not all the time. As far as I could tell the map was behaving correctly despite these errors, so I wasn't sure it was worth digging into right now, but we should probably add a ticket to follow up on it since it could be a symptom of a larger issue. What do you think?

Pod/Classes/ViewControllers/ARComponentViewController.m
ds300: good thinking! I just checked iOS 10 and it was broken 👀 

Seems the `safeAreaInsets` prop is not being set on iOS 10 but I don't see an obvious reason why, and I have to sign off now. Any ideas? If not I'll debug tomorrow morning.

ashfurrow: Oh weird – I would expect `UIEdgeInsetsZero` to get passed through in iOS 10. My only idea would be that the `viewSafeAreaInsetsDidChange` function won't get called, so we're relying on `viewDidLoad` alone, which might have a different runtime behaviour than `viewSafeAreaInsetsDidChange`. I'll be online Friday too, if you want to debug together 👍

src/lib/Scenes/Map/GlobalMap.tsx
ashfurrow: Hmm, that's a good question. I think this change makes sense, even without the promise rejections you were seeing, but I'm not sure if it's worth digging in further.

@l2succes what do you think?

Pod/Classes/ViewControllers/ARComponentViewController.m
ds300: OK this is fixed, and I simplified the injection code significantly too.

As far as I could tell, `setProperty` wasn't triggering an update on iOS 10, so I moved the pre-ios-11 code to the init method.

src/lib/Scenes/Map/GlobalMap.tsx
ds300: Hmm I think this might need further investigation after all. Seeing things like this now: 
image which is related to the same 'map ref is undefined sometimes' error, and should probably not be happening.

I checked and it's happening on master for me, so it's unrelated to the changes in this PR. One thing I noticed that might be relevant is this:

image

Pod/Classes/ViewControllers/ARComponentViewController.m
```diff
@@ -7,6 +7,7 @@ @interface ARComponentViewController ()
 @property (nonatomic, strong, readonly) AREmission *emission;
 @property (nonatomic, strong, readonly) NSString *moduleName;
 @property (nonatomic, strong) NSDictionary *initialProperties;
+@property (nonatomic) bool safeAreaInsetsWereUpdated;
```
ashfurrow:
```suggestion
@property (nonatomic) BOOL safeAreaInsetsWereUpdated;
```

PR #1703, opened 2019-07-08
[PURCHASE-1224, PURCHASE-1230] Small adjustments to the new artwork page

lilyfromseattle:
# PURCHASE-1224: "View in room" button should be centered

__Before:__
Screen Shot 2019-07-08 at 4 38 16 PM

__After:__
Screen Shot 2019-07-08 at 4 40 58 PM

# PURCHASE-1230: "•" between artist name and follow should be round and align correctly with artist name

__Before:__
Screen Shot 2019-07-08 at 1 25 54 PM

__After:__
Screen Shot 2019-07-08 at 1 48 21 PM

#trivial

### DIFF
--- a/src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
+++ b/src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
@@ -74,7 +74,7 @@ export class ArtworkActions extends React.Component<ArtworkActionsProps> {
       <View>
         <Flex flexDirection="row">
           <TouchableWithoutFeedback onPress={() => this.handleArtworkSave()}>
-            <UtilButton pr={3} width="110px">
+            <UtilButton pr={3}>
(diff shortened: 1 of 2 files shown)


ds300:
😬 oooh just saw that the typography fixes were not propagated to emission yet so it's probably not that.

So two things:

- It's weird that you and I have different font rendering. Any ideas if there's something different on your end? Could try a clean rebuild anyway.
- I think that regardless we can just merge this and then when the typography fixes are properly applied we might need to tweak the follow button again. Not a big deal.

Can you move this branch to the artsy repo and make a new PR? Peril is complaining about the fork thing.

ds300:
Doing some more reviews and it turns out @kierangillen took care of the artist name alignment issue over here: https://github.com/artsy/emission/pull/1704/files#r301323777 and made a nice refactor too.

So let's cherry pick the commit for the view-in-room button and leave the other commit out when making the new non-fork PR :+1:

ds300:
closing in favour of #1707

PR #1705, opened 2019-07-08
[WIP] PURCHASE-1201: Adds remaining Artwork Contexts to the OtherWorks section of the Artwork View

sweir27:
**THIS IS WIP DUE TO:**
- Palette upgrade (https://github.com/artsy/palette/pull/534)
- Waiting for `ArtworkContextShow` to be available in metaphysics v2

This PR adds the support for the remaining contexts that an artwork can be in:
- ArtworkContextAuction
- ArtworkContextFair
- ArtworkContextShow

See [this notion doc](https://www.notion.so/artsy/Artwork-View-Other-Works-Grids-d4bdf91f0c604bff9bbc1eb48d0596ed) for functional specs (which grids should show and when) and [this zeplin file](zpl.io/VDAkX8r) for visual specs.

It's a big diff but I tried to be thoughtful with my commit history and tests, so feel free to follow that as a way to review. 

In lieu of a screenshot for every context (there are a bunch), here's what the links look like:

Screen Shot 2019-07-08 at 6 52 43 PM


sweir27:
The tests will fail until the palette bump, but this should be reviewable in the meantime!

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkContextAuction.tsx
```diff
@@ -0,0 +1,37 @@
...
+      return (
+        <ContextSectionsContainer>
+          <AuctionArtworkGrid artwork={props.artwork} />
+        </ContextSectionsContainer>
+      )
+    }
+  },
+  {
+    artwork: graphql`
+      fragment ArtworkContextAuction_artwork on Artwork @argumentDefinitions(excludeArtworkIDs: { type: "[String!]" }) {
```
sweir27: In the web implementation, we conditionally fetch for these depending on the value of `isClosed`, but that's easier because these are all rendered below the fold and fetched client-side.

We can do that here if we want, but I also want to verify with Sam whether or not it's desired-- we may be able to remove some complexity here.

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkContextFair.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <FairArtworkGrid artwork={props.artwork} />
+        <ShowArtworkGrid artwork={props.artwork} />
+        <ArtistArtworkGrid artwork={props.artwork} />
+        <RelatedArtworkGrid artwork={props.artwork} />
+      </ContextSectionsContainer>
+    )
+  },
+  {
+    artwork: graphql`
+      fragment ArtworkContextFair_artwork on Artwork {
```
ashkan18: do we need `@argumentDefinitions(excludeArtworkIDs: { type: "[String!]" })` here to exclude current artwork?

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkContextShow.tsx
```diff
@@ -0,0 +1,31 @@
...
+        <ShowArtworkGrid artwork={props.artwork} />
+        <ArtistArtworkGrid artwork={props.artwork} />
+        <PartnerArtworkGrid artwork={props.artwork} />
+        <RelatedArtworkGrid artwork={props.artwork} />
+      </ContextSectionsContainer>
+    )
+  },
+  {
+    artwork: graphql`
+      fragment ArtworkContextShow_artwork on Artwork {
```
ashkan18: same question here about `@argumentDefinitions(excludeArtworkIDs: { type: "[String!]" })`

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkGrids/AuctionArtworkGrid.tsx
```diff
@@ -0,0 +1,57 @@
...
+  artwork: AuctionArtworkGrid_artwork
+}
+
+export class AuctionArtworkGrid extends React.Component<AuctionArtworkGridProps> {
+  render() {
+    const {
+      artwork: { sale },
+    } = this.props
+
+    if (!sale) {
```
ashkan18: technically we should never get into this `if` right?

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkContextFair.tsx
sweir27: Ahhh good catch.

src/lib/Scenes/Artwork/Components/OtherWorks/ArtworkContexts/ArtworkGrids/AuctionArtworkGrid.tsx
sweir27: It would be weird but we could, especially depending on where this fragment container is consumed.

sweir27:
The palette upgrade is necessary so that we get the icon!

src/lib/Scenes/Artwork/Components/OtherWorks/ContextGridCTA.tsx
```diff
@@ -0,0 +1,26 @@
...
+  render() {
+    const { href, label } = this.props
+
+    return (
+      <TouchableWithoutFeedback onPress={() => href && SwitchBoard.presentNavigationViewController(this, href)}>
+        <Text>
+          <Sans size="2" textAlign="left" weight="medium">
+            {label}
+          </Sans>
+          <ArrowRightIcon fill="black30" height="8px" />
```
ds300: :thinking: should we either make href required or make this disappear when `href` is not given?

sweir27: Ooh good catch.

ds300: Also, `8px`/`black30` seemed small/faint to me so I tried to find this on zeplin but couldn't find it on the responsive artwork specs or in the purchase ios artwork specs. Do you have a link?

ds300: the icon in the spec is slightly different but I think making the text size `"3"`, leaving the icon at its default height, and adding a `ml={1}` on the icon achieves about the right weight balance.

```tsx
          <Sans size="3" textAlign="left" weight="medium">
            {label}
          </Sans>
          <ArrowRightIcon fill="black30" ml={1} />
```

image

Looks much better in context! :)

sweir27: This does look good! I was following the specs that @samchieng gave me in slack, but this does look a bit clearer.

Will see what she things! 🙏

samchieng: ooo yes this is much better, thanks @ds300 and @sweir27!

PR #1706, opened 2019-07-09
Ensure `edition_of` is never an empty string

zephraph:
While testing something unrelated I ran across an invariant issue where `edition_of` returned an empty string.

This is related to a lot of other fixes that @kierangillen did in https://github.com/artsy/emission/pull/1559

See https://github.com/facebook/react-native/issues/20764 for more context.

#trivial


sweir27:
Thanks!

PR #1710, opened 2019-07-10
Revert "Remove gravityID in favor of internalID and slug"

zephraph:
Reverts artsy/emission#1691

See incident-46


ashfurrow:
CI failed on account of the Danger error above. I'm merging as-is rather than wait to rebuild CI.

PR #1720, opened 2019-07-12
[GROW-1392] Fixes crash and map UX for city guides

ashfurrow:
This code was [recently refactored](https://github.com/artsy/emission/commit/a73a34de3abfc1e6df2176a1b6a903726b7fed83) by @alloy  but I don't quite understand how this was _ever_ working. The ref we were storing was always `undefined`. According to [the docs](https://reactjs.org/docs/refs-and-the-dom.html), the call to `.root` is unnecessary. @ds300 ran into a similar problem here: https://github.com/artsy/emission/pull/1700/files#r300014570

In any case, this is working now! h/t to @ansor4 for debugging this with me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Removes InvertedButton - kierangillen
 - Adds ContextCard - kierangillen
 - Fixes BottomAlignedButton keyboard clipping - kierangillen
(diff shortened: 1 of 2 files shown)


l2succes:
Awesome! Thanks @ashfurrow.

merge on green

PR #1721, opened 2019-07-12
[PURCHASE-1259, PURCHASE-1266, PURCHASE-1269, PURCHASE-1273] iOS artwork page small text and styling fixes

lilyfromseattle:
# [PURCHASE-1259] Visual updates to "Consign with Artsy" text

"Consign with Artsy" copy should use Unica instead of Garamond
Spacing should be 20px between Partner Name and "Consign with Artsy" text

Screen Shot 2019-07-12 at 1 39 07 PM

# [PURCHASE-1266] Small text fixes: "Artwork details" label should be sentence case and change "About this work" to "About the work"

"Artwork Details" --> "Artwork details"
“About this work” → “About the work” (to be consistent with web)

Screen Shot 2019-07-12 at 1 38 34 PM

# [PURCHASE-1269] Add 10px spacing above the "Artwork classifications" link and a period at the end of the sentence. The period should not be clickable.

Screen Shot 2019-07-12 at 1 37 47 PM

# [PURCHASE-1273] Add 3px spacing after "From Artsy Specialist" label

[PURCHASE-1259]: https://artsyproduct.atlassian.net/browse/PURCHASE-1259
[PURCHASE-1266]: https://artsyproduct.atlassian.net/browse/PURCHASE-1266
[PURCHASE-1269]: https://artsyproduct.atlassian.net/browse/PURCHASE-1269
[PURCHASE-1273]: https://artsyproduct.atlassian.net/browse/PURCHASE-1273

#trivial


src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
```diff
@@ -124,11 +124,15 @@ export class ArtworkTombstone extends React.Component<ArtworkTombstoneProps, Art
           </Serif>
         )}
         {artwork.attribution_class && (
-          <Text style={{ textDecorationLine: "underline" }} onPress={() => this.handleTap("/artwork-classifications")}>
-            <Serif color="black60" size="3t">
+          <Serif color="black60" size="3t" mt={1}>
+            <Text
+              style={{ textDecorationLine: "underline" }}
+              onPress={() => this.handleTap("/artwork-classifications")}
```
kierangillen: Maybe we should add this `onPress` on a `TouchableWithoutFeedback` element instead of binding it to a `Text` element.

src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
```diff
@@ -16,20 +16,25 @@ export class ArtworkExtraLinks extends React.Component<ArtworkExtraLinksProps> {
...
   }
 
   render() {
     const { artwork } = this.props
     const consignableArtistsCount = artwork.artists.filter(artist => artist.is_consignable).length
 
-    return <Box>{!!consignableArtistsCount && this.renderConsignmentsLine(consignableArtistsCount)}</Box>
+    return (
+      <Box>
+        <Spacer mb={2} />
```
kierangillen: Would it be possible to add this spacer to the parent component above this component and not add margin directly to this component?

lilyfromseattle: We can, but if we want the space to only render when there are links we'd have to add some extra fields and logic to the parent because not all artwork pages have links here.

src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
lilyfromseattle: I used text so I could add the underline. Are you suggesting I wrap the text element in a TouchableWithoutFeedback?

kierangillen: Yea exactly! Just for the `onPress`

src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
kierangillen: Cool, I'll leave it up to you, generally I think we want to keep from adding margin to components in this way.

PR #1725, opened 2019-07-12
Remove gravity ID in favor of slug/internalID

zephraph:
This was reverted in #1710 and was originally #1691. Basically I restored the branch and rebased on master to get back to this point. 

This was reverted due to the app breaking in incident-46. Ultimately it was determined that the persisted query map was broken which caused the incident. We'll need to take care in this PR (and the paired MP PR https://github.com/artsy/metaphysics/pull/1818) that we don't repeat that mistake. 

#skip_new_tests


zephraph:
I still need to update the query map so let's not merge yet, heh.

PR #1726, opened 2019-07-12
WIP: AUCT-472: Update copy on registration pending view

yuki24:
finishes https://artsyproduct.atlassian.net/browse/AUCT-472

This is a simple change, but there's a question of whether or not the icon should be removed: https://artsyproduct.atlassian.net/browse/AUCT-472, so let's wait until we know the answer.

## Screenshot

Screen Shot 2019-07-12 at 5 35 37 PM


yuki24:
It seems like the FOSSA scan is stuck while the actual status is green. I'll just go ahead and merge this PR.

I've also added a bit more test coverage to make sure the icon is hidden only when the status is pending.

PR #1730, opened 2019-07-15
Update copy on credit card registration screen

yuki24:
finishes https://artsyproduct.atlassian.net/browse/AUCT-470

making a new PR that kicks off a build and Peril will let me know what I should do...

### Registration screen

Screen Shot 2019-07-15 at 5 33 01 PM

### Credit card form

Screen Shot 2019-07-16 at 11 08 05 AM

PR #1732, opened 2019-07-16
Add image thumbnail to the confirm bid screen 

yuki24:
finishes https://artsyproduct.atlassian.net/browse/AUCT-473

 * Adds `image { url(...) }` to the GraphQL query to retrieve a URL to the `small` image of the artwork
 * Replaces the `<Containers>` with a `<ScrollView>` so smaller screens can host all these elements without getting them squashed
 * Updates tests, snapshots and storybooks to reflect the design change
 * The `“A valid credit card is required for bidding.”` message has already been added as part of https://github.com/artsy/emission/pull/1730 so we get that for free everywhere else

### iPhone XS

Screen Shot 2019-07-16 at 3 51 46 PM

### iPhone SE

Screen Shot 2019-07-16 at 3 53 06 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,11 @@
 
 ### Master
 
+- Display the image of an artwork on the confirm bid screen - yuki24
+
 ### 1.12.6
 
 - Added partner type to partner card - lilyfromseattle
-
 - Removes InvertedButton - kierangillen
 - Adds ContextCard - kierangillen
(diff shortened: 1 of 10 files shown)


yuki24:
> If so, what's the motivation for not implementing such a test.

Passing a prop to a component is very basic even though the source is GraphQL and there is no conditional logic here. Typos will be caught by `yarn type-check` which automatically gets triggered on every `git push`, so it should be good.

yuki24:
Merge on green

PR #1734, opened 2019-07-17
PURCHASE-1201: Adds support for all "Other Works" grids on the Artwork view

sweir27:
Builds off of https://github.com/artsy/emission/pull/1705 (to be closed) and https://github.com/artsy/metaphysics/pull/1827

The logic for when to render a "context grid" on an artwork view now lives in Metaphysics!

This updates the artwork view to pull `contextGrids` directly and display the data that's returned.

Screen Shot 2019-07-17 at 12 38 02 PM

Screen Shot 2019-07-17 at 12 38 14 PM

#skip_new_tests


src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -252,7 +222,7 @@ export const ArtworkContainer = createFragmentContainer(Artwork, {
 
       ...PartnerCard_artwork
       ...AboutWork_artwork
-      ...OtherWorks_artwork
+      ...OtherWorks_artwork @relay(mask: false)
```
sweir27: cc @ds300 This seems to be the root of the error I mentioned in slack.

src/lib/Scenes/Artwork/Components/OtherWorks/index.tsx
```diff
@@ -1,37 +1,64 @@
...
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
-import { ArtworkContextArtistFragmentContainer as ArtworkContextArtist } from "./ArtworkContexts/ArtworkContextArtist"
+import { ContextGridCTA } from "./ContextGridCTA"
+import { Header } from "./Header"
 
-export const OtherWorksFragmentContainer = createFragmentContainer<{
-  artwork: OtherWorks_artwork
-}>(
+export const populatedGrids = grids => {
```
sweir27: We get back an array of grids from metaphysics, but it's possible that metaphysics will return a grid with 0 artworks (otherwise we'd have to make a bunch of extra fetches from within MP).

This helper method just returns the grids that have artworks. We reuse it in `Artwork.tsx` when deciding whether or not to push on the `OtherWorks` component at all.

src/lib/Scenes/Artwork/Artwork.tsx
ds300: we resolved this offline by adding both masked and unmasked fragment spreads for OtherWorks!

src/lib/Scenes/Artwork/Components/OtherWorks/index.tsx
ds300: We could type this `grids` parameter

```suggestion
export const populatedGrids = (grids: OtherWorks_artwork['contextGrids']) => {
```

src/lib/Scenes/Artwork/Components/OtherWorks/index.tsx
```diff
@@ -1,37 +1,64 @@
...
+  if (grids && grids.length > 0) {
+    return filter(grids, grid => {
+      const artworkNodes = grid.artworks && grid.artworks.edges && grid.artworks.edges.map(({ node }) => node)
```
ds300: is there a reason to map over the edges here? Could maybe just do something like

    return grid.artworks && grid.artworks.edges && grid.artworks.edges.length > 0

sweir27: Ah! Good catch.

PR #1735, opened 2019-07-17
PURCHASE-1144 - Artwork Analytics

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1144

- Adds 4 Artworks to Emission menu for testing purposes
- Updates `react-tracking@7.0.1` 

- https://docs.google.com/spreadsheets/d/1P6qUe7yFPzly5JYicAa6eQ3BAy9s7Bx1_6k-Q5hB-Pw/edit#gid=868762451

- https://docs.google.com/spreadsheets/d/1P6qUe7yFPzly5JYicAa6eQ3BAy9s7Bx1_6k-Q5hB-Pw/edit#gid=0

#skip_new_tests


Example/Emission/ARRootViewController.m
```diff
@@ -234,14 +237,41 @@ - (ARCellData *)jumpToUserDocs
   }];
 }
 
-- (ARCellData *)jumpToArtwork
+- (ARCellData *)jumpToArtwork1
 {
-  return [self tappableCellDataWithTitle:@"Artwork" selection:^{
+  return [self tappableCellDataWithTitle:@"Artwork1" selection:^{
```
sweir27: Can we make these names meaningful? (i.e. `Non Commercial Artwork`, `Inquiry Artwork`, `Artwork with multiple images`, etc. Feel free to choose what is most relevant or just describe the ones you've already chosen!

src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -23,6 +24,16 @@ interface Props {
...
 
+@screenTrack<Props>(props => ({
+  context_screen: Schema.PageNames.ArtworkPage,
+  context_screen_owner_type: Schema.OwnerEntityTypes.Artwork,
+  context_screen_owner_slug: props.artwork.slug,
+  context_screen_owner_id: props.artwork.internalID,
+  availability: props.artwork.availability,
+  acquireable: props.artwork.is_acquireable,
+  offerable: props.artwork.is_offerable,
+  biddable: props.artwork.is_biddable,
```
sweir27: Do we want `inquireable` here too?

src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
```diff
@@ -1,18 +1,27 @@
...
   handleTap(href: string) {
     SwitchBoard.presentNavigationViewController(this, href)
   }
 
+  @track(() => {
+    return {
+      action_name: Schema.ActionNames.ConsignWithArtsy,
+      action_type: Schema.ActionTypes.Tap,
+      context_module: Schema.ContextModules.ArtworkExtraLinks,
+    } as any
```
sweir27: Do these `as any`s speak to the need for a type update somewhere?

Example/Emission/ARRootViewController.m
kierangillen: We could for sure, but I was thinking to make them generic so that people can interchange a variety of ID's and not have to worry about renaming. I still think we'll be changing them in and out, I just hope not as often now. What do you think?

src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
kierangillen: May want to pair on the typing here. I spent some time poking around with the typing here with @l2succes a while ago and then just recently again and it seemed complicated.

Example/Emission/ARRootViewController.m
sweir27: I think it'll still be more helpful for our future selves to have some guidance as to what these are, even if it's rough (i.e. if I want to look at a work with multiple images, I don't want to have to remember that that's `Artwork 2` vs. `Artwork 1`).

Maybe we could keep one generic and then make the others have slightly more specific titles? I'm not too worried about the drift.

src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
sweir27: If it's very complicated I will defer to @ds300 for some 👀!

src/lib/Scenes/Artwork/Artwork.tsx
kierangillen: it's not in the schema but ill add it.

PR #1738, opened 2019-07-22
Remove .only and fix tests in ConfirmBid-tests.tsx

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-500

See [my comment below](https://github.com/artsy/emission/pull/1738#issuecomment-513883755) for more details. #trivial

### DIFF
--- a/src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
+++ b/src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
@@ -109,7 +109,7 @@ it("displays the artwork title correctly without date", () => {
   expect(serifChildren(component)).not.toContain(`${saleArtwork.artwork.title},`)
 })
 
-describe.only("checkbox and payment info display", () => {
+describe("checkbox and payment info display", () => {
   it("shows no checkbox or payment info if the user is registered", () => {
     const component = renderer.create(
       <BiddingThemeProvider>
--- a/src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
+++ b/src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
@@ -5,15 +5,13 @@ import { NativeModules, Text, TouchableWithoutFeedback } from "react-native"
 import "react-native"
 import * as renderer from "react-test-renderer"
 
-import Spinner from "../../../Spinner"
-
 import { LinkText } from "../../../Text/LinkText"
 import { BidInfoRow } from "../../Components/BidInfoRow"
 import { Checkbox } from "../../Components/Checkbox"
 
 import { BidResultScreen } from "../BidResult"
 import { BillingAddress } from "../BillingAddress"
-import { ConfirmBid } from "../ConfirmBid"
+import { ConfirmBid, ConfirmBidProps } from "../ConfirmBid"
 import { CreditCardForm } from "../CreditCardForm"
 
 jest.mock("../../../../metaphysics", () => ({ metaphysics: jest.fn() }))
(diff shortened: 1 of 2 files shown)


dleve123:
@yuki24 Just marked this PR as a WIP (I believe that's what you are conveying in the PR description). Super excited about this cleanup!

yuki24:
The danger comment has been removed but here is the original failure on Travis: https://travis-ci.org/artsy/emission/jobs/562192561

I've just pushed the fixes and now this is ready to ready. It seems like the entire spec is very out-dated so I had to change almost every single test. Here are the two main changes:

### 1. Replacing `setState`s

```ts
component.root.instance.setState({ conditionsOfSaleChecked: true })
```

was failing since now the top-level component is `<BiddingThemeProvider>` that does not have the state `conditionsOfSaleChecked`. Fixed by replacing them with:

```ts
component.root.findByType(Checkbox).props.onPress()
```

So we don't have to know about the implementation details of the TOS checkbox.

### 2. Remove the `.instance` from `findByType(...).instance.props`

`instance.props` was returning `undefined` and any call chained with `props` was failing. It seems like we no longer have to call the `.instance` function to get props and simply removing them fixed some of the tests.

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -168,12 +174,13 @@ describe("when pressing bid button", () => {
...
-    component.root.instance.setState({ conditionsOfSaleChecked: true })
+    component.root.findByType(Checkbox).props.onPress()
     relay.commitMutation = jest.fn()
+    const placeBidButton = findPlaceBidButton(component)
 
-    component.root.findByType(Button).instance.props.onPress()
+    placeBidButton.props.onPress()
 
-    expect(component.root.findAllByType(Spinner).length).toEqual(1)
+    expect(placeBidButton.props.loading).toEqual(true)
```
yuki24: We probably do not have to check the presence of the `<Spinner>` component here and can rely on the `loading` prop instead.

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -598,22 +640,16 @@ describe("ConfirmBid for unqualified user", () => {
...
-        },
-      })
+    expect(stripe.createTokenWithCard.mock.calls.length).toEqual(1)
+
+    const modal = component.root.findByType(Modal)
+
+    expect(modal.props.detailText).toEqual(
+      "There was a problem processing your information. Check your payment details and try again."
     )
+    expect(modal.props.visible).toEqual(true)
```
yuki24: It seems like now we have a modal within the `<ConfirmBid>` component to show errors, so updating the old test to reflect the actual change.

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -769,13 +782,19 @@ describe("ConfirmBid for unqualified user", () => {
...
+      expect(relay.commitMutation).toHaveBeenCalledWith(
+        expect.any(Object),
+        expect.objectContaining({
+          variables: {
+            input: {
+              phone: "111 222 4444",
+            },
+          },
+        })
+      )
```
yuki24: The `updatePhoneNumber()` call also seems like a new mutation, so adding it to the happy path test.

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -810,9 +829,10 @@ const serifChildren = comp =>
     .map(c => (c.props.children.join ? c.props.children.join("") : c.props.children))
     .join(" ")
 
-const saleArtwork = {
+const saleArtwork: ConfirmBid_sale_artwork = {
```
yuki24: Here I'm adding a lot of type annotations to make it safer.

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -821,10 +841,13 @@ const saleArtwork = {
...
   sale: {
-    id: "best-art-sale-in-town",
-    endsAt: "2018-05-10T20:22:42+00:00",
+    slug: "best-art-sale-in-town",
+    live_start_at: "2018-05-09T20:22:42+00:00",
+    end_at: "2018-05-10T20:22:42+00:00",
   },
   lot_label: "538",
+  " $fragmentRefs": null, // needs this to keep TS happy
+  " $refType": null, // needs this to keep TS happy
```
yuki24: I wasn't sure how I could keep TS happy without specifying these two types, but I think there might be a good way to do so. cc @alloy

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -810,9 +829,10 @@ const serifChildren = comp =>
     .map(c => (c.props.children.join ? c.props.children.join("") : c.props.children))
     .join(" ")
 
-const saleArtwork = {
+const saleArtwork: ConfirmBid_sale_artwork = {
+  internalID: "internal-id",
   artwork: {
-    id: "meteor shower",
+    slug: "meteor-shower",
```
yuki24: `id` vs `slug` continues... cc @zephraph

erikdstock:
merge on green

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
alloy: Specifying null is fine 👍

dblandin:
I found a related rule in the Jest ESLint plugin if we want to consider running that via Circle/Danger: https://github.com/jest-community/eslint-plugin-jest/blob/098219b4614192b365e51b0de68deac9caae1d68/docs/rules/no-focused-tests.md

PR #1742, opened 2019-07-22
Fixes an issue where placing a bid hangs

yuki24:
fixes https://artsyproduct.atlassian.net/browse/AUCT-499
blocked by https://github.com/artsy/emission/pull/1741

eThis is the actual fix for the bug. Basically the issue is that `result.position.id` was not replaced even after we switched to MP v2. As a result of this, `undefined` is passed on to the polling query which causes the server to return an error. This PR replaces all the occurances of `id` in the `<ConfirmBid>` component.

I have another PR that adds more type checks to mitigate the same type of risks: https://github.com/artsy/emission/compare/master...add-more-type-checks

### DIFF
--- a/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
+++ b/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
@@ -61,7 +61,7 @@ interface ConfirmBidState {
 
 const MAX_POLL_ATTEMPTS = 20
 
-const queryForBidPosition = (bidderPositionID: string) => {
+const queryForBidderPosition = (bidderPositionID: string) => {
   return metaphysics({
     query: `
       query ConfirmBidBidderPositionQuery($bidderPositionID: String!) {
@@ -259,7 +259,7 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
   createBidderPosition() {
     commitMutation<ConfirmBidCreateBidderPositionMutation>(this.props.relay.environment, {
       onCompleted: (results, errors) =>
-        isEmpty(errors) ? this.verifyBidPosition(results) : this.presentErrorResult(errors),
+        isEmpty(errors) ? this.verifyBidderPosition(results) : this.presentErrorResult(errors),
       onError: this.presentErrorResult.bind(this),
       mutation: graphql`
         mutation ConfirmBidCreateBidderPositionMutation($input: BidderPositionInput!) {
@@ -290,7 +290,7 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
     })
   }
 
-  verifyBidPosition(results) {
+  verifyBidderPosition(results) {
     const { result } = results.createBidderPosition
 
     if (result.status === "SUCCESS") {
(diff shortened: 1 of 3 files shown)


src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -64,14 +64,14 @@ const MAX_POLL_ATTEMPTS = 20
 const queryForBidPosition = (bidderPositionID: string) => {
```
dleve123: while we're here, thoughts on renaming to queryforBid**der**Position as a `BidderPosition` is a thing, but a `BidPosition` isn't a thing?

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -64,14 +64,14 @@ const MAX_POLL_ATTEMPTS = 20
...
   return metaphysics({
     query: `
-      {
+      query ConfirmBidBidderPositionQuery($bidderPositionID: String!) {
```
dleve123: out of curiosity, can you provide some more context for this change? Why name the query?

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -803,7 +811,7 @@ const mockRequestResponses = {
           message_header: "Success",
           message_description_md: "",
           position: {
-            internalID: "some-bidder-position-id",
+            internalID: "bidder-position-id-from-mutation",
```
dleve123: what's the level of effort requested to refactor the `pollingForBid` fixture to use relay like this fixture? Could we resolve these TODOs in this PR? Seems like that would increase the robustness of these tests for any future refactorings

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
yuki24: This used to be a dynamic query due to the presence of the string interpolation, which halts static analysis done by GraphQL. I've changed it to use `variables` to inject `bidderPositionID`. The query name could be any name, but I followed Relayjs' convention of `ComponentNameQueryName(Query|Mutation)`.

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
yuki24: The expectations for polling was simply missing and this basically backfills test coverage. The first exception specifically tests that the `internalID` from the mutation results is properly passed on to the first polling request:

https://github.com/artsy/emission/blob/ec0e94d6403122496c4ac3eb226378d16588377b/src/lib/Components/Bidding/Screens/ConfirmBid.tsx#L297

And the second one specifically tests the handling of the one from the polling results:

https://github.com/artsy/emission/blob/ec0e94d6403122496c4ac3eb226378d16588377b/src/lib/Components/Bidding/Screens/ConfirmBid.tsx#L317

While in production both calls would return the exact same id, I've used two different ids in test to make sure each one of the calls works as expected.

dleve123: This lead to a IRL conversation about the feasibility about generating types for these fixtures. See this comment: https://github.com/artsy/emission/pull/1744/files#r306406212

TL;DR is that we aren't going to worry about this type enhancement in this PR.

PR #1743, opened 2019-07-22
Display an error message when polling fails

yuki24:
finishes https://artsyproduct.atlassian.net/browse/AUCT-501
blocked by https://github.com/artsy/emission/pull/1742

 * [x] self-QA

## Successful case

ezgif com-gif-maker

## Failing case

AUCT-501

tested with the diff of:

```diff
diff --git a/src/lib/Components/Bidding/Screens/ConfirmBid.tsx b/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
index c001482c2..7222f6a07 100644
--- a/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
+++ b/src/lib/Components/Bidding/Screens/ConfirmBid.tsx
@@ -294,7 +294,7 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
     const { result } = results.createBidderPosition

     if (result.status === "SUCCESS") {
-      this.bidPlacedSuccessfully(result.position.internalID)
+      this.bidPlacedSuccessfully(result.position.id)
     } else {
       this.presentBidResult(result)
     }
@@ -317,7 +317,7 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
       // initiating new request here (vs setInterval) to make sure we wait for the previous call to return before making a new one
       setTimeout(
         () =>
-          queryForBidderPosition(bidder_position.position.internalID)
+          queryForBidderPosition(bidder_position.position.id)
             .then(this.checkBidderPosition.bind(this))
             .catch(error => this.presentErrorResult(error)),
         1000
```

PR #1744, opened 2019-07-22
Add more type checks

yuki24:
blocked by https://github.com/artsy/emission/pull/1743

The danger message is a false alarm as this PR does not include any behavioral change.

#trivial #skip_new_tests

### DIFF
--- a/src/__generated__/BidderPositionQuery.graphql.ts
+++ b/src/__generated__/BidderPositionQuery.graphql.ts
@@ -1,0 +1,235 @@
+/* tslint:disable */
+
+import { ConcreteRequest } from "relay-runtime";
+export type BidderPositionQueryVariables = {
+    readonly bidderPositionID: string;
+};
+export type BidderPositionQueryResponse = {
+    readonly me: {
+        readonly bidder_position: {
+            readonly status: string;
+            readonly message_header: string | null;
+            readonly message_description_md: string | null;
+            readonly position: {
+                readonly internalID: string;
+                readonly suggested_next_bid: {
+                    readonly cents: number | null;
+                    readonly display: string | null;
+                } | null;
+            } | null;
+        } | null;
+    } | null;
+};
+export type BidderPositionQuery = {
+    readonly response: BidderPositionQueryResponse;
+    readonly variables: BidderPositionQueryVariables;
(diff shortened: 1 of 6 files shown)


src/lib/Components/Bidding/Screens/ConfirmBid.tsx
```diff
@@ -62,16 +68,18 @@ interface ConfirmBidState {
 const MAX_POLL_ATTEMPTS = 20
 
 const queryForBidPosition = (bidderPositionID: string) => {
-  return metaphysics({
+  // TODO: Auto-generate types specifically for this query rather than using an
```
dleve123: @yuki24 and I just synced about this IRL. We can do something similar to what we do in Prediction so that we can get relay's nice type generation without needing to actually couple this query to a GraphQL component.

Relevant prediction source that we looked at: https://github.com/artsy/prediction/blob/master/src/server/externals/metaphysics/GetSaleMeta.ts

yuki24: I've extracted the `BidderPositionQuery` out of the `<ConfirmBid>` component to generate the types for it: https://github.com/artsy/emission/pull/1744/files#diff-17aaff619e4681074c7b728228f47ae4

alloy: I think we could take this one step further even. Relay exposes a `fetchQuery` function which means we don’t need the fake `GraphQL` function, we can retire the `metaphysics` function (I think); and potentially get some further benefits like cache invalidation in a single place, access to middlewares, etc.

But that’s probably all more for a follow-up PR.

alloy:
Great stuff @yuki24! 🙌

src/lib/Components/Bidding/Screens/ConfirmBid.tsx
yuki24: Very nice, it was a snap to switch to `fetchQuery`: https://github.com/artsy/emission/pull/1744/commits/097e3ba836a3c2a58a0122a45b6d5e1f7f5fd471

alloy:
Sweeeeet :100:

PR #1752, opened 2019-07-24
PURCHASE-1250 Fullscreen carousel

ds300:
# ImageCarousel

Hello 👋 

This PR introduces the full-screen image carousel. It's a bit of a beast, so make sure you [read the architectural description of the code changes](https://github.com/artsy/emission/pull/1752/files#diff-f00693397990492537087b3b9bc8730b) before diving in for a review. 

full-screen-carousel-final

#skip_new_tests


src/lib/Components/OpaqueImageView/OpaqueImageView.tsx
```diff
@@ -11,14 +11,14 @@ import {
...
-
-const GeminiHost = "d7hftxdivxxvm.cloudfront.net"
-const ImageQuality = 80
+import { createGeminiUrl } from "./createGeminiUrl";
 
 interface Props {
   /** The URL from where to fetch the image. */
   imageURL?: string
 
+  disableGemini?: boolean
```
sweir27: What is this `disableGemini` for?

src/lib/Components/OpaqueImageView/OpaqueImageView.tsx
```diff
@@ -79,19 +84,22 @@ export default class OpaqueImageView extends React.Component<Props, State> {
...
-      const type = this.state.aspectRatio ? "fit" : "fill"
-      const width = String(this.state.width)
-      const height = String(this.state.height)
-      // tslint:disable-next-line:max-line-length
-      return `https://${GeminiHost}/?resize_to=${type}&width=${width}&height=${height}&quality=${ImageQuality}&src=${encodeURIComponent(
-        imageURL
-      )}`
-    } else {
-      return null
+      if (disableGemini) {
```
sweir27: Ah I see it here-- curious in which cases this is necessary?

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/ImageZoomView.tsx
```diff
@@ -0,0 +1,231 @@
...
+        scrollViewRef.current.scrollResponderZoomTo({
+          x: locationX - w / 2,
+          y: locationY - h / 2,
+          width: w,
+          height: h,
+        })
+      }
+    })
+
+    // as a perf optimisation, when doing the 'zoom im' transition, we only render the
```
sweir27: `zoom in` 🙈

package.json
```diff
@@ -79,6 +79,8 @@
     "@mapbox/react-native-mapbox-gl": "l2succes/react-native-mapbox-gl#fix-gesture-recognizer",
     "@react-native-community/cameraroll": "1.0.5",
     "lodash": "4.17.13",
+    "mobx": "^5.13.0",
```
sweir27: @ds300 When this PR is out of `Draft` form, would you mind adding some explanation here for choosing mobx as your state management system? We have some precedent for having an RFC-style writeup when adding new libraries like this, especially one that's (I think) new to Artsy!.

ds300: I would like to add a follow-up task to remove mobx once this gets out.

The rationale for using it originally was as follows:

1. We needed a way to selectively update leaf components without drilling props down through their parents. This was to avoid excess re-renders of ancestor components which can be quite costly on iOS.
2. The normal way to do this is with context, but we would have needed one context value per state value, to avoid independent state values affecting each other's consuming components. This is not a very idiomatic way to do things, and is quite cumbersome to set up.
3. Another concern in the days of hooks is that sometimes callbacks need to reference state values. With `useState` and plain `useContext` this means that the callbacks need to be regenerated, which can again cause re-renders. In addition to that, callbacks only care about the value of a piece of state when they actually get called. So if you have a callback that gets called way less often than the state it reads gets updated, you are doing a lot of unnecessary re-rendering. Mobx can make all of that go away through its dark proxy magic.

However I think we should not be using mobx going forward. It's fine if you really buy into it as a whole-app thing, but I don't want it to be another strange and complicated thing that people have to learn in order to work on the image carousel. It's behaviour is surprising enough already.

src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarouselContext.tsx
```diff
@@ -0,0 +1,104 @@
...
+  })
+  const tracking = useTracking()
+
+  return useMemo(
+    () => ({
+      state,
+      images,
+      embeddedImageRefs,
+      embeddedFlatListRef,
+      dispatch: (action: ImageCarouselAction) => {
```
damassi: Why manually create `dispatch` vs using `useReducer`?

src/lib/Scenes/Artwork/Components/ImageCarousel/qa-script.md
```diff
@@ -0,0 +1,53 @@
...
+    - [ ] The image index fades out when I zoom in
+    - [ ] I can dismiss the full screen mode using the close button
+    - [ ] I can dismiss the full screen mode by swiping vertically
+    - [ ] If I switch to a different image than the one I tapped on and dismiss full screen mode, then
+          the image and pagination dot on the main carousel update to match.
+    - After having zoomed in
+      - [ ] I can still swipe between images
+      - [ ] If I swipe to a different image, wait for the animation to finish, and then swipe back again, the zoom level resets
+      - [ ] The image index fades back in when I zoom out
+      - [ ] I can double tap to zoom out
```
damassi: 👍 💯

src/lib/Scenes/Artwork/Components/ImageCarousel/ImageWithLoadingState.tsx
```diff
@@ -18,74 +20,65 @@ interface ImageWithLoadingStateProps {
...
+            backgroundColor: "white",
             flex: 1,
           }}
         >
-          <Spinner />
+          <Animated.View
+            style={{
+              opacity: spinnerOpacity,
+              // give the image a subtle silhouette while the spinner is displaying
+              // to keep the balance of the page and set an appropriate user expectation
```
damassi: Thanks for all of the code-level comments / documentation left throughout this work, it's very helpful 👍

package.json
damassi: > I would like to add a follow-up task to remove mobx once this gets out

What are you thinking for afterwards? I agree Mobx is a bit confounding, especially for a single component (as opposed to app-level).

src/lib/Components/Artist/Articles/Article.tsx
```diff
@@ -5,7 +5,7 @@ import { StyleSheet, Text, TextStyle, TouchableWithoutFeedback, View, ViewProper
 
 import fonts from "lib/data/fonts"
 import SwitchBoard from "lib/NativeModules/SwitchBoard"
-import ImageView from "../../OpaqueImageView"
+import ImageView from "../../OpaqueImageView/OpaqueImageView"
```
damassi: #minor - elsewhere, paths to `OpaqueImageView` are absolute

src/lib/Components/OpaqueImageView/createGeminiUrl.tsx
```diff
@@ -0,0 +1,19 @@
+export function createGeminiUrl({
```
damassi: Very similar functionality is scattered throughout Emission; might be worth adding a `TODO` to replace refs with this utility function, and to move this util up a level at that time.

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/ImageCarouselCloseButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { observer } from "mobx-react"
+import React, { useContext } from "react"
+import { TouchableOpacity, View } from "react-native"
+import { ImageCarouselContext } from "../ImageCarouselContext"
+import { boxShadow } from "./boxShadow"
+
+// taken from https://github.com/artsy/eigen/blob/0831853cb574566415f3bd8b3908b26b61f61eec/Artsy/View_Controllers/Util/ARNavigationController.m#L125
+const CLOSE_BUTTON_MARGIN = 12
+
+export const ImageCarouselCloseButton: React.FC<{ onClose(): void }> = observer(({ onClose }) => {
```
damassi: https://mobx.js.org/refguide/observer-component.html

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/ImageCarouselCloseButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <View
+          style={{
+            width: 40,
+            height: 40,
+            paddingLeft: CLOSE_BUTTON_MARGIN,
+            paddingTop: CLOSE_BUTTON_MARGIN,
+            paddingRight: 20,
+            paddingBottom: 20,
+          }}
+        >
```
damassi: Can this be `<Box>`? Under the hood we use `View`

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/ImageCarouselCloseButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <View
```
damassi: And can this be `<Flex>`?

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/ImageZoomView.tsx
```diff
@@ -0,0 +1,243 @@
...
+          y: locationY - h / 2,
+          width: w,
+          height: h,
+        })
+      }
+    })
+
+    // as a perf optimisation, when doing the 'zoom in' transition, we only render the
+    // current zoomable image
+    // in place of the other images we just render a blank box
```
damassi: extra line break in comment

damassi:
Superb work @ds300. Code is clear and easy to follow 👍 

As previously mentioned introducing the Mobx reactive model might a bit confusing for some folks (vs, say, Redux, which the team already knows and has similar performance-centric characteristics), but even that might be overkill for a single component. Since the idea is to eek out as much perf as possible here maybe it's worth it, however.

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/ImageCarouselCloseButton.tsx
ds300: I've used View in a bunch of places where I could have used Flex and Box. This was mainly to keep the time between tapping an image and showing the full screen carousel as short as possible. styled-components adds some overhead when computing and applying styles, which doesn't seem worth the DX of shorthand property names to me.

src/lib/Components/OpaqueImageView/OpaqueImageView.tsx
ds300: The `OpaqueImageView` uses the width+height props you pass into the component to generate the gemini URL. This was a problem for me in situations where the thumbnail in the embedded carousel was a different size than the image in the full screen carousel. It meant that the two images had different URLs and therefore the full screen one would still be loading while the transition into full screen mode happened.

So I decided to optionally decouple gemini url generation to guarantee a cache hit when entering full screen mode.

sweir27: Ah! Maybe the naming is confusing me then-- can we call this bool `resizeImage` or `useDefaultImageUrl` or something like that?

`disableGemini` sounds more like we're talking about an entire system being available/unavailable which threw me off.

ds300: Renamed to `rawImageURL`, good suggestion! :+1:

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/ImageCarouselCloseButton.tsx
ds300: good catch!

src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarouselContext.tsx
ds300: `useReducer` is a way to update immutable state, but I opted for mutable state with `mobx`. it doesn't really make sense to mix the two.

src/lib/Components/OpaqueImageView/createGeminiUrl.tsx
ds300: Can you point out an example? I'm struggling to find it.

package.json
ds300: Not sure. I searched around for new hooks-friendly libs but couldn't find anything small and well-established enough to make sense. Rolled my own tiny util at one point https://gist.github.com/ds300/db215cdcbd9844a4d65f78bd58de7252 but went with mobx instead because I thought it would be more maintainable in the long run. Now I kinda wish I'd stuck with the one in that gist.

damassi: Heh, that gist is a really good study.

src/lib/Components/OpaqueImageView/createGeminiUrl.tsx
damassi: Looking around, this comment applies to Reaction, not Emission.

ds300:
Thanks for the comments @damassi :pray: Super helpful stuff!

@sweir27 I think this is 🍏

sweir27:
@ds300 this looks good!

Some merge conflicts but #mergeOnGreen !!! 🎉

alloy:
Has there been a RFC to add these Mobx dependencies?

ashfurrow:
@alloy no, and that's a good point, although our use of it is still up for discussion: https://github.com/artsy/emission/pull/1752#discussion_r308811841 We pulled it in here to unblock this work. Maybe an RFC would be a place where @ds300 could discuss mobx and motivations to seek alternatives?

Separately, we should leverage Danger or Peril to flag PRs that add dependencies as needing RFCs. [The RFC to require RFCs](https://github.com/artsy/README/issues/117) is good, but without tooling to remind us, we'll forget (_I_ forgot, in this case).

alloy:
@ashfurrow Thanks for the link. Yeah it would be good to open a RFC for further discussion.

package.json
alloy: Unsure of all the details and we should discuss that on a dependency RFC, but want to add that Relay’s local schema for state management might also be an option. We have no experience with it yet, but, if anything, you can trust Relay for performance needs.

ds300:
I'd prefer to avoid further discussion on mobx if possible because we're definitely going to remove it, hopefully before we ship our first eigen release with the new artwork page. There have already been discussions around why it was used in the first place and why we want to get rid of it but haven't gotten rid of it yet, but if you want to open those discussions to the wider team then that's fine with me. I personally don't feel it warrants wider discussion but if others are interested I can provide a summary of the discussion so far.

alloy:
I’m absolutely fine not discussing Mobx, especially if we don’t include it in a prod release 👍 But if it gets replaced with another dep that Emission doesn’t already have it definitely needs a RFC.

PR #1753, opened 2019-07-25
Remove relay mask on OtherWorks when deciding whether to render the section

sweir27:
Per this slack discussion: https://artsy.slack.com/archives/C02BAQ5K7/p1564063430017600

It's bad practice to unmask relay data, so this updates the `Artwork.tsx` component to explicitly pull in the data it needs to decide whether or not to render the `OtherWorks` component.

#trivial #skip_new_tests


src/lib/Scenes/Artwork/Components/OtherWorks/OtherWorks.tsx
```diff
@@ -8,7 +8,7 @@ import { createFragmentContainer, graphql } from "react-relay"
 import { ContextGridCTA } from "./ContextGridCTA"
 import { Header } from "./Header"
 
-export const populatedGrids = (grids: OtherWorks_artwork["contextGrids"]) => {
+export const populatedGrids = grids => {
```
alloy: Why remove the type here? Functions preferably do type their params.

sweir27: I removed this typing since it no longer applies (it did when we were re-using the fragment container's query).

But I can add something back.

sweir27:
@alloy is this good now?

PR #1755, opened 2019-07-26
[Artwork view] Artwork commercial flow

kierangillen:
- Adds inquiry and BNMO flow to Artwork view

## Tickets

- https://artsyproduct.atlassian.net/browse/PURCHASE-1338
- https://artsyproduct.atlassian.net/browse/PURCHASE-1310
- https://artsyproduct.atlassian.net/browse/PURCHASE-1311
- https://artsyproduct.atlassian.net/browse/PURCHASE-1312

## BNMO

Screen Shot 2019-07-26 at 5 36 27 PM

## Inquiry

Screen Shot 2019-07-26 at 5 38 47 PM

#skip_new_tests


src/lib/Scenes/Artwork/Components/CommercialButtons.tsx
```diff
@@ -0,0 +1,226 @@
...
+    isCommittingCreateOfferOrderMutation: false,
+    isCommittingCreateOrderMutation: false,
+  }
+
+  onMutationError(error) {
+    // FIXME: Handle error
+    console.error("src/lib/Scenes/Artwork/Components/CommercialButtons.tsx", error)
+  }
+
+  handleCreateOrder() {
```
sweir27: When we build out the `Buy now` button for auctions, we may want to re-use this logic.

src/lib/Scenes/Artwork/Components/CommercialButtons.tsx
```diff
@@ -0,0 +1,226 @@
...
+            })
+          },
+          onError: this.onMutationError,
+        })
+      }
+    })
+  }
+
+  handleInquiry = () => {
+    SwitchBoard.presentModalViewController(this, `/inquiry/${this.props.artwork.slug}`)
```
sweir27: Is there a good reason to use `slug` here instead of `internalID`?

src/lib/Scenes/Artwork/Components/CommercialButtons.tsx
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  renderButtons = () => {
+    const { is_acquireable, is_offerable, is_inquireable } = this.props.artwork
+    const { isCommittingCreateOrderMutation, isCommittingCreateOfferOrderMutation } = this.state
+    if (is_offerable && is_acquireable) {
```
sweir27: #minor can we add a space above this line? It's getting a little cramped

src/lib/Scenes/Artwork/Components/CommercialButtons.tsx
```diff
@@ -0,0 +1,226 @@
...
+}
+
+export const CommercialButtonsFragmentContainer = createFragmentContainer(CommercialButtons, {
+  artwork: graphql`
+    fragment CommercialButtons_artwork on Artwork {
+      slug
+      internalID
+      is_acquireable
+      is_offerable
+      is_biddable
```
sweir27: Are we using this yet?

src/lib/Scenes/Artwork/Components/__tests__/CommercialButtons-tests.tsx
```diff
@@ -0,0 +1,79 @@
...
+describe("CommercialButtons", () => {
+  artwork = {
+    slug: "abbas-kiarostami-untitled-7",
+    internalID: "5b2b745e9c18db204fc32e11",
+    is_acquireable: false,
+    is_offerable: false,
+    is_biddable: false,
+    is_inquireable: true,
+  }
+  it("renders inquierable button", () => {
```
sweir27: These are weird words 😅 , but should be `inquireable` (and `acquireable` below) to match internal naming.

src/lib/Scenes/Artwork/Components/CommercialButtons.tsx
sweir27: Also, I don't recall the internals but should we also handle an edition-set picker for inquiries and pass along the edition set ID accordingly?

kierangillen: I don't know the answer to the second question, what or who would be a good reference for that?

src/lib/Scenes/Artwork/Components/__tests__/CommercialButtons-tests.tsx
kierangillen: Ah yeah, fixed the wording here.

src/lib/Scenes/Artwork/Components/CommercialButtons.tsx
kierangillen: I think of `slug` as an external ID which should be used for URL's and similar client facing situations. Also this is the URL used elsewhere, as the routing works.

sweir27: Since this is all internal routing within the app, it seems that `internalID` would be more appropriate? Not a big deal in any case. For other places where we might benefit from caching or other niceties, we may prefer `internalID` out of necessity but I doubt that would come into play much for the inquiry flow.

src/lib/Scenes/Artwork/Components/__tests__/CommercialButtons-tests.tsx
```diff
@@ -0,0 +1,111 @@
...
+describe("CommercialButtons", () => {
+  artwork = {
+    slug: "abbas-kiarostami-untitled-7",
+    internalID: "5b2b745e9c18db204fc32e11",
+    is_acquireable: false,
+    is_offerable: false,
+    is_inquireable: true,
+  }
+  it("renders button for Contact Gallery button if is_inquireable", () => {
+    const component = mount(<CommercialButtons relay={{ environment: {} } as RelayProp} artwork={artwork as any} />)
```
sweir27: Can we properly type this `artwork`?

src/lib/Scenes/Artwork/Components/__tests__/CommercialButtons-tests.tsx
```diff
@@ -0,0 +1,111 @@
...
+let artwork = {
+  slug: "abbas-kiarostami-untitled-7",
+  internalID: "5b2b745e9c18db204fc32e11",
+  is_acquireable: false,
+  is_offerable: false,
+  is_inquireable: true,
+}
+
+describe("CommercialButtons", () => {
+  artwork = {
```
sweir27: Is this definition necessary given it matches the default above?

src/lib/Scenes/Artwork/Components/__tests__/CommercialButtons-tests.tsx
```diff
@@ -0,0 +1,111 @@
...
+      is_inquireable: false,
+    }
+
+    const component = mount(<CommercialButtons relay={{ environment: {} } as RelayProp} artwork={artwork as any} />)
+    const BuyNowButton = component.find(Button).at(0)
+
+    relay.commitMutation = jest.fn()
+    BuyNowButton.props().onPress()
+
+    expect(relay.commitMutation).toHaveBeenCalled()
```
sweir27: Instead of verifying that we call `commitMutation`, can we instead mock the mutation response directly and verify that after clicking the button/receiving that response, we call `Switchboard` with the correct value?

src/lib/Scenes/Artwork/Components/__tests__/CommercialInformation-tests.tsx
```diff
@@ -62,13 +72,65 @@ describe("CommercialInformation", () => {
...
+    )
+    expect(
+      component
+        .find(Sans)
+        .at(0)
+        .render()
+        .text()
+    ).toMatchInlineSnapshot(`"Contact For Price"`)
+  })
+  it("renders seller info correctly for non-commercial works", () => {
```
sweir27: This test looks the same as the one above-- was it meant to be updated for the non-commercial case?

src/lib/Scenes/Artwork/Components/__tests__/CommercialInformation-tests.tsx
```diff
@@ -62,13 +72,65 @@ describe("CommercialInformation", () => {
...
+    expect(
+      component
+        .find(Sans)
+        .at(0)
+        .render()
+        .text()
+    ).toMatchInlineSnapshot(`"Contact For Price"`)
+  })
+
+  it("renders artwork availability correctly", () => {
```
sweir27: Actually all three of these look the same 😅

src/lib/Scenes/Artwork/Components/__tests__/CommercialEditionSetInformation-tests.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const component = mount(
+      <Theme>
+        <CommercialEditionSetInformation setEditionSetId={() => null} artwork={artwork as any} />
+      </Theme>
+    )
+
+    expect(component.html()).toContain("$1")
+  })
+
+  it("changes display price to selected edition set", () => {
```
sweir27: This likely belongs in the test of a higher-level component, but two cases that we should be sure to test are:
- When you have one edition set and the work is enabled for buy now/make offer, when you click one of these buttons, we send the edition set ID with our input payload
- When you have multiple edition sets, when you select one of these, that's the ID that gets sent with the input payload.

src/lib/Scenes/Artwork/Components/__tests__/CommercialButtons-tests.tsx
```diff
@@ -0,0 +1,111 @@
...
+  })
+
+  it("commits the Buy Now mutation", () => {
```
sweir27: #minor but this should probably be `"commits the Make Offer mutation"`

data/complete.queryMap.json
```diff
@@ -108,6 +108,7 @@
...
+  "4c30eb7c3732d005b5537e60c0c419dd": "query CommercialButtonsTestsQuery {\n  artwork(id: \"artworkID\") {\n    ...CommercialButtons_artwork\n    id\n  }\n}\n\nfragment CommercialButtons_artwork on Artwork {\n  slug\n  internalID\n  is_acquireable\n  is_offerable\n  is_inquireable\n}\n",
```
ashfurrow: 😭 someday I want to resurrect https://github.com/artsy/emission/pull/1300 to fix these, if anyone else is interested!

src/lib/Scenes/Artwork/Components/CommercialButtons.tsx
```diff
@@ -0,0 +1,226 @@
...
+    }
+    this.setState({ isCommittingCreateOrderMutation: true }, () => {
+      if (relay && relay.environment) {
+        commitMutation<CommercialButtonsOrderMutation>(relay.environment, {
+          mutation: graphql`
+            mutation CommercialButtonsOrderMutation($input: CommerceCreateOrderWithArtworkInput!) {
+              commerceCreateOrderWithArtwork(input: $input) {
+                orderOrError {
+                  ... on CommerceOrderWithMutationSuccess {
+                    __typename
```
ashfurrow: Should this go on the `orderOrError`, rather than within the `CommerceOrderWithMutationSuccess` spread?

src/lib/Scenes/Artwork/Components/CommercialButtons.tsx
```diff
@@ -0,0 +1,226 @@
...
+    }
+    this.setState({ isCommittingCreateOfferOrderMutation: true }, () => {
+      if (relay && relay.environment) {
+        commitMutation<CommercialButtonsOfferOrderMutation>(relay.environment, {
+          mutation: graphql`
+            mutation CommercialButtonsOfferOrderMutation($input: CreateOfferOrderWithArtworkInput!) {
+              ecommerceCreateOfferOrderWithArtwork(input: $input) {
+                orderOrError {
+                  ... on OrderWithMutationSuccess {
+                    __typename
```
ashfurrow: Same here, I think?

src/lib/Scenes/Artwork/Components/CommercialEditionSetInformation.tsx
```diff
@@ -0,0 +1,126 @@
...
+  setEditionSetId: (editionSetID: string) => void
+}
+
+interface State {
+  selectedEdition: EditionSet
+}
+
+export class CommercialEditionSetInformation extends React.Component<Props, State> {
+  state = {
+    selectedEdition: this.props.artwork.editionSets[0],
```
ashfurrow: Is there any chance of this erroring if `this.props.artwork.editionSets` is `undefined`?

kierangillen: Good call fixing this.

src/lib/Scenes/Artwork/Components/CommercialButtons.tsx
kierangillen: I'm not sure about the answer for this, is there someone else with more context we could reach out to?

ashfurrow: Thinking about this again, I'm fairly certain it should go into `orderOrError`, since `__typename` exists on both, and not just `CommerceOrderWithMutationSuccess`. Here's another use of it:

https://github.com/artsy/emission/blob/6567bdce2e708520f6f02a8a837fe8e8b87052cb/src/lib/Components/Inbox/Conversations/ConversationSnippet.tsx#L188-L190

PR #1765, opened 2019-08-01
Add input textContentType for registration BillingAddress

ashfurrow:
During last week's Auctions QA session, I tried to register for a sale on an iPhone. I realized that the billing address inputs lacked `textContentType`, which is a property that tells iOS what data goes in each input field. This allows users to tap (for example) their name to get it input into the name field; this drastically improves the user experience of entering an address. 

Here's what it looks like in the simulator (`SIM-VALUE` would be replaced by the user's actual data, on-device): 

Screen Shot 2019-08-01 at 10 16 19

During this work, I noticed that the `StyledInput` component was not well-typed, so I added some type annotations to get type-checking and autocomplete working:

Screen Shot 2019-08-01 at 10 08 29

### DIFF
--- a/src/lib/Components/Bidding/Screens/BillingAddress.tsx
+++ b/src/lib/Components/Bidding/Screens/BillingAddress.tsx
@@ -21,7 +21,7 @@ import { validatePresence } from "../Validators"
 import { BackButton } from "../Components/BackButton"
 import { BiddingThemeProvider } from "../Components/BiddingThemeProvider"
 import { Container } from "../Components/Containers"
-import { Input } from "../Components/Input"
+import { Input, InputProps } from "../Components/Input"
 import { Title } from "../Components/Title"
 import { Address, Country } from "../types"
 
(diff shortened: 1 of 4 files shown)


src/lib/Components/Bidding/Screens/BillingAddress.tsx
```diff
@@ -35,15 +35,19 @@ interface StyledInputInterface {
...
     <Serif size="3" mb={2}>
       {label}
     </Serif>
-    <Input mb={3} error={Boolean(error)} {...props} />
-    {!!error && (
+    <Input mb={3} error={Boolean(errorMessage)} {...props} />
+    {!!errorMessage && (
       <Sans size="2" color="red100">
-        {error}
+        {errorMessage}
```
ashfurrow: I had to rename `error` to `errorMessage` because `InputProps` already has an `error?: boolean` field that conflicted, and I wanted to extend that interface.

yuki24:
LGTM, thanks!

PR #1767, opened 2019-08-01
Cleans up old, duplicated docs

ashfurrow:
I reviewed our Eigen+Emission docs for outdated or duplicated content ([here's the Eigen PR](https://github.com/artsy/eigen/pull/2878)).

This PR also deletes the VSCode doc because VSCode has since become the standard Artsy editor, and there's nothing in the doc that was specific to Emission (just Orta's setup to make VSCode behave like Xcode (what a chilling thought, indeed)).


alloy:
Nice one ✨

PR #1768, opened 2019-08-01
Fix copy on registration pending view

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-508

Just a quick copy update:

## Screenshot

Screen Shot 2019-08-01 at 5 06 14 PM

### DIFF
--- a/src/lib/Components/Bidding/Screens/RegistrationResult.tsx
+++ b/src/lib/Components/Bidding/Screens/RegistrationResult.tsx
@@ -38,9 +38,9 @@ const registrationCompleteMessage = {
 const registrationPendingMessage = {
   title: "Registration pending",
   description:
-    "You will receive an email from an Artsy specialist once your registration has been confirmed.\n" +
+    "Artsy is reviewing your registration and you will receive an email when it has been confirmed. Please email specialist@artsy.net with any questions.\n" +
     "\n" +
-    "In the meantime, you can still view works and watch lots that you’re interested in.",
+    "In the meantime, you can still view works and watch lots you’re interested in.\n",
 }
 
 const registrationErrorMessage = {
@@ -112,7 +112,7 @@ export class RegistrationResult extends React.Component<RegistrationResultProps,
             </Flex>
           </View>
           <Button variant="secondaryOutline" onPress={this.exitBidFlow} block width={100}>
-            Continue
+            {status === RegistrationStatus.RegistrationStatusPending ? "View works in this sale" : "Continue"}
           </Button>
         </Container>
       </BiddingThemeProvider>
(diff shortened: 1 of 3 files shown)


src/lib/Components/Bidding/Screens/RegistrationResult.tsx
```diff
@@ -38,9 +38,9 @@ const registrationCompleteMessage = {
 const registrationPendingMessage = {
   title: "Registration pending",
   description:
-    "You will receive an email from an Artsy specialist once your registration has been confirmed.\n" +
+    "Artsy is reviewing your registration and you will receive an email when it has been confirmed. Please email specialist@artsy.net with any questions.\n" +
```
dblandin: Do we usually wrap email addresses in a mailto link?

src/lib/Components/Bidding/Screens/RegistrationResult.tsx
```diff
@@ -38,9 +38,9 @@ const registrationCompleteMessage = {
...
     "\n" +
-    "In the meantime, you can still view works and watch lots that you’re interested in.",
+    "In the meantime, you can still view works and watch lots you’re interested in.\n",
```
dblandin: Is this new trailing new line for padding?

src/lib/Components/Bidding/Screens/RegistrationResult.tsx
dleve123: Considering that we use a [`mailto` for the registration failed component](https://github.com/artsy/emission/blob/04af01b03b463cda21e620f67b42a040e43f7556/src/lib/Components/Bidding/Screens/RegistrationResult.tsx#L48), I think it's safe to assume that we should do so here. Can you make that change @yuki24?

src/lib/Components/Bidding/Screens/RegistrationResult.tsx
dleve123: +1 to this question? Is this newline addition intentional?

src/lib/Components/Bidding/Screens/RegistrationResult.tsx
yuki24: Okay, the description can actually be markdown and it is very easy to make it a link:

Screen Shot 2019-08-02 at 10 51 50 AM

This link does not work on Emission due to an unhandled native event, but if I recall correctly this is handled properly in Eigen. We should QA this anwyay to make sure it works in the actual app.

yuki24:
#mergeongreen

src/lib/Components/Bidding/Screens/RegistrationResult.tsx
yuki24: There seems to be a bug around the email click but there is another PR that addresses it: https://github.com/artsy/emission/pull/1769 so it's likely that it still crashes but it'll be fixed eventually.

PR #1770, opened 2019-08-02
Design updates

kierangillen:
- Fixes `Make offer` mutation.
- Fixes `View all` button in dated `ForYou` `ArtworkCarousel` component.
- Makes `ShowItemRow` buttons without background `inline` so they have no padding.
- Adds space under Artist name (request from design).
- `Contact For Price` now renders as `Contact for price` (request from design).
- Passes artist name to consign text, e.g. `Want to sell a work by Santa Claus?` instead of `this artist` to be more explicit (request from design).

## ShowItemRowButton change

- Before

Screen Shot 2019-08-02 at 11 33 02 AM

- After

Screen Shot 2019-08-02 at 11 34 44 AM

#trivial
#skip_new_tests


src/lib/Scenes/Artwork/Components/CommercialInformation.tsx
```diff
@@ -3,7 +3,7 @@ import { CommercialInformation_artwork } from "__generated__/CommercialInformati
 import { capitalize } from "lodash"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
-import { ArtworkExtraLinks } from "./ArtworkExtraLinks"
```
sweir27: This was named this way to match our need to add additional, similar "links" for other flows.

See the BNMO case:
image

And the auctions case
image

kierangillen: Ah I see, I'll name it back to that!

PR #1785, opened 2019-08-09
[PURCHASE-1358] Attribution class FAQ is cut off

lilyfromseattle:
[PURCHASE-1358]

Back button and bottom nav were overlapping the attribution class FAQ. I added more padding to its container.

__Before:__
Screen Shot 2019-08-09 at 1 55 43 PM
Screen Shot 2019-08-09 at 1 55 35 PM

__After:__
Screen Shot 2019-08-09 at 1 54 27 PM
Screen Shot 2019-08-09 at 1 54 45 PM

[PURCHASE-1358]: https://artsyproduct.atlassian.net/browse/PURCHASE-1358

#trivial

### DIFF
--- a/src/lib/Scenes/ArtworkAttributionClassFAQ/ArtworkAttributionClassFAQ.tsx
+++ b/src/lib/Scenes/ArtworkAttributionClassFAQ/ArtworkAttributionClassFAQ.tsx
@@ -39,7 +39,7 @@ export class ArtworkAttributionClassFAQ extends React.Component<Props> {
     return (
       <Theme>
         <ScrollView>
-          <Box pt={this.props.safeAreaInsets.top} pb={this.props.safeAreaInsets.top} px={2}>
+          <Box pt={this.props.safeAreaInsets.top + 3} pb={this.props.safeAreaInsets.top + 4} px={2}>
             <Spacer mt={3} />
             <Serif mb={2} size="8">
               Artwork classifications


src/lib/Scenes/ArtworkAttributionClassFAQ/ArtworkAttributionClassFAQ.tsx
```diff
@@ -39,7 +39,7 @@ export class ArtworkAttributionClassFAQ extends React.Component<Props> {
     return (
       <Theme>
         <ScrollView>
-          <Box pt={this.props.safeAreaInsets.top} pb={this.props.safeAreaInsets.top} px={2}>
+          <Box pt={this.props.safeAreaInsets.top + 3} pb={this.props.safeAreaInsets.top + 4} px={2}>
```
ashfurrow: I don't think that manually adding padding here is the right approach. The other components in Emission already avoid the padding somehow, I believe by using a flex container. There are two approaches I can find: 

Adding a containing flex wrapper:

https://github.com/artsy/emission/blob/b65d5727e24619d2c42e43495f9c8e2149a51140/src/lib/Containers/Conversation.tsx#L24-L26

Or modifying the props of the outermost container:

https://github.com/artsy/emission/blob/b65d5727e24619d2c42e43495f9c8e2149a51140/src/lib/Components/Consignments/Screens/Welcome.tsx#L97

I think the second approach would work here, but I haven't tested it out yet 🙈

alloy:
@lilyfromseattle Any updates on this?

ds300:
I'm just going to merge Lily's version here. It seems `flex` is not related because the container bounds do not respect the menu/back button. And, as far as I can tell, the other pages in emission also use manual padding to solve this problem.

PR #1786, opened 2019-08-10
[Consignments] Switch to use relay and fix query to use camelCase

alloy:
Fixes https://artsyproduct.atlassian.net/browse/PLATFORM-1684


src/lib/Components/Consignments/Screens/ConsignmentsArtist.tsx
```diff
@@ -50,19 +46,25 @@ export default class Artist extends React.Component<Props, State> {
     this.searchForQuery(text)
   }
 
+  // TODO: Add throttling
```
alloy: This is for another time

PR #1789, opened 2019-08-12
[PURCHASE-1143] Removes artwork set functionality

ashfurrow:
This PR removes the concept of an "artwork set", which is used by Eigen to provide left-to-right swiping between artworks in a grid. 

This is a companion PR to https://github.com/artsy/eigen/pull/2884, which it is blocking. 

#skip_new_tests


kierangillen:
#mergeongreen

PR #1791, opened 2019-08-12
[Artwork View] Auction flow buttons

kierangillen:
- Adds Auction bid buttons to Artwork view
- Moves `BidButton`, `BuyNowButton` and `MakeOfferButton` into children sub components of `CommercialButtons`

buttons_registrationcomplete

#skip_new_tests
- Added `#skip_new_tests` for rearranged button files which tests are covered in `CommercialButtons-tests` file


src/lib/Scenes/Artwork/Components/CommercialButtons/BidButton.tsx
```diff
@@ -0,0 +1,153 @@
+import { Button, Sans } from "@artsy/palette"
+import { BidButton_artwork } from "__generated__/BidButton_artwork.graphql"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import React from "react"
+import { createFragmentContainer, graphql, RelayProp } from "react-relay"
+
+export const PREDICTION_URL = "https://live-staging.artsy.net"
```
ashfurrow: We'll probably want this to be `live.artsy.net`. I've checked and we literally do the same thing for both:

https://github.com/artsy/eigen/blob/a677b504ccdb7fad552f7749912d2f048a925198/Artsy/App/ARSwitchBoard.m#L325-L326

But using the prod URL feels better.

src/lib/Scenes/Artwork/Components/CommercialButtons/BidButton.tsx
```diff
@@ -0,0 +1,153 @@
...
+  relay: RelayProp
+}
+
+export class BidButton extends React.Component<BidButtonProps> {
+  setMaxBid = (newVal: number) => {
+    this.setState({ selectedMaxBidCents: newVal })
+  }
+
+  handleBid = () => {
+    SwitchBoard.presentModalViewController(this, `/bid/${this.props.artwork.slug}`)
```
ashfurrow: I don't think this is the right URL format, it should be `/auction/:id/bid/:artwork_id`, right?

src/lib/Scenes/Artwork/Components/CommercialButtons/BidButton.tsx
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  redirectToRegister = () => {
+    const { sale } = this.props.artwork
+    SwitchBoard.presentModalViewController(this, `/auction-registration/${sale.slug}`)
```
ashfurrow: These `presentModalViewController` calls should be `presentNavigationViewController` (Eigen knows to present the right controllers modally, and to present the others as pushes.)

src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx
```diff
@@ -0,0 +1,79 @@
...
+import React from "react"
+import { createFragmentContainer, graphql, RelayProp } from "react-relay"
+import { BidButtonFragmentContainer as BidButton } from "./BidButton"
+import { BuyNowButtonFragmentContainer as BuyNowButton } from "./BuyNowButton"
+import { MakeOfferButtonFragmentContainer as MakeOfferButton } from "./MakeOfferButton"
+
+export interface CommercialButtonProps {
+  artwork: CommercialButtons_artwork
+  relay: RelayProp
+  editionSetID?: string
```
ashfurrow: Can we add a comment here describing why we're using a prop for edition set ID, and not including it in the fragment? It makes sense to me now, but won't in six months.

src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export class CommercialButtons extends React.Component<CommercialButtonProps> {
+  handleInquiry = () => {
+    SwitchBoard.presentModalViewController(this, `/inquiry/${this.props.artwork.slug}`)
+  }
+
+  handleBid = () => {
+    SwitchBoard.presentModalViewController(this, `/bid/${this.props.artwork.slug}`)
```
ashfurrow: Same here with `presentNavigationViewController` vs `presentModalViewController`. Also I don't think this `handleBid` is getting called anywhere?

src/lib/Scenes/Artwork/Components/CommercialButtons/BidButton.tsx
kierangillen: Ah thanks

src/lib/Scenes/Artwork/Components/CommercialButtons/BidButton.tsx
kierangillen: Cool updating.

src/lib/Scenes/Artwork/Components/CommercialButtons/BidButton.tsx
```diff
@@ -0,0 +1,154 @@
...
+      const myLastMaxBid = hasMyBids && myLotStanding.mostRecentBid.maxBid.cents
+      const increments = artwork.saleArtwork.increments.filter(increment => increment.cents > (myLastMaxBid || 0))
+      const firstIncrement = increments && increments.length && increments[0]
+      const incrementCents = firstIncrement && firstIncrement.cents
+
+      return (
+        <Button width={100} size="large" block onPress={() => this.redirectToBid(incrementCents)}>
+          {hasMyBids ? "Increase max bid" : "Bid"}
+        </Button>
+      )
```
ds300: woah the logic in this method is hard to follow 😅 I don't know if it's possible to simplify so this is not a blocking comment. I'd be a bit scared to edit this code though haha.

src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx
```diff
@@ -0,0 +1,76 @@
...
+export class CommercialButtons extends React.Component<CommercialButtonProps> {
+  handleInquiry = () => {
+    SwitchBoard.presentNavigationViewController(this, `/inquiry/${this.props.artwork.slug}`)
+  }
+
+  renderButtons = () => {
+    const { artwork } = this.props
+    const { isAcquireable, isOfferable, isInquireable, isInAuction } = artwork
+
+    if (isInAuction && artwork.sale && !artwork.sale.isClosed) {
```
ds300: not a big deal but these two latter conditions are already taken care of in the bid button render method, so you could remove these checks and also the sale.isClosed selection from this fragment.

src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx
```diff
@@ -0,0 +1,76 @@
...
+          Contact gallery
+        </Button>
+      )
+    } else {
+      return <></>
+    }
+  }
+
+  render() {
+    return <>{this.renderButtons()}</>
```
ds300: haha this indirection seems a bit useless now 🙈

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/BidButton-tests.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+    it("with bidder registration pending approval", async () => {
+      const artwork = merge({}, ArtworkFromLiveAuctionRegistrationOpen, BidderPendingApproval)
+      const wrapper = await getWrapper(artwork)
+
+      expect(wrapper.find(Button).text()).toContain("Enter live bidding")
+    })
+
+    it("with registered bidder", async () => {
+      const artwork = merge({}, ArtworkFromLiveAuctionRegistrationOpen, RegisteredBidder)
```
ds300: #nonblocking why use lodash's merge instead of `{..spreading}`?

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/CommercialButtons-tests.tsx
```diff
@@ -43,76 +34,78 @@ const componentWithQuery = async ({ mockArtworkData, mockOrderMutationResults, m
...
-      <CommercialButtons
-        relay={{ environment: {} } as RelayProp}
-        artwork={artwork as CommercialButtonProps["artwork"]}
-      />
-    )
-    expect(component.text()).toContain("Contact gallery")
+    const commercialButtons = await relayComponent({
+      artwork,
+    })
+    expect(commercialButtons.text()).toContain("Contact gallery")
```
ds300: This test case appears to not match the description. looks like the description is correct but the test case is a duplication of the one above.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/CommercialButtons-tests.tsx
```diff
@@ -43,76 +34,78 @@ const componentWithQuery = async ({ mockArtworkData, mockOrderMutationResults, m
   })
 }
 
+const relayComponent = async ({ artwork }) => {
```
ds300: would it make sense to add a test for the bid button in here?

src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx
kierangillen: Agreed!

src/lib/Scenes/Artwork/Components/CommercialButtons/BidButton.tsx
kierangillen: This was copied over from Reaction so gonna leave it as it reflects working logic.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/CommercialButtons-tests.tsx
kierangillen: Added!

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/CommercialButtons-tests.tsx
kierangillen: Fixed this.

src/lib/Scenes/Artwork/Components/CommercialButtons/__tests__/BidButton-tests.tsx
kierangillen: Going to keep as is for now.

ds300:
#mergeOnGreen

PR #1797, opened 2019-08-16
Initial setup for Collection page

l2succes:
Related: https://artsyproduct.atlassian.net/browse/GROW-1420

Lays the groundwork for building a first class collection page in our iOS app.
- Adds a new "Collection" option in home menu within emission app
- Maps native View Controller to a React Component via AppRegistry

Spec: https://zpl.io/V0wPor9

#skip_new_tests (Will have to add those)


src/lib/Scenes/Collection/Collection.tsx
```diff
@@ -0,0 +1,31 @@
+import { Box, Sans } from "@artsy/palette"
+import React, { Component } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+
+interface CollectionProps {
+  collection: any
```
mbilokonsky: how come this is marked as "any" - isn't a "collection" a graphQL query here?

l2succes: Yup that's correct, good catch. I just left it as `any` as a placeholder while the relay compiler was catching up with the module renaming and quick directory changes.

xtina-starr:
It seems like there is a failing spec.

PR #1801, opened 2019-08-19
Adds auction extra links

kierangillen:
- Adds Auction extra links

Screen Shot 2019-08-19 at 1 44 14 PM


src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
```diff
@@ -22,6 +22,14 @@ export class ArtworkExtraLinks extends React.Component<ArtworkExtraLinksProps> {
     SwitchBoard.presentNavigationViewController(this, `/inquiry/${artwork.slug}`)
   }
 
+  handleReadOurAuctionFAQsTap = () => {
+    // FIXME: Add auction FAQs navigation here
```
ds300: Might be good to add an alert here that says "TODO" or somesuch so people doing QA before this is implemented don't get confused.

src/lib/Scenes/Artwork/Components/AuctionCountDownTimer.tsx
```diff
@@ -27,11 +27,11 @@ export class AuctionCountDownTimer extends React.Component<AuctionCountDownTimer
     }
 
     this.interval = setInterval(() => {
-      if (DateTime.local() > sale.startAt) {
+      if (DateTime.local() > DateTime.fromISO(sale.startAt)) {
```
kierangillen: @lilyfromseattle This was the fix for the countdown timer not counting down

src/lib/Scenes/Artwork/Components/AuctionCountDownTimer.tsx
```diff
@@ -58,6 +58,7 @@ export class AuctionCountDownTimer extends React.Component<AuctionCountDownTimer
     return (
       <TimeRemaining
         labelWithTimeRemaining={sale.formattedStartDateTime}
+        labelWithoutTimeRemaining={sale.formattedStartDateTime}
```
kierangillen: @lilyfromseattle And here the label wasn't appearing after the auction had ended.

PR #1806, opened 2019-08-21
Ensure filter selections don't automatically reset

zephraph:
Resolves [PLATFORM-1687](https://artsyproduct.atlassian.net/browse/PLATFORM-1687)

When trying to select a filter from a shows fitlered artwork view, the picker would always reset when a selection was made. 

show-select-fail

This issue was caused because the picker was keying off of `id` from `counts` which was changed to `internalID`. The typing was insufficient and thus this wasn't caught. 

Actions I took:

1. Update `id` to `internalID` where appropriate
2. Update the types to ensure future changes would be caught


alloy:
Ace. I'll update this when I make my `internalID` to `value` change 👍

PR #1807, opened 2019-08-21
PURCHASE-1387 - Adds VAT included in price & replaces FilterArtworksConnection.count.internalID

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1387
- Adds `VAT included in price` to artworks when `true`
- Removes `FilterArtworksConnection.count.internalID` and replaces for `FilterArtworksConnection.count.value` @alloy 

Screen Shot 2019-08-21 at 5 07 21 PM

#skip_new_tests


src/lib/Scenes/Artwork/Components/CommercialPartnerInformation.tsx
```diff
@@ -34,6 +34,11 @@ export class CommercialPartnerInformation extends React.Component<Props> {
                 {artwork.shippingInfo}
               </Sans>
             )}
+            {artwork.priceIncludesTax && (
```
kierangillen: @lilyfromseattle lets circle back and add tests here once you merge your current PR adding the test file for this component.

src/lib/Containers/Gene.tsx
```diff
@@ -364,7 +364,6 @@ export default createPaginationContainer(
           aggregations {
             slice
             counts {
-              internalID
               name
```
alloy: This is not sufficient, it needs to be replaced by `value`. I'll add some typings to the code to make sure it is flagged as required.

```suggestion
               value
               name
```

PR #1811, opened 2019-08-22
Ensure keys are unique by removing reliance on slugs

zephraph:
This addresses platform-1690

When updating emission to use metaphysics v2, there were a few instances where the keys were transitioned over to point to be slugs.

Slugs aren't guaranteed to be unique, so the slug references have been changed to ids. 

I did notice there were some components which used an `index` as a key which I don't think is strictly a best practice. I did _not_ try to fix those in this PR.


src/lib/Scenes/Home/Components/ForYou/Components/FairsRail.tsx
```diff
@@ -93,6 +93,7 @@ export default createFragmentContainer(FairsRail, {
   fairs_module: graphql`
     fragment FairsRail_fairs_module on HomePageFairsModule {
       results {
+        internalID
```
alloy: It’s better to use `id` for this, as Relay will already implicitly select it under the hood anyways and thus potentially saving some overfetching.

alloy: Should the `internalID` selection still be added?

PR #1812, opened 2019-08-23
Deep zoom

ds300:
# Hello! 👋 

This PR introduces a preliminary 'Deep Zoom' overlay for the image carousel's full screen mode.

deep_zoom_1

We use the zooming/panning capabilities of a regular scroll view under the hood (was already present in ImageZoomView) and then layer over that a series of 'tiles', arranged in a pyramid which you see from a birds-eye view, with the highest-resolution tiles at the top and the lowest at the bottom.

There's a debug mode to help visualise this:

deep_zoom_debug

It's especially useful for tweaking the algorithms to decide which tiles to show and when to load them etc.

I added a lot of comments in the code to help understand what's going on.

There's a couple of caveats:

1. We were using `OpaqueImageView` to render the tiles. This worked beautifully on emission but right before submitting this PR I checked the feature in eigen and found that many images had tearing at the tile seams when zooming. That's obviously unacceptable so I switched to react-native's `Image` last-minute. We can try to resolve that as part of this PR review process but if it doesn't happen we can add a follow up task to try to switch back to `OpaqueImageView` later or maybe delete the extra features I introduced to `OpaqueImageView` if we decide to stick with `Image`.
2. Performance is acceptable, even on very large images, but still doesn't feel as snappy as I suspect it could do. I'd like to add a follow-up ticket to try to tweak the pyramid algorithms to be more selective about which tiles to show, to abort partial image requests that are no longer high-priority, to avoid render bottlenecks (https://twitter.com/djsheldrick/status/1164841287700439040), and maybe to guess where the user will look next and preload/prerender images there.
3. Unrelated tests are failing for some reason and I still need to figure out why.

#skipNewTests
#skip_new_tests


Pod/Classes/Core/RCTScrollView+EnclosingScrollView.m
```diff
@@ -121,32 +122,80 @@ - (RCTDescendantScrollEvent *)scrollEventForScrollView:(UIScrollView *)scrollVie
...
+
 }
 
 @end
 
+@interface RCTScrollView (RCTEnclosingScrollView)
+- (void)optOutOfParentScrollEvents;
+- (void)optOutOfAllScrollEvents;
+- (void)optInToAllScrollEvents;
+@end
```
alloy: From a purity perspective, we should probably be swizzling the methods we override and invoke the ‘original’ method as per OOP semantics.

In any case, I'm ok with being pragmatic, but the category name probably doesn’t match what it does well anymore. Perhaps we should just call the category `Artsy` or `Emission`?

Pod/Classes/Core/RCTScrollView+EnclosingScrollView.m
```diff
@@ -164,6 +213,9 @@ - (void)sendScrollEventWithName:(NSString *)eventName
                                                       scrollViewZoomScale:self.scrollView.zoomScale
                                                                  userData:userData
                                                             coalescingKey:coalescingKey];
+
+  // static int transforms = 0;
+  // NSLog(@"eventses %d", transforms++);
   [[self valueForKey:@"_eventDispatcher"] sendEvent:scrollEvent];
```
alloy: Just to be sure, does the `_eventDispatcher`, whatever type that may be, not already provide the ability to halt and event sending?

Pod/Classes/OpaqueImageViewComponent/AROpaqueImageView.h
```diff
@@ -4,4 +4,6 @@
 @property (nonatomic, strong, readwrite) NSURL *imageURL;
 @property (nonatomic, strong, readwrite) UIColor *placeholderBackgroundColor;
 @property (nonatomic, readwrite) BOOL *noAnimation;
+@property (nonatomic, readwrite) BOOL *failSilently;
+@property (nonatomic, readwrite) BOOL *highPriority;
```
alloy: Declaring a scalar property is not typically done as a pointer. I suspect things just happen to work because of `NULL` pointer checking? (I would also annotate them explicitly as `assign`.)

```suggestion
@property (nonatomic, assign, readwrite) BOOL highPriority;
```

(Same applies to the above.)

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/DeepZoom/DeepZoomLevel.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+  // This is a tiny perf thing, but to help the react reconciler do less work, we cache the shallow JSX
+  // for each tile
+  const tileCache: { [url: string]: JSX.Element } = useMemo(() => ({}), [])
+
+  // We use a 'fingerprint' of which rows and columns are currently being shown to decide whether to update
+  // the `tiles` state.
+  const lastFingerprint = useRef("")
+
+  // Here we calculate the transform for the whole level. It's a hariy one, pay attention
```
alloy: 💇‍♂

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/ImageZoomView.tsx
```diff
@@ -134,109 +135,197 @@ export const ImageZoomView: React.RefForwardingComponent<ImageZoomView, ImageZoo
...
+      const { pageX, pageY } = ev.nativeEvent
+      if (Math.ceil(zoomScale.current) >= maxZoomScale) {
+        ARScrollViewHelpers.smoothZoom(
+          findNodeHandle(scrollViewRef.current.getNode()),
+          -marginHorizontal,
+          -marginVertical,
+          width,
+          height
+        )
+        // resetZoom()
```
alloy: Still needed?

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/VerticalSwipeToDismiss.tsx
```diff
@@ -87,7 +87,7 @@ export const VerticalSwipeToDismiss: React.FC<{ onClose(): void }> = observer(({
       style={[{ opacity }]}
       snapToInterval={screenHeight}
       decelerationRate="fast"
-      scrollEventThrottle={16}
+      scrollEventThrottle={0.000000001}
```
alloy: Why is this needed?

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/useEventStream.ts
```diff
@@ -0,0 +1,36 @@
...
+  useEffect(
+    () => {
+      stream.__internal_emitter.addListener("event", listener)
+      return () => {
+        stream.__internal_emitter.removeListener("event", listener)
+      }
+    },
+    [listener]
+  )
+}
```
alloy: Nice 👌

src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
```diff
@@ -96,6 +91,17 @@ export const ImageCarouselFragmentContainer = createFragmentContainer(ImageCarou
...
+        Image {
+          TileSize
+          Url
+          Format
+          Size {
+            Width
+            Height
+          }
+        }
+      }
```
alloy: Errrr, what’s with the naming of these fields?? We should fix these to be `camelCase`.

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/DeepZoom/DeepZoomTile.tsx
```diff
@@ -0,0 +1,136 @@
...
+        top,
+        left,
+        width,
+        height,
+        opacity,
+      }}
+    >
+      <Image onLoad={onLoad} source={{ uri: url }} style={{ width, height }} />
+      {/*
+        // TODO: Figure out why we get tearing at tile seams when using this in eigen but not emission
```
alloy: I do notice some unrounded values here, which at best is just a guess, but something to look into https://github.com/artsy/emission/blob/a2168c4b30b6b495f678b7aad84ad853dc6c5d97/Pod/Classes/OpaqueImageViewComponent/AROpaqueImageView.m#L12-L13

ds300:
@alloy tracking heap size in Emission example app

```
Normal-sized image

Emission menu
113 MiB
Artwork page
135 MiB
Full screen carousel zoomed in
150 MiB
Close full screen
142 MiB
Back to full screen 
150 MiB
Scroll to next image and zoom in + pan around
158 MiB
Close full screen
148 MiB
Back into full screen and zoom + pan around on same two images
155 MiB

Huge image

Emission menu
113 MiB
Artwork page
133 MiB
Full screen carousel zoomed in
173 MiB
Panning around and zooming erratically for a while
jumps around between 160 MiB and 185 MiB (mostly < 175)
Close full screen
158 MiB
Back into full screen and zoom + pan around
Same as before: jumps around between 160 MiB and 185 MiB
```

So it feels quite stable, adding about 50 MiB max on top of the base artwork page for a huge image. What do you think? Also, do you think it would be worth repeating these tests with eigen?

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/VerticalSwipeToDismiss.tsx
ds300: unfortunately this throttles both native-side events and over-the-bridge events. We need to get updates for every single scroll event on the native side. I'd seen '16' mentioned in the docs as being the value to pass if you want to opt in to all scroll events, but after debugging I found out that it still drops quite a few events, which is bad news if you're animating things in unison with a scroll view. Setting the throttle to '0' has weird special semantics (only triggers one event at the start of a scroll), so the next best thing is a number very very close to 0.

One of the PRs I want to make to RN after this work is to make it so that this throttle doesn't necessarily apply to both events coming over the bridge and events on the native-side. Ain't nobody wants to throttle events with useNativeDriver 🤷‍♀

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/ImageZoomView.tsx
ds300: fixed

Emission.podspec
```diff
@@ -38,6 +38,9 @@ podspec = Pod::Spec.new do |s|
   # To ensure a consistent image cache between app/lib
   s.dependency 'SDWebImage', '>= 3.7.2', '< 4'
 
+  # For custom animations in DeepZoomOverlay
+  s.dependency 'INTUAnimationEngine'
```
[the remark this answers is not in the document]
ds300: There is now

Example/Emission.xcodeproj/xcshareddata/xcschemes/Emission.xcscheme
```diff
@@ -81,23 +81,6 @@
...
-            key = "SRCROOT"
-            value = "$(SRCROOT)"
-            isEnabled = "YES">
-         </EnvironmentVariable>
-         <EnvironmentVariable
-            key = "OS_ACTIVITY_MODE"
-            value = "disable"
-            isEnabled = "YES">
-         </EnvironmentVariable>
-      </EnvironmentVariables>
```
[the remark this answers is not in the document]
ds300: @ashfurrow helped me remove these to make `NSLog` work on my machine. I have no idea whether those changes should have remained local on my machine. What do you think?

Pod/Classes/Core/ARScrollViewHelpers.m
```diff
@@ -0,0 +1,155 @@
...
+RCT_EXPORT_METHOD(smoothZoom:(nonnull NSNumber *)tag x:(nonnull NSNumber *)x y:(nonnull NSNumber *)y w:(nonnull NSNumber *)w h:(nonnull NSNumber *)h)
+{
+  UIView* view = [[AREmission sharedInstance].bridge.uiManager viewForReactTag:tag];
+
+  if ([view isKindOfClass:RCTScrollView.class]) {
+    __weak RCTScrollView* weakScrollView = (RCTScrollView *) view;
+
+    // first disable scrolling so the user can't interrupt the animation
+    // TODO: (this doesn't seem to actually work, needs more investigation)
+    weakScrollView.scrollView.scrollEnabled = NO;
```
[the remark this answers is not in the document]
ds300: Ah yeah of course. Fixed in  [https://github.com/artsy/emission/pull/1812/commits/7d86d191ecfcc4dff5364c57f5ec258d2b0ee258](https://github.com/artsy/emission/pull/1812/commits/7d86d191ecfcc4dff5364c57f5ec258d2b0ee258)

Pod/Classes/Core/RCTScrollView+EnclosingScrollView.m
ds300: I'm not sure what you're getting at regarding the overridden methods, but I changed the category to Artsy.

Pod/Classes/Core/RCTScrollView+EnclosingScrollView.m
ds300: doesn't seem to

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/DeepZoom/DeepZoomTile.tsx
ds300: Hmm yeah, could be that scaleFactor is different between emission and eigen. Will look into it.

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/VerticalSwipeToDismiss.tsx
alloy: I see. Perhaps it could be a `SEND_ALL_EVENTS` constant?

alloy:
> So it feels quite stable, adding about 50 MiB max on top of the base artwork page for a huge image. What do you think? Also, do you think it would be worth repeating these tests with eigen?

Yeah that feels good. I won’t expect any differences, but yeah might be good to test in Eigen again once a new version of Emission is used in Eigen.

Example/Emission.xcodeproj/xcshareddata/xcschemes/Emission.xcscheme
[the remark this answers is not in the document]
alloy: Gotcha. `OS_ACTIVITY_MODE` should be related to that, it was only added to workaround some noise that I'm sure Apple fixed by now. The others, though, should not be related to logging, so perhaps leave those in.

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/DeepZoom/DeepZoomTile.tsx
ds300: nope scale factor is 2.0 on both so the mystery continues... 🤔

src/lib/Scenes/Artwork/Components/ImageCarousel/FullScreen/VerticalSwipeToDismiss.tsx
ds300: maybe, let's see!

src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
ds300: i aliased the fields to be lower-cased so we can fix this more easily later

Example/Emission.xcodeproj/xcshareddata/xcschemes/Emission.xcscheme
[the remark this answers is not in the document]
ds300: ok added those back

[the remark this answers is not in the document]
ds300: ok added those back

ds300:
@alloy this should be 🍏now. Feel free to squash and merge.

alloy:
Let's do this! :shipit:

PR #1820, opened 2019-08-28
PURCHASE-1434 - Small typography change

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1434

#trivial

### DIFF
--- a/src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
+++ b/src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
@@ -62,8 +62,8 @@ export class ArtworkExtraLinks extends React.Component<ArtworkExtraLinksProps> {
             Have a question?{" "}
             <Text style={{ textDecorationLine: "underline" }} onPress={() => this.handleReadOurAuctionFAQsTap()}>
               Read our auction FAQs
-            </Text>{" "}
-            or{" "}
+            </Text>
+            . or{" "}
             <Text style={{ textDecorationLine: "underline" }} onPress={() => this.handleAskASpecialistTap()}>
               ask a specialist
             </Text>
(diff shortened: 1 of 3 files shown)


src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
```diff
@@ -62,8 +62,8 @@ export class ArtworkExtraLinks extends React.Component<ArtworkExtraLinksProps> {
             Have a question?{" "}
             <Text style={{ textDecorationLine: "underline" }} onPress={() => this.handleReadOurAuctionFAQsTap()}>
               Read our auction FAQs
-            </Text>{" "}
-            or{" "}
+            </Text>
+            . or{" "}
```
ds300: shouldn't `or` become `Or` now?

src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
```diff
@@ -156,26 +156,26 @@ export class ArtworkTombstone extends React.Component<ArtworkTombstoneProps, Art
...
-              {artwork.partner && (
+              {!!artwork.partner && (
                 <Serif color="black100" size="3t" weight="semibold">
                   {artwork.partner.name}
                 </Serif>
               )}
-              {artwork.saleArtwork &&
-                artwork.saleArtwork.estimate && (
+              {!!artwork.saleArtwork &&
+                !!artwork.saleArtwork.estimate && (
```
ds300: What's up with all these double-bangs? I think the only time that casting to boolean actually matters is when the value might be `0`.

alloy: It's likely to avoid returning an empty string, which is illegal in React Native _outside_ of `<Text>…</Text>`.

src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
alloy: I would assume so, yes. Shall we make this change and get it in the beta?

```suggestion
            . Or{" "}
```

src/lib/Scenes/Artwork/Components/ArtworkTombstone.tsx
alloy: (This is probably going over-board a bit and I think that might not be the best precedent to set, but for now it's also not to the detriment of its functioning.)

PR #1822, opened 2019-08-29
[Core] Refactor screen dimensions to be dynamic.

alloy:
Updates screen dimensions on orientation change and emits an event.

#skip_new_tests


Pod/Classes/Core/RNDynamicScreenDimensions.h
```diff
@@ -0,0 +1,6 @@
+#import <React/RCTBridgeModule.h>
+#import <React/RCTEventEmitter.h>
+
+@interface RNDynamicScreenDimensions : RCTEventEmitter <RCTBridgeModule>
```
alloy: @ds300 We should not be using somebody else’s namespace. I.e. this needs to be prefixed with `AR`, but we can do that in follow-up.

Pod/Classes/Core/RNDynamicScreenDimensions.m
```diff
@@ -0,0 +1,112 @@
...
+#import <Emission/AREmission.h>
+#import <objc/runtime.h>
+
+@interface RNDynamicScreenDimensions ()
+@property (nonatomic, strong, readwrite) NSDictionary *currentDimensions;
+@property (nonatomic, readwrite) BOOL isBeingObserved;
+@end
+
+
+NSDictionary *getSafeAreaInsets()
```
alloy: Marking these functions as `static` will allow the compiler to inline them. Unsure what the default is.

src/setupJest.ts
```diff
@@ -162,3 +163,26 @@ if (process.env.ALLOW_CONSOLE_LOGS !== "true") {
...
+  }
+
+  return {
+    getCurrentScreenDimensions: () => screenDimensions,
+    ProvideScreenDimensions: ({ children }) => {
+      return React.createElement(React.Fragment, null, children)
+    },
+    useScreenDimensions: () => screenDimensions,
+  }
+})
```
alloy: Nice 👌

Pod/Classes/Core/RNDynamicScreenDimensions.m
```diff
@@ -0,0 +1,112 @@
...
+     name:UIDeviceOrientationDidChangeNotification
+     object:[UIDevice currentDevice]];
+  }
+  return self;
+}
+
+-(void)dealloc
+{
+  [[NSNotificationCenter defaultCenter] removeObserver:self];
+}
```
alloy: 👍

src/setupJest.ts
```diff
@@ -108,7 +108,7 @@ NativeModules.ARSwitchBoardModule = {
   presentMediaPreviewController: jest.fn(),
   presentArtworksSet: jest.fn(),
 }
-NativeModules.RNDynamicScreenDimensions = {
+NativeModules.ARDynamicScreenDimensions = {
```
ds300: we can actually delete this

ds300: done

PR #1825, opened 2019-08-29
PURCHASE-1396 - Adds Buy Now button to artwork with isBuyNowable

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1396
- Adds `BuyNowButton` to artworks in auction when `isBuyNowable: true` (https://zpl.io/a3R91m8)
- Small update to copy

Screen Shot 2019-08-29 at 3 16 01 PM

## In action

#skip_new_tests


src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
```diff
@@ -62,8 +62,8 @@ export class ArtworkExtraLinks extends React.Component<ArtworkExtraLinksProps> {
             Have a question?{" "}
             <Text style={{ textDecorationLine: "underline" }} onPress={() => this.handleReadOurAuctionFAQsTap()}>
               Read our auction FAQs
-            </Text>
-            . Or{" "}
+            </Text>{" "}
+            or{" "}
```
ds300: Accidental revert?

ds300: didn't we just change this the other day? is this an accidental revert?

kierangillen: Yes but it was incorrect before. This is a fix.

PR #1826, opened 2019-08-30
Carousel fixes

ds300:
Couple of small things

1. Screen dimensions updates seemed to be subject to a race condition on my iPad simulator at least. Calling `getScreenDimensions` synchronously with the `orientationChanged` callback was returning the pre-rotation values. So i moved the call onto the main thread right before the event is fired across the bridge.
2. Images whose maximum resolution is smaller than the viewport were being blown up to fill the screen in the full screen carousel. On iPad this could look *really* bad so I prevented it from happening.

#trivial
#skip_new_tests

### DIFF
--- a/Pod/Classes/Core/ARDynamicScreenDimensions.m
+++ b/Pod/Classes/Core/ARDynamicScreenDimensions.m
@@ -64,17 +64,18 @@ RCT_EXPORT_MODULE()
 
 - (void)orientationChanged:(NSNotification *)note
 {
-  self.currentDimensions = getScreenDimensions();
   __weak ARDynamicScreenDimensions *wself = self;
 
-  if (self.isBeingObserved) {
-    dispatch_async(dispatch_get_main_queue(), ^{
-      __strong ARDynamicScreenDimensions *sself = wself;
-      if (sself && sself.isBeingObserved) {
-        [sself sendEventWithName:@"change" body:sself.currentDimensions];
-      }
-    });
-  }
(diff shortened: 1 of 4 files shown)


Pod/Classes/Core/ARDynamicScreenDimensions.m
```diff
@@ -64,17 +64,18 @@ + (BOOL)requiresMainQueueSetup
...
+  dispatch_async(dispatch_get_main_queue(), ^{
+    __strong ARDynamicScreenDimensions *sself = wself;
+    if (!sself) {
+      return;
+    }
+    sself.currentDimensions = getScreenDimensions();
+    if (sself.isBeingObserved) {
+      [sself sendEventWithName:@"change" body:sself.currentDimensions];
+    }
+  });
```
alloy: The documentation for this notification indicates that the `NSNotification` object actually holds the value, which seems like a less brittle approach and perhaps less blocking of the main thread https://developer.apple.com/documentation/uikit/uideviceorientationdidchangenotification?language=objc

alloy: Or did you mean the values in the size struct?

ds300: yeah the sizes and safe area insets were wrong too.

alloy: Gotcha

PR #1827, opened 2019-08-30
Check for scrollview before scrolling

kierangillen:
- Fixes crash reported today on tabs

#skip_new_tests
#trivial

### DIFF
--- a/src/lib/Components/ScrollableTabBar.tsx
+++ b/src/lib/Components/ScrollableTabBar.tsx
@@ -103,7 +103,7 @@ export default class ScrollableTabBar extends React.Component<ScrollableTabBarPr
   }
 
   centerOnTab = (index: number) => {
-    if (!this.els[index] || this.scrollView) {
+    if (!this.els[index] || !this.scrollView) {
       return
     }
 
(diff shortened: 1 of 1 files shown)


src/lib/Components/ScrollableTabBar.tsx
```diff
@@ -109,7 +109,7 @@ export default class ScrollableTabBar extends React.Component<ScrollableTabBarPr
 
     const { width: screenWidth } = Dimensions.get("window")
     // If the scrollview doesn't need to scroll, don't try to scroll it!
-    if (this.scrollViewWidth < screenWidth) {
+    if (this.scrollViewWidth < screenWidth || !this.scrollView) {
```
alloy: Let's move it to the start of the function. Nothing needs to be done when we don’t have this ref.

src/lib/Components/ScrollableTabBar.tsx
```diff
@@ -103,7 +103,7 @@ export default class ScrollableTabBar extends React.Component<ScrollableTabBarPr
   }
 
   centerOnTab = (index: number) => {
-    if (!this.els[index]) {
+    if (!this.els[index] || this.scrollView) {
```
alloy: This is the inverse of what you want.

alloy:
Thanks!

#MergeOnGreen

alloy:
Btw, this is missing a ticket link and/or to the notion card.

PR #1831, opened 2019-09-03
[v2] Remove more root fields

alloy:
Part of https://artsyproduct.atlassian.net/browse/PLATFORM-1734

Relies on https://github.com/artsy/metaphysics/pull/1927


src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
```diff
@@ -113,8 +113,8 @@ export const MaxBidScreen = createRefetchContainer(
     `,
   },
   graphql`
-    query SelectMaxBidRefetchQuery($saleArtworkID: String!) {
-      sale_artwork: saleArtwork(id: $saleArtworkID) {
+    query SelectMaxBidRefetchQuery($saleArtworkNodeID: ID!) {
+      node(id: $saleArtworkNodeID) {
         ...SelectMaxBid_sale_artwork
```
alloy: @yuki24 @dleve123 Another FYI to triple-check in next QA. This is refetching the sale artwork node using the `node` root field with its global object ID.

src/lib/Scenes/Home/Components/Sales/Components/LotsByFollowedArtists.tsx
```diff
@@ -1,63 +1,43 @@
...
 
     return (
-      <ScrollView onScroll={isCloseToBottom(this.loadMore)} scrollEventThrottle={400}>
-        <SectionHeader title={title} />
-        <Container>
-          <GenericGrid artworks={artworks as any} />
-          {this.state.fetchingMoreData && <Spinner style={{ marginTop: 20 }} />}
-        </Container>
-      </ScrollView>
+      <Container>
```
kierangillen: Nit: May want to wrap in a `<Theme>` and use `<Box p={1}>` here instead of this container. Also the section header can get `pb={1}`

src/lib/Scenes/Home/Components/Sales/index.tsx
```diff
@@ -21,7 +21,7 @@ class Sales extends React.Component<Props, State> {
   }
 
   get data() {
-    const { sales } = this.props.query
+    const sales = this.props.query.salesConnection.edges.map(({ node }) => node)
```
kierangillen: Any chance `edges` or `salesConnection` here could be `null`?

kierangillen:
A couple minor things, but feel free to merge as is.

src/lib/Scenes/Home/Components/Sales/index.tsx
alloy: `salesConnection` can’t be, `edges` might but I expect things to then first blow up in MP.

src/lib/Scenes/Home/Components/Sales/Components/LotsByFollowedArtists.tsx
alloy: I didn’t really want to change much more than what was needed. Would these changes be better in any way other than consistency perhaps?

kierangillen: Cleaner, and if the design system ever changes it'll keep it consistent, better practice but not mandatory.

alloy: Good point re design system changes 👍

kierangillen:
#mergeongreen

PR #1832, opened 2019-09-03
Fixes Artist Header & Gene Header Follow buttons

kierangillen:
- Artist Header was keeping the `follow` state in local state, moves the state into the relay object so we have the correct immediate data.
- Also fixes Gene Header with the same fix
- Replaces uppercase tab names to capitalized to be inline with current design system.


src/lib/Components/Artist/Header.tsx
```diff
@@ -115,59 +103,92 @@ class Header extends React.Component<Props, State> {
...
+          },
+          optimisticResponse: {
+            followArtist: {
+              artist: {
+                id,
+                isFollowed: !isFollowed,
+              },
+            },
+          },
+          updater: store => {
```
sweir27: I don't _think_ you need this updater function since the relay store will update automatically and the shape of the mutation response looks standard.

kierangillen: Ok, looks like it's needed on the Gene mutation but not the Artist. Updated!

sweir27:
Looks good pending the merge conflict fix!

#mergeongreen

PR #1834, opened 2019-09-04
PURCHASE-1393 & PURCHASE-1449

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1393 - Fixes buyers premium link 
- https://artsyproduct.atlassian.net/browse/PURCHASE-1449 - Removes FAQ and ask a specialist if the artwork isn't acquirable. 

#trivial


src/lib/Scenes/Artwork/Components/AuctionPrice.tsx
```diff
@@ -11,8 +12,10 @@ interface AuctionPriceProps {
 
 export class AuctionPrice extends React.Component<AuctionPriceProps> {
   handleBuyersPremiumTap = () => {
-    // FIXME: This tap needs to be added
-    return null
+    const auctionSlug = this.props.artwork && this.props.artwork.sale && this.props.artwork.sale.slug
```
sweir27: Can we use the ID here instead of the slug? It shouldn't matter functionally but since this is for internal navigation that is invisible to users it seems odd to use slugs instead of actual IDs.

PR #1836, opened 2019-09-05
Render all carousel images

ds300:
This reverts #1835 😬 That PR was fine, it's just that to really squash this bug I want to try a different approach. This PR achieves the same effect of preloading all carousel images, but while also _rendering_ all images.

After talking to @alloy I became convinced that preloading alone would not solve the issue we have seen, and that the problem is almost certainly `FlatList` which has been allegedly quite buggy historically, especially around whether or not lazily-rendered items appear.

The evidence I have for this is that both times when this kind of bug has appeared and I have seen it with my own eyes, the loading 'placeholder' background was not being show. It was just white. If this issue were caused by not prefetching images, we would expect to see a grey box with a loading spinner instead of a pure white background.

So it seems that `FlatList` is fine with rendering N items up-front, but occasionally chokes when pulling in more items as the user scrolls. It will be impossible for me to debug `FlatList` when I can't reproduce the issue locally and other people can't seem to reproduce it on-demand, so instead this PR introduces a workaround.

As already mentioned, that workaround is simply rendering all images up front. We have the luxury of being able to do that since the images are quite small, and we could get up to about 30 thumbnails before we need to even think about memory usage.

Network usage is a different thing. Here we set the first image to be a 'high pritority' download, which pushes it to the front of the `SDWebImage` download queue. That should help with the artwork page feeling snappy, but not as much as if we could force the images to download sequentially.

Another minor issue is that image downloads can't be cancelled with our current implementation of `OpaqueImageView` so if the user is on a bad connection and is navigating between pages faster than the thumbnails can load, they'll build up a queue of image downloads which will further hurt their network performance. This was an issue with the previous approach too. I'm not sure if `SDWebImage` supports cancelling downloads but if it does we should probably update `OpaqueImageView` to take advantage of it.

#skip_new_tests 
#trivial

### DIFF
--- a/Pod/Classes/OpaqueImageViewComponent/AROpaqueImageView.m
+++ b/Pod/Classes/OpaqueImageViewComponent/AROpaqueImageView.m
@@ -62,7 +62,7 @@ LoadImage(UIImage *image, CGSize destinationSize, CGFloat scaleFactor, UIColor *
 {
   // This will cancel an in-flight download operation, if one exists.
   self.imageURL = nil;
-  
+
   if (self.noAnimation) {
     [super setImage:image];
   } else {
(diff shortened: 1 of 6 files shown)


alloy:
Makes sense, this looks good to me and good call on the high priority 👍 

As for your question, `SDWebImage` can cancel downloads.

PR #1843, opened 2019-09-09
PURCHASE-1397 - Artwork view pull to refresh

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1397
- Adds pull to refresh to artwork view


sweir27:
This implementation looks solid!

The gif makes it look like the refresh is almost instantaneous 😮. Is that just an artifact of the animation? (i.e. have you verified that it's actually pulling in new data?).

Other than that ☝️ question this looks good to go!

kierangillen:
@sweir27 Yes I verified its running a new query which is outputted in the console.

PR #1845, opened 2019-09-10
Report all data load errors when not in tests

ds300:
Follow up from artsy/metaphysics#1936

The above issue was tricky to debug because the error was being swallowed by `renderWithLoadProgress`.

`renderWithLoadProgress` does invoke `console.error` but only for errors with a specific shape. It turned out that our error yesterday didn't match that shape. So I moved the `console.error` invocation to make it more inclusive for _all_ errors 🌈 

#trivial

### DIFF
--- a/src/lib/utils/renderWithLoadProgress.tsx
+++ b/src/lib/utils/renderWithLoadProgress.tsx
@@ -18,15 +18,18 @@ export default function<T extends RelayContainer<any>>(
       // In tests we want errors to clearly bubble up.
       if (typeof jest !== "undefined") {
         throw error
+      } else {
+        console.error(error)
       }
+
       const networkError = error as any
       if (networkError.response && networkError.response._bodyInit) {
         let data = networkError.response._bodyInit || "{}"
         try {
           data = JSON.parse(data)
+          console.error("Error data", data)
           // tslint:disable-next-line:no-empty
         } catch (e) {}
-        console.error(`Metaphysics Error: ${error.message}\n`, data)
       }
 
       if (retrying) {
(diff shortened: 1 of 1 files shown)


src/lib/utils/renderWithLoadProgress.tsx
```diff
@@ -18,15 +18,18 @@ export default function<T extends RelayContainer<any>>(
...
+
       const networkError = error as any
       if (networkError.response && networkError.response._bodyInit) {
         let data = networkError.response._bodyInit || "{}"
         try {
           data = JSON.parse(data)
+          console.error("Error data", data)
           // tslint:disable-next-line:no-empty
         } catch (e) {}
-        console.error(`Metaphysics Error: ${error.message}\n`, data)
```
alloy: I’m not sure I understand, wouldn’t the previous implementation always print the error data unless `JSON.parse()` throws an error (in which case printing out the data on line 30 would also not occur)?

ds300: The previous version would only print the error if it matched the shape `{response: {_bodyInit}}`. Check the `if` condition above.

alloy: Hmm, I’m not following, both these locations are inside the same `if` body.

ds300: Yeah I kept the `console.error` for this extra data. But now there's also a console.error outside of this `if` block, up above, which gets called as long as this code is not being executed at test time.

alloy: Oh yeah, I saw that one and that’s great, but that’s not the change I was responding to here 😄

alloy: Ok, so it was just confusion. Thanks for clarifying :+1:

alloy:
#MergeOnGreen

PR #1849, opened 2019-09-10
PURCHASE-1464 [BidFlow] Fixes bug when clicking Next between fields in Billing Address form

sweir27:
Currently, clicking "Next" to navigate to the next field in the bid flow's billing address form causes the app to crash with the error `RCTFatalException: Unhandled JS Exception: undefined is not an object (evaluating 't.addressLine1.root.focus')`

Fixed it by removing the reference to `root`. This component hasn't been touched meaningfully in a while, so I'm not sure what made this break all of a sudden.

next-bid-flow

#trivial


ds300:
weird, it's probably a leftover from the styled-components v4 upgrade. #mergeOnGreen 👍

PR #1855, opened 2019-09-11
[BNMO] Uses metaphysics string to display VAT information

sweir27:
This is the emission-side of https://github.com/artsy/reaction/pull/2800

image

I also updated the logic slightly around when to show shipping and tax information based on what we do on desktop (only show it for BNMO works).

#trivial


kierangillen:
#mergeongreen

PR #1859, opened 2019-09-12
Restricts git diff range on CI skip check

ashfurrow:
I noticed that #1837 (which has only JS changes) was running native CI ([in this build](https://circleci.com/gh/artsy/emission/2823)). We have a yarn script to skip native CI on pull requests that have no native code changes, but it wasn't working. What was happening was, if native code got changed on master, that would "count" as a changed native file on every PR (unless they rebased). This change restricts the git diff to the most common ancestor commit of the PR's branch (ie: the commit it branched off of master from). #trivial

### DIFF
--- a/scripts/ci-skip-native-if-possible.js
+++ b/scripts/ci-skip-native-if-possible.js
@@ -29,7 +29,7 @@ if (branch === "master") {
 }
 
 /** @type {String[]} */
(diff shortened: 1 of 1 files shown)


kierangillen:
Wow, nice catch and fix!

PR #1861, opened 2019-09-13
PURCHASE-1484 prevent crashes when staging image data is corrupt

ds300:
@kierangillen saw some crashes in the app when staging was flaky that were related to certain artworks not having full image data. Here we add a fail-safe so the app will not crash if that happens again, it will just not render any images.

https://artsyproduct.atlassian.net/browse/PURCHASE-1484

#trivial


kierangillen:
#mergeongreen

PR #1863, opened 2019-09-16
PURCHASE-1488 increase image carousel base resolution

ds300:
#trivial

https://artsyproduct.atlassian.net/browse/PURCHASE-1488

### DIFF
--- a/src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
+++ b/src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
@@ -44,8 +44,8 @@ export const ImageCarousel = observer((props: ImageCarouselProps) => {
             height,
             url: createGeminiUrl({
               imageURL: image.image_url.replace(":version", "normalized"),
-              width,
-              height,
+              width: width * 1.5,
+              height: height * 1.5,
             }),
             deepZoom: image.deepZoom,
           }
--- a/src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
+++ b/src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
@@ -17,6 +17,8 @@ export interface ImageCarouselProps {
   images: ImageCarousel_images
 }
 
+const THUMBNAIL_PIXEL_DENSITY = 1.5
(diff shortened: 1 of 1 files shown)


src/lib/Scenes/Artwork/Components/ImageCarousel/ImageCarousel.tsx
```diff
@@ -44,8 +44,8 @@ export const ImageCarousel = observer((props: ImageCarouselProps) => {
             height,
             url: createGeminiUrl({
               imageURL: image.image_url.replace(":version", "normalized"),
-              width,
-              height,
+              width: width * 1.5,
```
sweir27: This seems slightly arbitrary-- maybe we could extract the `1.5` into a constant?

(Also, how did you arrive at this number?)

ds300: I just scaled the resolution up until the increase in resolution/quality when the lightbox image fades in is barely noticeable. I assume that that fade-in is what Sam was noticing and asking us to prevent the user from noticing.

sweir27: I only notice the fade-in when I click to view the image full screen. Should it be happening on the base artwork view as well? I'm worried that this approximation might not be optimal for all image shapes, but by adding this + fading into the slightly larger version, we might be good.

ds300: > I only notice the fade-in when I click to view the image full screen. 

yeah that's the fade-in I'm talking about. when you click to enter full screen it goes like:

1. 'zoom' the carousel thumbnail into the center of the screen. If the image is portrait, then it will also get larger during this process, making it's resolution even lower relative to the physical screen.
2. the deep zoom overlay fades in above the thumbnail, showing a higher-quality version of the same image.

The disparity between the thumbnail resolution and the deep zoom overlay resolution is much smaller now, since we increased the thumbnail resolution, so that deep zoom overlay fade-in animation should be less jarring, almost imperceptible.

if you're still seeing it on some artworks then we might need to increase the thumbnail resolution some more.

sweir27:
#mergeongreen

PR #1865, opened 2019-09-17
Use screen pixel density rather than constant

ds300:
As @alloy pointed out on slack, some iOS devices have a pixel ratio of 1:3 so we should ask the user's device how much to scale the image by.

#trivial


alloy:
It's a little unclear to me why a bunch of the work that's encapsulated in `OpaqueImageview` needs to be redone here (is there opportunity to clean this up in the future?), but functionally this looks good to me 👍 so #MergeOnGreen.

ds300:
`OpaqueImageView` previously exclusively derived image URLs directly from rendered size.

During the shared element transition for the full screen image carousel we need to render the same image at slightly different sizes. The second one needs to be shown instantly so re-downloading a slightly different-sized version of the same image was not an option and we had to ensure that the URL was stable so we'd get a cache hit.

alloy:
Gotcha, that makes sense. I do think that `OpaqueImageView` could easily be made to do that, though.

PR #1866, opened 2019-09-17
[Home] Stop scroll-to-top behaviour on Artists/Auctions tabs.

alloy:
Fixes https://artsyproduct.atlassian.net/browse/PLATFORM-1781

### DIFF
--- a/src/lib/Scenes/Home/index.tsx
+++ b/src/lib/Scenes/Home/index.tsx
@@ -17,6 +17,8 @@ import { ForYouRenderer, WorksForYouRenderer } from "lib/relay/QueryRenderers"
 import { SalesRenderer } from "lib/Scenes/Home/Components/Sales/Relay/SalesRenderer"
 import renderWithLoadProgress from "lib/utils/renderWithLoadProgress"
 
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/Home/index.tsx
```diff
@@ -123,21 +134,15 @@ export default class Home extends React.Component<Props, State> {
       wider letter-spacing. Going forward, this would ideally be dealt with through letter indentation. */}
           <Tab tabLabel=" Artists">
             <WorksForYouRenderer
-              render={renderWithLoadProgress(query => (
-                <WorksForYou query={query as any} />
-              ))}
```
alloy: When these functional components were defined inline, they would be new instances on each render. Thus, changing a prop on the query renderer would lead to this entire tree re-rendering, as React has no way of knowing if this is the same component (technically it wasn’t).

alloy:
Self-merging in the interest of cutting a new build asap.

#MergeOnGreen

src/lib/Scenes/Home/index.tsx
ashfurrow: Gotcha, I was thinking the bug was caused by something like this, but couldn't get it to stick. Thanks Eloy!

PR #1872, opened 2019-09-17
Adjust deep zoom levels for pixel ratio changes

ds300:
in #1863 we updated the image carousel thumbnail to use the device's pixel ratio to decide image resolution, but forgot to update the deep zoom overlay to do the same. This PR basically makes it so that if you zoom in (either very slowly or with perfect network) you can't tell when one layer ends and the other begins. Whereas before there would be noticeable points while zooming when the image would become more detailed.

#trivial


alloy:
LGTM 👍

PR #1873, opened 2019-09-17
Fixes empty space in Artwork view

kierangillen:
- We had not updated the logic in the parent component when to show `ExtraLinks` now that we're no longer showing if the artwork only has `isInquireable`

#trivial

## Before

Screen Shot 2019-09-17 at 12 32 09 PM

## After

Screen Shot 2019-09-17 at 12 31 54 PM


sweir27:
#mergeongreen

PR #1876, opened 2019-09-17
[Artwork View] - Fix pull to refresh

kierangillen:
- Fixes bug which was only allowing pull to refresh once

#trivial


src/lib/Scenes/Artwork/Artwork.tsx
```diff
@@ -107,9 +107,17 @@ export class Artwork extends React.Component<Props, State> {
...
 
-    this.props.relay.refetch({ id: this.props.artwork.id, fetchContent: true }, null, () => {
-      this.setState({ refreshing: false })
-    })
+    this.props.relay.refetch(
+      { id: this.props.artwork.id },
+      {},
+      error => {
+        if (error) {
+          console.error("Artwork.tsx refretch query: ", error.message)
```
sweir27: `refretch` :)

sweir27:
Out of curiosity, which of the above changes was the root cause?

ashfurrow:
Ah, I bet it was the missing `{ force: true }`.

PR #1883, opened 2019-09-18
[package] Upgrade to relay v6

alloy:
Closes https://artsyproduct.atlassian.net/browse/PLATFORM-1797

No complications found, but holding off with merging until the current artwork work is out. /cc @sweir27 

- [x] Upgrades to relay v6
- [x] Removes unused typings
- [x] Fixes long standing bug with type emission around unions
- [x] Applies codemod to add `@raw_response_type` to all test operations
- [x] Fix all type errors and tests (see ac360d216f937d475827f4ac957d8d0318dcd440 for an example)

#skip_new_tests"


sweir27:
@alloy there's a trivial conflict here, but other than that is this good to go?

alloy:
The tests/types need to be finished first, it would be great for others to pick that up, though.

damassi:
@alloy - tests and types have been fixed per slack discussion.

src/lib/Scenes/Fair/Screens/__tests__/FairArtists-tests.tsx
```diff
@@ -20,7 +19,7 @@ describe("FairArtists", () => {
       `,
       mockData: {
         fair: fairFixture,
-      } as FairArtistsTestsQueryRawResponse,
+      },
```
alloy: In cases like these the idea was to actually fix the test, but that can be followed up on. I’m adding these typings back but comment them out for others to easily fix these disabled tests, if needed. /cc @ashfurrow

src/lib/Scenes/Map/__tests__/GlobalMap-tests.tsx
```diff
@@ -38,7 +37,7 @@ xdescribe("GlobalMap", () => {
         viewer: {
           city: CityFixture,
         },
-      } as GlobalMapTestsQueryRawResponse,
+      },
```
alloy: Ditto

src/lib/Scenes/Fair/Screens/__tests__/FairArtists-tests.tsx
alloy: (There’s a bunch more disabled tests related to fixture data not being correct. I’m not going to list them all out, you can find them as being disabled.)

alloy:
Alright, let’s do this :+1:

src/lib/Scenes/Fair/Screens/__tests__/FairArtists-tests.tsx
damassi: Ah, I didn't see the `xit` on the test; i was running `yarn jest` and then fixing the returned failures.

PR #1885, opened 2019-09-19
PURCHASE-1496: Auction actions and info still render when work has sold via buy now

lilyfromseattle:
# [PURCHASE-1496] 
Before this change, after buying the artwork, the BUY NOW button correctly goes away, but the artwork should show as SOLD and it's still allowing you to bid. Also, auction links and timer still render.

After this change, auction actions and info will be hidden if availability is "sold".

__Before:__
Screen Shot 2019-09-19 at 4 08 52 PM

__After:__
Screen Shot 2019-09-19 at 4 08 11 PM

#trivial

[PURCHASE-1496]: https://artsyproduct.atlassian.net/browse/PURCHASE-1496


src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
```diff
@@ -50,10 +50,12 @@ export class ArtworkExtraLinks extends React.Component<ArtworkExtraLinksProps> {
 
   renderFAQAndSpecialist = () => {
     const {
-      artwork: { isAcquireable, isInAuction, sale },
+      artwork: { isAcquireable, isInAuction, sale, availability },
     } = this.props
 
-    if (isInAuction && sale && !sale.isClosed) {
+    const hasNotSold = availability && availability !== "sold"
```
sweir27: Here (and elsewhere in these changes), I would use the metaphysics property [`isForSale` ](https://github.com/artsy/metaphysics/blob/master/src/schema/v2/artwork/index.ts#L360).

So line 58 could just be `if (isInAuction && sale && !sale.isClosed && isForSale)` and you can avoid this extra logic.

`availability` isn't binary, but I think that ☝️ is more correct anyways.

lilyfromseattle:
@sweir27 Justin and I couldn't figure out what was causing these weird typescript errors. Do you mind taking a look?

ashfurrow:
@lilyfromseattle The problem is because we've removed the `...CommercialButtons_artwork` from `CommercialInformation.tsx`. In order for Relay to forward on the right data to the `CommercialButtons` component, we need to spread over its fragment. I added it back in and it worked locally, so I committed and pushed the change 👍

PR #1886, opened 2019-09-20
PURCHASE-1509: Fix bid flow bug introduced while cleaning up fields

sweir27:
#trivial


src/lib/Components/Bidding/Screens/SelectMaxBid.tsx
```diff
@@ -101,6 +101,7 @@ export const MaxBidScreen = createRefetchContainer(
         id
         increments(useMyMaxBid: true) {
           display
+          cents # Used on the ConfirmBid screen
```
ashfurrow: This change makes sense, but maybe we could move it into the `ConfirmBid_sale_artwork` fragment instead? Or, more precisely, the `BidResult_sale_artwork` fragment (which `ConfirmBid_sale_artwork` spreads over)? This would keep the data we're trying to fetch close to the place where we use it. 

If that's too much of a refactor right now, we can ship this as-is (especially with this comment, thank you 🙏).

sweir27: I recall there being a reason that we need to pass the increments in this way (although can't recall right now 😢). The flow is pretty tricky and I don't _think_ this was a simple oversight, so I'd rather not do the refactor right now if that's OK!

ashfurrow: Sounds good, let's merge 👍

PR #1891, opened 2019-09-24
Track more auction events

ashkan18:
# Change
https://artsyproduct.atlassian.net/browse/PURCHASE-1486
Added `@track` around few auction related actions based on ☝️ 

# Refactor?
`BidButton` logic was little complicated and hard to read, ended up consolidating some checks in one place and also tried to break `render` to call two other renders in `if` statement.

Let me know if there is a better way to make this more readable and if my changes don't help.

#skip_new_tests (for existing schema.ts file)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 ### 1.17.2
 
 - Refetches Artwork page query on re-appear - ash & purchase team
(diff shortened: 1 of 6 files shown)


ashkan18:
thanks @ds300 ! going to merge.

PR #1893, opened 2019-09-25
[PURCHASE-1519] Fixes issue where the Artwork page would show a refresh animation on re-appear

ashfurrow:
Tracked here: https://www.notion.so/artsy/When-viewing-a-modal-i-e-the-Attribution-Class-one-going-back-to-the-artwork-page-triggers-a-ref-295d8c5c0d204f748ee1da194fa4c607


sweir27:
Looks good! #mergeongreen

PR #1895, opened 2019-10-01
PURCHASE-1541 - Fixes extralinks

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1541
- Ask a specialist now links to `specialist@`
- Adds link to Auction FAQ

#trivial

### DIFF
--- a/src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
+++ b/src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
@@ -22,12 +22,12 @@ export class ArtworkExtraLinks extends React.Component<ArtworkExtraLinksProps> {
     action_type: Schema.ActionTypes.Tap,
     context_module: Schema.ContextModules.ArtworkExtraLinks,
   })
-  handleAskASpecialistTap() {
+  handleAskASpecialistTap(emailAddress) {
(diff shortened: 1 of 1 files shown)


src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
```diff
@@ -27,7 +27,7 @@ export class ArtworkExtraLinks extends React.Component<ArtworkExtraLinksProps> {
     const mailtoSubject = `Inquiry on ${artwork.title}`.concat(
       artwork.artist && artwork.artist.name ? ` by ${artwork.artist.name}` : ""
     )
-    Linking.openURL(`mailto:orders@artsy.net?subject=${mailtoSubject}`)
+    Linking.openURL(`mailto:specialist@artsy.net?subject=${mailtoSubject}`)
```
sweir27: We need to add some logic here--

If the artwork is available for BNMO, this should go to `orders@` but if it's in an auction it goes to `specialist@`.

kierangillen: Ah I see, I thought that was maybe the case but it wasn't in the ticket :) Will update now.

PR #1897, opened 2019-10-01
PURCHASE-1540 -  Uses em dash between text in ReadMore

kierangillen:
- Uses em dash between text instead in unexpanded text in `ReadMore` component
- https://artsyproduct.atlassian.net/browse/PURCHASE-1540

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### Master
 
 - Fixes crash with missing artist images - ash
+- Uses em dash between text in ReadMore component instead of line break - kierangillen
 
 ### 1.17.6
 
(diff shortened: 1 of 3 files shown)


ds300:
Yeah that's not ideal since the spacing of two line breaks is not the same as the spacing between two block-rendered Serif components.

When Sarah and i worked on this we specifically added a `<Text>` wrapper around the truncated version of the text to avoid line breaks. The correct way to add line breaks back in would be to remove the `<Text>` wrapper I think. But I wonder if we should add an inline separator at the same place you've done it here, instead of adding line breaks. That would avoid the truncated text getting too tall and visually distracting? 

Something like an em dash?

image

ds300:
or an en dash

image

or a middle dot

image

samchieng:
I'm on board for the em dash!

sweir27:
@kierangillen couple of test failures, otherwise looks 👍 !

kierangillen:
@sweir27 Updated the logic a bit, fixed the tests and added a new test for this behavior. Should be good to go now.

PR #1900, opened 2019-10-02
Remove mobx

ds300:
To explain what I've replaced mobx with it's good to explain why I used mobx in the first place.

The image carousel needs global state. Many different parts of the carousel code need to know about things like what stage of the full-screen lifecycle it's in, what the current image index is, etc. The only ways that react provides out-of-the-box to pass global state into child components are as follows:

- Prop drilling
  This is when the global state is stored in the grandest of parents and passed down through children via props.
- Context.
  This is when you use `React.createContext` and `React.useContext` to avoid the need to prop drill.

Prop drilling has bad perf characteristics. It can increase reconciliation times to the point where the UI feels sluggish. (reconciliation is the name React uses for calculating + applying changes from (re-)renders)

Context is a little better because then only the components that use the context value actually get updated.

There are still problems with Context, however. The first is that it's not ergonomic to have a different context for every single state value, so it's tempting to cram all state together into a single context value, and then that means that we still get redundant updates.

The bigger problem for us, however, is that when passing values via context you pass the raw value itself, which leads to a situation where 'stale closures' are possible. Stale closures are people's main gripe with react hooks. The react-sanctioned way around them is to use `useMemo` and `useCallback` with carefully-managed dependency arrays. But if the value that they close over is actually changing over time this is still not ideal because it leads to re-renders. Those re-renders can be avoided if you use a 'wrapper' for the state values. And we need those re-renders caused by stale closures to be avoided because they can also lead to noticeable sluggishness.

MobX is one way to get a state value that is both wrapped (and therefore never stale in closures) and also reactive (so that components re-render at the appropriate time). This pair of properties completely solves our perf issues. But using mobx gave us other issues, as we've seen.

I could not find any other popular react state management libraries which solve both of these problems, so I wrote the smallest and dumbest one I could think of.

It works like this:

```ts
const [stateWrapper, setStateValue] = useGlobalState("value")

console.log(stateWrapper.current) // => "value"

setStateValue("cheese")

console.log(stateWrapper.current) // => "cheese"
```

By default, components which access the `.current` property do not re-render when the value changes. That can be enabled by doing the following in a component

```ts
stateWrapper.useUpdates()
```
 
then whenever `setStateValue` is called with a value which is different from the current one, the component will re-render.

Meanwhile, in closures, you can access `stateWrapper.current` at any time and get a fresh value.

I chose `.current` to mirror the ref api.

#skip_new_tests


src/lib/Scenes/Artwork/Components/ImageCarousel/useGlobalState.ts
```diff
@@ -0,0 +1,56 @@
...
+  useEffect(() => {
+    isMounted.current = true
+    return () => {
+      isMounted.current = false
+    }
+  })
+  return useMemo(() => () => isMounted.current && setEpoch(x => x + 1), [])
+}
+
+export class GlobalState<T> {
```
ashfurrow: We could add a short description here, maybe even a link back to this PR, for context.

ds300: great idea! done

PR #1902, opened 2019-10-02
Follow up on original attempt to add auction analytics

ashkan18:
# Problem
After testing realized for Bid and Max Bid we don't see any analytics when clicking bid button.

# Cause
`@track` method was added to two methods that were not used anywhere.

# Solution
Move track method to `handleRegisterBid` method and conditionally set the proper `action_name`.

### DIFF
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/BidButton.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/BidButton.tsx
@@ -16,7 +16,7 @@ export interface BidButtonProps {
 const registrationWasAttempted = sale => !!sale.registrationStatus
 const isRegisteredToBid = sale => registrationWasAttempted(sale) && sale.registrationStatus.qualifiedForBidding
 const watchOnly = sale => sale.isRegistrationClosed && !isRegisteredToBid(sale)
-const getMyLotStanding = artwork => artwork.myLotStanding && artwork.myLotStanding[0]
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/Artwork/Components/CommercialButtons/BidButton.tsx
```diff
@@ -16,26 +16,11 @@ export interface BidButtonProps {
 const registrationWasAttempted = sale => !!sale.registrationStatus
 const isRegisteredToBid = sale => registrationWasAttempted(sale) && sale.registrationStatus.qualifiedForBidding
 const watchOnly = sale => sale.isRegistrationClosed && !isRegisteredToBid(sale)
+const getMyLotStanding = artwork => artwork.myLotStanding && artwork.myLotStanding[0]
```
sweir27: Minor style thing, but if you make these class methods you can access props directly and avoid having to pass the objects in each time. (That _feels_ clearer to me but I'm not an expert)

kierangillen: I believe this could also error if myLotStanding is an empty array. May want to check do this: `artwork.myLotStanding && artwork.myLotStanding.length && artwork.myLotStanding[0]`

ashkan18: @sweir27 i kind of thought not using props makes it little more obvious what this method gets and how those attributes are defined, but if you feel strong i can move them to class.

PR #1906, opened 2019-10-03
PURCHASE-1543 - Artwork Auction refresh on state change

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1543
- This PR moves the auction status into local state, this allows the component to refresh views based on status changes without querying metaphysics.


src/lib/Scenes/Artwork/Components/ArtworkExtraLinks.tsx
```diff
@@ -61,10 +63,10 @@ export class ArtworkExtraLinks extends React.Component<ArtworkExtraLinksProps> {
   renderFAQAndSpecialist = () => {
     const {
       artwork: { isAcquireable, isOfferable, isInAuction, sale, isForSale },
+      auctionState,
     } = this.props
 
-    if (isInAuction && sale && !sale.isClosed && isForSale) {
-      // FIXME: Verify logic when to show Auction "ask a specialist"
+    if (isInAuction && sale && !sale.isClosed && isForSale && auctionState !== "hasEnded") {
```
sweir27: I think we can remove the check for `sale.isClosed` here (since `auctionState` should handle that)

src/lib/Scenes/Artwork/Components/CommercialInformation.tsx
```diff
@@ -14,19 +15,26 @@ interface CommercialInformationProps {
   artwork: CommercialInformation_artwork
 }
 
+export type AuctionState = "hasStarted" | "isLive" | "hasEnded" | "isPreview" | null
```
sweir27: We'd talked before about not allowing this to be null and instead making sure we're correctly setting the initial state. I _think_ that would be preferable since `null` doesn't have any meaning in this case.

src/lib/Scenes/Artwork/Components/CommercialInformation.tsx
```diff
@@ -71,12 +80,73 @@ export class CommercialInformation extends React.Component<CommercialInformation
...
+        this.setState({
+          auctionState: "isPreview",
+        })
+      } else if (DateTime.local() > DateTime.fromISO(sale.startAt)) {
+        this.setState({
+          auctionState: "hasStarted",
+        })
+      }
+
+      // Set intervals to check for state changes
```
sweir27: Just noting that the auction timer employs a similar functionality w.r.t. keeping track of time and changing state accordingly.

Would it be worth trying to combine those functionalities or do you feel that they are separate enough? Under the hood the auction should be in the same state either way, so conceptually it seems they could be shared. But from a code organization perspective this might be better?

Curious what you think.

kierangillen: I was thinking this too, I just removed the interval from the countdown timer component.

src/lib/Scenes/Artwork/Components/__tests__/CommercialInformation-tests.tsx
```diff
@@ -38,7 +38,7 @@ describe("CommercialInformation", () => {
     }
     const component = mount(
       <Theme>
-        <CommercialInformation artwork={workInEndedAuction} />
+        <CommercialInformation artwork={workInEndedAuction as any} />
```
kierangillen: Ugh dont love these `as any` but this got hairy, we should ticket to clean up these mock artworks

src/lib/Scenes/Artwork/Components/AuctionCountDownTimer.tsx
```diff
@@ -60,7 +24,7 @@ export class AuctionCountDownTimer extends React.Component<AuctionCountDownTimer
         labelWithoutTimeRemaining={sale.formattedStartDateTime}
         labelFontSize="2"
         timerFontSize="4t"
-        countdownEnd={auctionStatus === "hasNotStarted" ? sale.startAt : sale.endAt}
+        countdownEnd={auctionState === "isPreview" ? sale.startAt : sale.endAt}
```
sweir27: Looking at this now... how does this deal with sales that haven't gone live yet? In that case the end time should be the `sale.liveStartAt`.

kierangillen: Created this ticket for this task: https://artsyproduct.atlassian.net/browse/PURCHASE-1584

PR #1910, opened 2019-10-08
ME-17 Send RetryErrorBoundary errors to sentry

ds300:
https://artsyproduct.atlassian.net/browse/ME-17

#trivial 
#skip_new_tests

There's a separate method for capturing `Error`s called `Sentry.captureException` but I found that the stack trace disappeared when using that method 🤔 So I decided to just capture the stack trace as a string. We'll need to figure out why the stack trace is disappearing if we ever want to set up symbolication for eigen (that's where you tell sentry about the source maps for your js bundle so it can replace bundled JS line numbers with source TS line numbers)

### DIFF
--- a/src/lib/Components/RetryErrorBoundary.tsx
+++ b/src/lib/Components/RetryErrorBoundary.tsx
@@ -1,4 +1,5 @@
 import React from "react"
+import { Sentry } from "react-native-sentry"
 import LoadFailureView from "./LoadFailureView"
 
 enum ErrorState {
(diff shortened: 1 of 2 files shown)


ashfurrow:
#mergeongreen

ds300:
Thanks for fixing the mock :sweat: :heart:

ashfurrow:
No worries! It was only because I've made the same mistake before that I could fix it quickly 😅

PR #1911, opened 2019-10-08
ME-39 - Adds spinner to home page tabs when loading more

kierangillen:
#trivial

- https://artsyproduct.atlassian.net/browse/ME-39


kierangillen:
@ashfurrow I needed to hook into the render cycle and rerender the view when new content was loaded so I had to opt the loading into the local state and not use the relay `isLoading()` function.

PR #1912, opened 2019-10-09
PURCHASE-1584: Update AuctionCountDownTimer to handle liveStartAt

lilyfromseattle:
# [PURCHASE-1584] 

The `AuctionCountDownTimer` component currently does not handle artworks with auction `liveStartAt` timestamps.

The component should also be able to handle a single date even if another isn't available. E.g. an artwork with a `startAt` but not an `endAt` date, or vice versa.

Instead of relying on `formattedStartDateTime` I made a method that returns the strings based on auctionState, so that it will update when the state changes. I also added some date formatting helper functions.

Screen Shot 2019-10-09 at 5 39 29 PM

#skip_new_tests
[PURCHASE-1584]: https://artsyproduct.atlassian.net/browse/PURCHASE-1584


src/lib/Scenes/Artwork/Components/__tests__/AuctionCountDownTimer-tests.tsx
```diff
@@ -1,33 +1,99 @@
...
+      const endAt = oneDayAgo
+      const auctionState = "hasEnded"
+      const date = timeUntil(startAt, liveStartAt, endAt, auctionState)
+      expect(date).toEqual("Ended Aug 14")
+    })
+
+    it("returns 'Ended' string with year when auction hasEnded over a year ago", () => {
+      const startAt = oneDayAgo
+      const liveStartAt = oneDayAgo
+      const endAt = oneYearFromNow
```
sweir27: These tests are tricky because they both pass in timestamps _and_ refer to `auctionState` (which itself depends on those same timestamps).

In this case, because the auction state is `hasEnded`, this `endAt` timestamp should be sometime in the past.

src/lib/Scenes/Artwork/Components/__tests__/AuctionCountDownTimer-tests.tsx
```diff
@@ -1,33 +1,99 @@
...
+      const liveStartAt = oneDayAgo
+      const endAt = oneDayFromNow
+      const auctionState = "isLive"
+      const date = timeUntil(startAt, liveStartAt, endAt, auctionState)
+      expect(date).toEqual("In progress")
+    })
   })
 
-  it("renders formattedStartDateTime", () => {
+  xit("renders correct label for timer", () => {
```
sweir27: Why has this been disabled?

src/lib/Scenes/Artwork/Components/__tests__/AuctionCountDownTimer-tests.tsx
```diff
@@ -1,33 +1,99 @@
...
+      expect(date).toEqual("Ended Aug 14")
+    })
+
+    it("returns 'Ended' string with year when auction hasEnded over a year ago", () => {
+      const startAt = oneDayAgo
+      const liveStartAt = oneDayAgo
+      const endAt = oneYearFromNow
+      const auctionState = "hasEnded"
+      const date = timeUntil(startAt, liveStartAt, endAt, auctionState)
+      expect(date).toEqual("Ended Aug 15, 2020")
```
sweir27: Now that we're correctly setting the timezone based on the user's locale, it seems like we should be _showing_ the timezone here! It's not in the spec but we should ask Sam what she thinks.

src/lib/Scenes/Artwork/Components/AuctionCountDownTimer.tsx
```diff
@@ -2,29 +2,63 @@ import { TimeRemaining } from "@artsy/palette"
...
   artwork: AuctionCountDownTimer_artwork
   auctionState: AuctionState
 }
 
+export const timeUntil = (startAt, liveStartAt, endAt, auctionState) => {
+  if (auctionState === "isPreview") {
+    return `Starts ${formatDateTime(startAt)}`
+  } else if (liveStartAt && auctionState === "hasStarted") {
+    return `Live ${formatDateTime(liveStartAt)}`
+  } else if (auctionState === "isLive") {
```
kierangillen: And here we want to use "hasStarted" on line 18.

src/lib/utils/formatDates.tsx
```diff
@@ -0,0 +1,25 @@
+import moment from "moment"
```
kierangillen: Could we use `Luxon` here instead of `Moment`, we shouldn't be using Moment at all anymore in Emission.

kierangillen:
@sweir27 @lilyfromseattle I think it could be really helpful to write all the states that are possible clearly in a comment in this file, and then from there work out the logic for each condition. Then we can use this as a reference for tests and for all the displays. It could also be helpful then to capture a specific state and have all of the functions use that as a source of truth, that way the coundtowntimer and the date display are using the same source and we will never see weird state conditions.

src/lib/Scenes/Artwork/Components/AuctionCountDownTimer.tsx
```diff
@@ -2,29 +2,63 @@ import { TimeRemaining } from "@artsy/palette"
...
 
 interface AuctionCountDownTimerProps {
   artwork: AuctionCountDownTimer_artwork
   auctionState: AuctionState
 }
 
+export const timeUntil = (startAt, liveStartAt, endAt, auctionState) => {
+  if (auctionState === "isPreview") {
+    return `Starts ${formatDateTime(startAt)}`
+  } else if (liveStartAt && auctionState === "hasStarted") {
```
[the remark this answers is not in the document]
sweir27: `hasStarted` does not mean that the auction is open for live bidding (that's what `isLive` is reserved for).

`hasStarted` means that the auction is open for pre-bidding on the website (so the `startAt` is in the past, but `liveStartAt` is in the future).

This condition _could_ be updated to check for the `liveStartAt` timestamp, but that should be captured already in the auction state.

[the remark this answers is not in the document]
kierangillen: Okay got it, this may be a good reason to document these conditions in comments in this component as the logic is complicated.

src/lib/Scenes/Artwork/Components/AuctionCountDownTimer.tsx
```diff
@@ -2,29 +2,63 @@ import { TimeRemaining } from "@artsy/palette"
...
+  } else {
+    return null
+  }
+}
+
 export class AuctionCountDownTimer extends React.Component<AuctionCountDownTimerProps> {
+  countdownValue(sale, auctionState) {
+    if (auctionState === "isPreview") {
+      return sale.startAt
+    } else if (auctionState === "hasStarted" && sale.liveStartAt) {
```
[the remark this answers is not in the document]
sweir27: Either one could work here.

Any type of auction can be in preview mode, so technically the `Starts <>` messaging works for a live auction before it opens for pre-bidding.

src/lib/Scenes/Artwork/Components/__tests__/AuctionCountDownTimer-tests.tsx
```diff
@@ -1,33 +1,99 @@
 import { Theme, TimeRemaining } from "@artsy/palette"
 import { mount } from "enzyme"
 import { ArtworkFixture } from "lib/__fixtures__/ArtworkFixture"
-import { Settings } from "luxon"
+import moment from "moment"
```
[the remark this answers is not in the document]
sweir27: Good call.

sweir27:
@kierangillen good idea on documenting.

@lilyfromseattle would you like to take a stab? I'm happy to give feedback!

Re: the question of whether the countdown timer and date display should depend on auction state-- they _could_ but when we discussed before it didn't seem necessary for it to update in realtime. If it feels like a clearer architectural choice we should definitely do that, but will leave it up to you two.

src/lib/Scenes/Artwork/Components/__tests__/AuctionCountDownTimer-tests.tsx
lilyfromseattle: Whoops, meant to put oneYearAgo

kierangillen:
The parent component has a `setInterval` keeping track of the auction state, we should still use that as an overall source of truth and that will update this component when the auction state changes. We should probably look into having some local state also in this component for all of the different display possibilities.

src/lib/Scenes/Artwork/Components/AuctionCountDownTimer.tsx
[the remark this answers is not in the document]
lilyfromseattle: I asked about this and the general consensus was we should show "Starts" instead of "Live"

lilyfromseattle:
@kierangillen Not sure what you mean, as I'm already using auctionState as a source of truth

src/lib/Scenes/Artwork/Components/__tests__/CommercialInformation-tests.tsx
```diff
@@ -1,6 +1,7 @@
 import { Sans, Theme } from "@artsy/palette"
 import { mount } from "enzyme"
 import { ArtworkFixture } from "lib/__fixtures__/ArtworkFixture"
+import "moment-timezone"
```
sweir27: Do we need this import anymore?

src/lib/Scenes/Artwork/Components/__tests__/AuctionCountDownTimer-tests.tsx
```diff
@@ -1,33 +1,97 @@
...
+  const dateNow = 1565870400000 // 2019-08-15T12:00:00.000Z in milliseconds
+  let oneDayAgo
+  let oneDayFromNow
+  let oneYearAgo
+
+  beforeEach(() => {
+    jest.useFakeTimers()
+
+    // 2019-08-15T12:00:00.000Z
+    Date.now = () => dateNow
```
kierangillen: @lilyfromseattle We need to reset this back to the real `Date.now` after the tests complete.

src/lib/utils/__tests__/formatDates-tests.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const dateNow = 1525983752000 // Thursday, May 10, 2018 8:22:32.000 PM UTC in milliseconds
+let oneDayFromNow
+let oneYearFromNow
+
+beforeEach(() => {
+  jest.useFakeTimers()
+
+  // Thursday, May 10, 2018 8:22:32.000 PM UTC
+  Date.now = () => dateNow
```
kierangillen: Same here, we can add a `const realNow = Data.now` at the top of the file and then `afterAll(() => Date.now = realNow)` in the test

PR #1913, opened 2019-10-10
Fix query for messages list in conversation

ds300:
This was using the relay node ID instead of the impulse conversation ID
#trivial


ashfurrow:
Makes sense – what issue was this causing? Anything user-facing?

ds300:
Yeah this affected the pagination fetch rather than the initial page fetch, so if there were more messages that weren't loaded initially you could never load them.

ashfurrow:
Ah got it. Since the most recent messages we're on top, I think we're okay to wait for the next release to deploy this to users 👍

alloy:
Oof, good catch, thanks!

PR #1914, opened 2019-10-11
ME-18 Send graphql request metrics to datadog

ds300:
https://artsyproduct.atlassian.net/browse/ME-18

- Added datadog logging of graphql fetch times along with operation name + id

After this gets merged we need to make a small change to eigen in `ARAppDelegate+Emission.m`

image


src/lib/relay/middlewares/timingMiddleware.ts
```diff
@@ -0,0 +1,18 @@
...
+  return next => req => {
+    const startTime = Date.now()
+    const operation = req.operation.name || "UnknownOperation"
+    return next(req).then(res => {
+      const duration = Date.now() - startTime
+      volleyClient.send({
+        type: "timing",
+        name: "graphql-request-duration",
+        timing: duration,
+        tags: [`operation:${operation}`],
```
ashfurrow: In addition to the query `name`, could we include the query `id`? That way, we can track performance improvements if we change the query structure.

ds300: great idea! :+1:

ds300:
@ashfurrow it feels kinda silly to thread this `volleyURL` env var through to eigen when it's not being used there. Actually it feels kinda silly to have the env var for this at all, since it's value is hard-coded anyway 🤷‍♂

Would you object to my keeping the URLs in the `volleyClient.ts` file for now?

ashfurrow:
@ds300 Hmm, yeah that's a good point. I was envisioning a way for Eigen to also report load times, but I agree that for now it's fine to keep the logic in Emission 👍

src/lib/utils/volleyClient.ts
```diff
@@ -0,0 +1,105 @@
...
+    { trailing: true }
+  )
+  constructor(public readonly serviceName: string) {}
+  send(metric: VolleyMetric) {
+    this.queue.push(metric)
+    this._dispatch()
+  }
+}
+
+const isStaging = gravityURL.includes("staging")
```
ashfurrow: Oh, could we also update this to check the Emission setup object's `inStaging` property?

ds300: I refactored to use a global `env` option which is either `'production'`, `'staging'`, or `'test'` since we seem to need this distinction in a few different places.

ds300:
@ashfurrow  this should be 🍏

Pod/Classes/Core/AREmission.m
```diff
@@ -32,6 +43,8 @@ - (NSDictionary *)constantsToExport
     @"predictionURL": self.predictionURL,
     @"userAgent": self.userAgent,
     @"options": self.options,
+    // production | staging | development | test
```
ashfurrow: It doesn't look like `development` belongs here, based on the assert?
```suggestion
    // production | staging | test
```

PR #1916, opened 2019-10-14
ME-21 - Moving chevron

kierangillen:
- https://artsyproduct.atlassian.net/browse/ME-21

## Now:

...nothing happens

#trivial
#skip_new_tests


sweir27:
Wait wait but why was it moving before? 😮

kierangillen:
@sweir27 Tbh not exactly sure but I think it could be a RN bug with the `Text` element adding whitespace before the SVG, I think probably best to avoid wrapping SVGs with `Text` elements 🤷‍♂

kierangillen:
@sweir27 Actually the margin-left props is the issue, for some reason the `Text` element appears to be rendering it over and over while `onPress` is held down and its `ML` props are being applied `++`

sweir27:
@kierangillen whoa. Good find!!

PR #1917, opened 2019-10-14
ME-32 - Fixes Composer text input clip

kierangillen:
- https://artsyproduct.atlassian.net/browse/ME-32

## After change

#trivial
#skip_new_tests


ds300:
Oh interesting! How is this offset related to the status bar height?

PR #1921, opened 2019-10-17
Fixes a few issues on the artwork view while auction is open for live bidding

sweir27:
Addresses: https://artsyproduct.atlassian.net/browse/PURCHASE-1597

Before:
image

There are a few issues with the above image:
- We should not be showing the "availability" message and corresponding partner information for this case.
- We should be showing the timer (at 0, with `In progress` as the label)
- We should be correctly showing `Enter live bidding` vs. `Watch live bidding`

I confirmed that out logic for when to show `Enter` vs. `Watch` is correct, so I suspect the original bug report was due to weirdness around auction state transitions. 

When I am able to register OR already registered:
Screen Shot 2019-10-17 at 2 38 50 PM

When I am not able to register:
Screen Shot 2019-10-17 at 2 18 26 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 - Add graphql request duration reporting via volley - ds300
 - Fixes `Contact Gallery` button showing up when artwork auction is closed - sepans
+- Updates CommercialInformation component to properly handle live auctions - sweir27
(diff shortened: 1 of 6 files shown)


artsyit:

PR #1922, opened 2019-10-17
[PURCHASE-1600] Hide 'contact gallery' button when artwork sale is closed

sepans:
Fixes [PURCHASE-1600](https://artsyproduct.atlassian.net/browse/PURCHASE-1600)

Before:

Screen Shot 2019-10-17 at 2 50 19 PM

After:

Screen Shot 2019-10-17 at 2 49 31 PM

### DIFF
--- a/src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx
+++ b/src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx
@@ -35,6 +35,8 @@ export class CommercialButtons extends React.Component<CommercialButtonProps> {
     const { isBuyNowable, isAcquireable, isOfferable, isInquireable, isInAuction, editionSets, isForSale } = artwork
     const noEditions = (editionSets && editionSets.length === 0) || !editionSets
 
+    const artworkSaleEnded = artwork.sale && artwork.sale.isClosed
+
     if (isInAuction && artwork.sale && auctionState !== "hasEnded" && isForSale) {
       return (
         <>
@@ -61,7 +63,7 @@ export class CommercialButtons extends React.Component<CommercialButtonProps> {
       return <BuyNowButton artwork={artwork} editionSetID={this.props.editionSetID} />
     } else if (isOfferable) {
       return <MakeOfferButton artwork={artwork} editionSetID={this.props.editionSetID} />
-    } else if (isInquireable) {
+    } else if (isInquireable && !artworkSaleEnded) {
       return (
         <Button onPress={() => this.handleInquiry()} size="large" block width={100}>
           Contact gallery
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx
```diff
@@ -35,6 +35,8 @@ export class CommercialButtons extends React.Component<CommercialButtonProps> {
     const { isBuyNowable, isAcquireable, isOfferable, isInquireable, isInAuction, editionSets, isForSale } = artwork
     const noEditions = (editionSets && editionSets.length === 0) || !editionSets
 
+    const artworkSaleEnded = artwork.sale && artwork.sale.isClosed
+
     if (isInAuction && artwork.sale && auctionState !== "hasEnded" && isForSale) {
```
sepans: Question: is there a difference between this `auctionState !== "hasEnded"` and `artwork.sale.isClosed`? If not isn't it better to update that to also use `isClosed`?

sweir27: We prefer to use `auctionState` for this because it will change client-side depending on the status of the timer.

When `auctionState` is instantiated, it will take into account whether the auction `isClosed`.

src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx
```diff
@@ -61,7 +63,7 @@ export class CommercialButtons extends React.Component<CommercialButtonProps> {
       return <BuyNowButton artwork={artwork} editionSetID={this.props.editionSetID} />
     } else if (isOfferable) {
       return <MakeOfferButton artwork={artwork} editionSetID={this.props.editionSetID} />
-    } else if (isInquireable) {
+    } else if (isInquireable && !artworkSaleEnded) {
```
sweir27: To replicate the logic on desktop, we can just use `!artwork.isInAuction` here!

https://github.com/artsy/reaction/blob/6f883d2603287af8dbed5671f24e889a614dc1dd/src/Apps/Artwork/Components/ArtworkSidebar/index.tsx#L51

sweir27: It'll accomplish the same thing but seems technically more correct as we never want to show the `Contact Gallery` button if an artwork is in an auction.

sepans:
Closing this in favor of https://github.com/artsy/emission/pull/1923

The spacing was complex so getting the spacing correct without the risk of breaking the layout of other screen variations was not possible so after discussing options with @zephraph decided to move the logic to the parent component to make sure neither the button no the spacer above it gets rendered

src/lib/Scenes/Artwork/Components/CommercialButtons/CommercialButtons.tsx
sepans: there is also this guy inside the `ArtworkSidebarBidAction`: https://github.com/artsy/reaction/blob/master/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarBidAction.tsx#L92 🤔

PR #1924, opened 2019-10-18
Switch to use formattedStartDateTime for meta all the time

ashkan18:
# Problem
https://artsyproduct.atlassian.net/browse/PURCHASE-1598

# Fix
The bigger fix was in https://github.com/artsy/metaphysics/pull/1995 where we fixed issue with `formattedStartDateTime` not checking `ended_at`.

Here we just remove our extra check and we rely on `formattedStartDateTime` to do all magic.


ashkan18:
@lilyfromseattle @sweir27 this is unblocked now since MP change is deployed

artsyit:

PR #1926, opened 2019-10-21
User is able to see buyer's premium cost on iOS

yuki24:
addresses https://artsyproduct.atlassian.net/browse/AUCT-674

## Todos

 * [x] ~Loading indicator visually broken?~ paired with @ashfurrow but we couldn't get to the root cause. We will re-visit once the rest of the work is done.
 * [x] Add more test coverage
 * [x] Use the `<Flex>` component from Palette
 * [x] Add a feature flag?

### Without payment info form

Screen Shot 2019-10-21 at 4 58 02 PM

### With payment info form

Screen Shot 2019-10-21 at 4 59 09 PM

#skip_new_tests


yuki24:
Feature flag has been added, CI is green and this is ready for review.

Example/Emission/ARLabOptions.m
```diff
@@ -22,7 +22,7 @@ + (void)initialize
     // object.
     //
     options = @{
-       @"nothingYet": @"No feature flags available"
+       @"enablePriceTransparency": @"Price tranparency"
```
dleve123:
```suggestion
       @"enablePriceTransparency": @"Price transparency"
```

note the typo in the suggestion!

src/lib/Components/Bidding/Screens/ConfirmBid/PriceSummary.tsx
```diff
@@ -0,0 +1,95 @@
...
+        display
+      }
+      subtotal {
+        display
+      }
+    }
+  `,
+})
+
+interface PriceSummaryProps extends Partial<PriceSummaryQueryVariables> {
```
dleve123: why is this a `Partial<T>` type? I guess considering that we know that the UI is providing a bid object, what protection/value does making this a union type of the query variables + others provide? Perhaps we can talk through this for a min so that I can learn here.

src/lib/Components/Bidding/Screens/ConfirmBid/PriceSummary.tsx
```diff
@@ -0,0 +1,95 @@
...
+            }
+          }
+        }
+      }
+    `}
+    variables={{
+      saleArtworkId,
+      bidAmountCents: bid.cents,
+    }}
+    render={renderWithLoadProgress(({ node }) => (
```
dleve123: if this doesn't work, should we still keep it around?

src/lib/Components/Bidding/Screens/ConfirmBid/index.tsx
```diff
@@ -412,9 +413,11 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
   }
 
   render() {
-    const { artwork, lot_label, sale } = this.props.sale_artwork
+    const { id, artwork, lot_label, sale } = this.props.sale_artwork
     const { requiresPaymentInformation, requiresCheckbox, isLoading } = this.state
     const artworkImage = artwork.image
+    const enablePriceTransparency =
```
dleve123: perhaps `enablePriceSummary` is the most straightforward name (related to a comment above).

src/lib/Components/Bidding/Screens/ConfirmBid/index.tsx
```diff
@@ -425,7 +428,7 @@ export class ConfirmBid extends React.Component<ConfirmBidProps, ConfirmBidState
               <Timer liveStartsAt={sale.live_start_at} endsAt={sale.end_at} />
             </Flex>
 
-            <View>
+            <Box>
```
dleve123: Out of curiosity, why the change to `Box`?

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -110,6 +122,55 @@ it("displays the artwork title correctly without date", () => {
...
+          },
+          subtotal: {
+            display: "$54,000.00",
+          },
+        },
+      },
+    },
+  }))
+
+  expect(component.root.findAllByType(Spinner).length).toEqual(0)
```
dleve123: If the loader code was working, would this test failure? I guess what is this assertion supposed to test?

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
```diff
@@ -110,6 +122,55 @@ it("displays the artwork title correctly without date", () => {
   expect(serifChildren(component)).not.toContain(`${saleArtwork.artwork.title},`)
 })
 
+it("can load and display price summary", () => {
```
dleve123: Have we considered / would it make sense to consider what happens when the MP query for the `calculatedCost` field fails? I guess this could happen when:

- MP is down. Presumably a lot more would fail.
- There's a bug within the code called from querying for `calculatedCost` (seems more likely)

Would it be testing the case that we have some form of error from MP to ensure our front-end is resilient to reasonable backend errors? Thoughts?

src/lib/Components/Bidding/Screens/__tests__/SelectMaxBid-tests.tsx
```diff
@@ -1,3 +1,7 @@
+jest.mock("lib/Components/Bidding/Screens/ConfirmBid/PriceSummary", () => ({
```
dleve123: not clear to me why this diff is needed - can you explain?

src/lib/Components/Bidding/__tests__/BidFlow-tests.tsx
```diff
@@ -1,3 +1,7 @@
+jest.mock("lib/Components/Bidding/Screens/ConfirmBid/PriceSummary", () => ({
```
dleve123: I guess this is a similar comment to the above, but does any dep of ConfirmBid within tests now require that we stub out the component? If so, that seems a bit unfortunate :(

src/lib/Components/Bidding/Screens/ConfirmBid/index.tsx
dleve123: So, assuming that we'll want to roll out this component to all users in production after beta'ing it with a smaller group of lab users, would we need to have another release of emission and eigen to remove this lab feature switch, therefore rolling it out to everyone?

Just want to make sure that we shouldn't be using Echo here to also enable us to turn this on across all of our devices.

Some relevant docs here that just wanna make sure are surfaced your way: https://github.com/artsy/emission/blob/master/docs/adding_a_lab_option.md

data/complete.queryMap.json
```diff
@@ -153,5 +153,17 @@
...
...em_show on Show {\n  internalID\n  slug\n  name\n  exhibition_period: exhibitionPeriod\n  end_at: endAt\n  images {\n    url\n  }\n  partner {\n    __typename\n    ... on Partner {\n      name\n    }\n    ... on Node {\n      id\n    }\n    ... on ExternalPartner {\n      id\n    }\n  }\n}\n\nfragment ArtistListItem_artist on Artist {\n  id\n  internalID\n  slug\n  name\n  initials\n  href\n  is_followed: isFollowed\n  nationality\n  birthday\n  deathday\n  image {\n    url\n  }\n}\n\nfragment GenericGrid_artworks on Artwork {\n  id\n  image {\n    aspect_ratio: aspectRatio\n  }\n  ...ArtworkGridItem_artwork\n}\n\nfragment ArtworkGridItem_artwork on Artwork {\n  title\n  date\n  sale_message: saleMessage\n  is_biddable: isBiddable\n  is_acquireable: isAcquireable\n  is_offerable: isOfferable\n  slug\n  sale {\n    is_auction: isAuction\n    is_closed: isClosed\n    display_timely_at: displayTimelyAt\n    id\n  }\n  sale_artwork: saleArtwork {\n    current_bid: currentBid {\n      display\n    }\n    id\n  }\n  image {\n    url(version: \"large\")\n    aspect_ratio: aspectRatio\n  }\n  artists(shallow: true) {\n    name\n    id\n  }\n  partner {\n    name\n    id\n  }\n  href\n}\n",
```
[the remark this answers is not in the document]
yuki24: On each Emission deploy GraphQL queries are sent to MP and cached there, and all Emission does is to send a key to the query so it doesn't have to send the entire query (which reduces bytes transferred over network). This is an automated change and normally you don't have to pay close attention unless you are tweaking the query map itself.

data/schema.graphql
```diff
@@ -1464,6 +1464,30 @@ type BuyersPremium {
   percent: Float
 }
 
+type BuyersPremiumAmount {
```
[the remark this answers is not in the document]
yuki24: We use `yarn sync-schema` to update this file.

src/lib/Components/Bidding/Screens/ConfirmBid/PriceSummary.tsx
yuki24: The only thing that's not working is the animation and the `<Spinner>` component itself is still mountable. I don't think it is worth dropping given that the `renderWithLoadProgress` takes care of [a lot of cases](https://github.com/artsy/emission/blob/6d0e5f50022e81938a29e22f916b1850754d46aa/src/lib/utils/renderWithLoadProgress.tsx#L16-L51) (showing a loader, presenting an error message in case of a failure, sending an error to Sentry).

src/lib/Components/Bidding/Screens/ConfirmBid/index.tsx
yuki24: We generally avoid `<div>` or `<View>` that are platform-specific and use `<Box>` (which is universal / copy / pasteable to RN).

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
yuki24: The `<Spinner>` component is still mountable and this test passes. The only thing that's not working is animation which is encapsulated within this component.

src/lib/Components/Bidding/Screens/ConfirmBid/index.tsx
yuki24: > after beta'ing it with a smaller group of lab users

Why do we need that level of feature flagging? Or do we actually need feature flagging at all?

src/lib/Components/Bidding/Screens/__tests__/SelectMaxBid-tests.tsx
yuki24: In test we generally don't mount any components that depend on `<QueryRenderer>` because they are just untestable (otherwise the error would be thrown).

```
Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

    Check the render method of `YourComponent`.
```

In `src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx`, we avoid this by using [an environment mock](https://github.com/artsy/emission/pull/1926/files#diff-e4b585673014c3d994c1ed23be87f302R129-R143). The price summary is perhaps not the thing we are interested in testing the most, so I think it makes sense to mock it out entirely.

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
yuki24: The `renderWithLoadProgress` also adds error handling and presenting a general error message. That said the `<LoadFailureView>` is designed to fit in a larger viewport and it gets squashed on a smaller screen like iPhone SE, and it is definitely something we should fix before release.

I've created a ticket for this https://artsyproduct.atlassian.net/browse/AUCT-674 on JIRA, but I would like us to be more iterative than making the implementation perfect in the first place.

src/lib/Components/Bidding/Screens/ConfirmBid/PriceSummary.tsx
yuki24: The `PriceSummaryQueryVariables` interface is auto-generated by Relay so it'll add a little safety for when we update the query argument. It has two arguments: `saleArtworkId` and `bidAmountCents`. I made it a partial since the object that gets injected into this component is a `Bid` object (we use both `bid.cents` and `bid.display` to render the component), and we don't use the other `bidAmountCents` argument in the `PriceSummaryQueryVariables` interface.

src/lib/Components/Bidding/Screens/ConfirmBid/PriceSummary.tsx
dleve123: Gotcha, wasn't aware of the other features here. Makes sense to me!

src/lib/Components/Bidding/Screens/__tests__/ConfirmBid-tests.tsx
dleve123: You linked to the main story, which you didn't create, but perhaps you meant to link to either:

- https://artsyproduct.atlassian.net/browse/AUCT-692 or
- https://artsyproduct.atlassian.net/browse/AUCT-695

How are these issues different btw? Is AUCT-692 strictly related to the visuals of the loading animation, while AUCT-699 is strictly related to the behavior when the endpoint is down?

---

Re being iterative: 9 times out of 10, I think we should totally optimize for iterative PRs and releases. I'm just being a bit cautious here as I know that there's no guarantee that iOS users update to the latest release.

So when you write:

> but I would like us to be more iterative than making the implementation perfect in the first place.

Are you suggesting that we allow a release an emission, but ensure we don't deploy Eigen until we resolved the mentioned issues? If so, totally down, but please communicate to Ash and the mobile experience team that we'll want the equivalent of a "deploy block" for Eigen until we resolve (to avoid the scenario of deploying emission mid-build). Perhaps this, in and of itself, is a good reason to have a lab feature here (as it would remove the need for such a deploy block).

dleve123: We want a feature flag like the following:

`if echo's feature flag is true OR the lab feature is enabled, then display the PriceSummary`.

this would allow us to:

1. deploy emission to eigen while we're working on this feature across multiple PRs as the lab feature is off by default, while simultaneously allowing us to.
2. Enable for _all users_ when we want to by turning on an echo flag.

---

- We are not going to handle the error states changes in this PR
- @yuki24 will create and work on a ticket to implement the echo based feature flag as a next up item.
  - ideally, we wouldn't deploy eigen until we have this echo based feature flag.

yuki24: > Perhaps this, in and of itself, is a good reason to have a lab feature here (as it would remove the need for such a deploy block).

If this ever becomes a release blocker, we could just remove it from the `<ComfirmBid>` component (literally a one-line change) and move it to the StoryBook. So I don't think Emission-only feature flagging still makes sense here. I don't think we should remove it at this point though since it's already there.

yuki24:
The CI is failing due to a compile error. I'm not sure if I could solve it but I'll look into it anyway.

yuki24:
Given that this is disabled by default and it wouldn't affect production code, I would say this is safe to merge at this point.

artsyit:

PR #1927, opened 2019-10-21
Adds react-native Partner view

kierangillen:
- This PR covers the first stage in releasing a react-native Partner view. Follow up PR's will be required for adding tests, updates to views, etc.

#skip_new_tests


Pod/Classes/ViewControllers/ARPartnerComponentViewController.m
```diff
@@ -0,0 +1,32 @@
+#import "ARPartnerComponentViewController.h"
+#import "ARGraphQLQueryPreloader.h"
+
+@implementation ARPartnerComponentViewController
+
++ (NSDictionary *)propsWithPartnerID:(NSString *)partnerID;
+{
+    BOOL isPad = UI_USER_INTERFACE_IDIOM() == UIUserInterfaceIdiomPad;
+    return @{ @"partnerID": partnerID, @"isPad": @(isPad) };
```
ashfurrow: We can use the existing `isPad` function in [`hardware.ts`](https://github.com/artsy/emission/blob/master/src/lib/Scenes/Artwork/hardware.ts) (though we may want to move that file to a higher-level directory).

Example/Emission/ARRootViewController.m
```diff
@@ -253,60 +249,18 @@ - (ARCellData *)jumpToArtwork
...
-{
-  return [self tappableCellDataWithTitle:@"Artwork With Institution Partner" selection:^{
-    id viewController = [[ARArtworkComponentViewController alloc] initWithArtworkID:@"jean-leon-gerome-the-guard"];
-    [self.navigationController pushViewController:viewController animated:YES];
-  }];
-}
-
-
-- (ARCellData *)jumpToArtworkAttributionClassFAQ
+- (ARCellData *)jumpToPartner2
```
ashfurrow: I take it these changes were for local development? 😂

src/lib/AppRegistry.tsx
```diff
@@ -87,6 +88,13 @@ interface ArtworkProps {
 
 const Artwork: React.SFC<ArtworkProps> = props => <ArtworkRenderer {...props} />
 
+interface PartnerProps {
+  partnerID: string
+  safeAreaInsets: SafeAreaInsets
+}
+
+const Partner: React.SFC<PartnerProps> = props => <PartnerRenderer {...props} />
```
ashfurrow: We should add some basic analytics tracking here before shipping. The Artwork page is the most recently-added component so it should represent the most up-to-date analytics schema; we can mimic that schema here and then get feedback from @anipetrov before shipping 👍

src/lib/Scenes/Partner/Components/PartnerOverview.tsx
```diff
@@ -0,0 +1,38 @@
+import { Box, Sans, Spacer } from "@artsy/palette"
+import { PartnerOverview_partner } from "__generated__/PartnerOverview_partner.graphql"
+import { ReadMore } from "lib/Components/ReadMore"
+import { get } from "lib/utils/get"
+import React from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+
+const textLimit = 50
```
ashfurrow: Do we want to increase this text limit for iPad? `hardware.ts` also has something for this, but the values might not match the design you have in mind.

src/lib/Scenes/Partner/Partner.tsx
```diff
@@ -0,0 +1,176 @@
...
+      onCompleted: () => setIsFollowedChanging(false),
+      onError: e => console.log("errors", e),
+      mutation: graphql`
+        mutation PartnerFollowMutation($input: FollowProfileInput!) {
+          followProfile(input: $input) {
+            profile {
+              id
+              slug
+              internalID
+              isFollowed
```
ashfurrow: I see why we'd need `isFollowed` but do we also need the other attributes here?

src/lib/Scenes/Partner/Partner.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  return (
+    <Theme>
+      <Flex style={{ flex: 1 }}>
+        <Spacer mt={60} />
+        <Flex flexDirection="row" flexWrap="nowrap" justifyContent="space-between">
+          <Box px={2} pb={2}>
+            <Serif size="6">{partner.name}</Serif>
+          </Box>
+          {isFollowed && (
```
ashfurrow: Seems like this `Button` won't get rendered if the partner isn't being followed?

Example/Emission/ARRootViewController.m
kierangillen: Yes, I want to have more than one partner example while developing.

artsyit:
:rocket: PR was released in v1.18.7 :rocket:

PR #1934, opened 2019-10-24
PURCHASE-1599: Consolidates Auction State across Artwork and Bid pages

sweir27:
This PR addresses some issues we were seeing with the auction state transitions.

The immediate changes included are:
- Removes our custom `AuctionCountdownTimer` component in favor of re-using the auction state handling and countdown timer that already exists in the bidding flow. This simplified the logic within the artwork view significantly.
- Removes relay caching on the artwork view. The state of the artwork (and in particular the corresponding _sale_ which is nested within) can change often during the course of an auction, so we need to be able to see the updated state immediately upon refreshing the view. cc @alloy in case you have another idea

A few things that are not necessarily ideal included in here:
- The AuctionCountdownTimer (removed here) used `luxon` instead of `moment`. As I am re-using the existing bid flow timer here (which uses `moment`), I did not refactor to remove that. This should probably be a standalone follow-up as it'll also affect the fair countdown timer.
- The `DurationProvider` wrapper provides a nice abstraction around countdowns, but it means that the view it encompasses is re-rendered every second. I didn't notice any major issues in the simulator, but it's something to watch.
- The existing timer also uses an Emission-specific component instead of the palette-backed `TimeRemaining`, simply because that would also be a significant refactor.

One outstanding issue related to caching:
Metaphysics' current caching scheme means that while we don't cache the response of the `Artwork`, we _do_ cache the response of the `Sale` that's returned from within the artwork.

This creates the following situation:
- A live auction integration sale _ends_
  --> The corresponding Artwork is marked as "Not for sale"
  --> The corresponding Sale is marked as "Closed"
- The artwork view won't _know_ that the auction has closed until Metaphysics' cache is busted, meaning we get a weird in-between state until that happens

This ☝️ is also an issue on force, although the UI handles that by just showing the "Enter Live Bidding" button for an extra minute or so after the live auction has technically closed. We should follow up on this one.

#### PREVIEW
Screen Shot 2019-10-24 at 6 49 44 PM

wee-preview-start-live

#### BEFORE LAI
Screen Shot 2019-10-24 at 6 51 14 PM

wee-start-LAI-live

#### DURING LAI
Screen Shot 2019-10-24 at 6 53 11 PM

#### PREVIEW
Screen Shot 2019-10-24 at 6 37 14 PM

wee-preview-start-oo

#### OPEN
Screen Shot 2019-10-24 at 6 39 14 PM

wee-start-end-oo

#### CLOSED
Screen Shot 2019-10-24 at 6 43 47 PM


src/lib/Components/Countdown/DurationProvider.tsx
```diff
@@ -74,9 +73,8 @@ export class DurationProvider extends React.Component<Props, State> {
   render() {
     const { timeLeftInMilliseconds } = this.state
     const { timeOffsetInMilliseconds, children } = this.props
-
     return React.cloneElement(children, {
-      duration: moment.duration(timeLeftInMilliseconds + (timeOffsetInMilliseconds || 0)),
+      duration: moment.duration(timeLeftInMilliseconds + (timeOffsetInMilliseconds || 0) || 0),
```
zephraph: What are the `||` operators here for?

sweir27: The first is because `timeOffsetInMilliseconds` might not exist (calculates the difference between local time and gravity system time).

The second is because the timestamp passed into this (`startAt`) might be null, which would make the `timeLeftInMilliseconds` a `NaN`, and this entire `moment.duration` object invalid.

There are a bunch of ways to skin this cat, but this change basically ensures that _whatever happens_, we won't ever be showing `NaN`s in our countdown timer.

Currently, if the date was invalid or in the past, our timers would show `NaNm NaNd NaNh NaNm`. After this change, invalid/past dates will show `00m 00d 00h 00m`.

kierangillen:
Nice refactor!! LGTM!

kierangillen:
Screen Shot 2019-10-25 at 10 16 54 AM

Does this entity header image look squished?

src/lib/Components/Bidding/Components/Timer.tsx
```diff
@@ -93,25 +99,28 @@ function currentTimerState({ isPreview, isClosed, liveStartsAt }: Props) {
   }
 }
 
-interface CountdownProps {
+export interface CountdownProps {
   duration: moment.Duration
-  label: string
+  label?: string
+  [propName: string]: {}
```
sweir27: This is an attempt to also accept arbitrary props that need to be passed from the wrapper component.

src/lib/Components/Bidding/Components/__tests__/Timer-tests.tsx
```diff
@@ -169,7 +169,7 @@ it("shows month, date, and hour adjusted for the timezone where the user is", ()
     </BiddingThemeProvider>
   )
 
-  expect(getTimerLabel(timer)).toEqual("Ends May 14, 1 PM")
+  expect(getTimerLabel(timer)).toEqual("Ends May 14, 1 PM PDT")
```
sweir27: With this change, the timer on the confirm bid page will also show timezone (this felt good/preferable).

src/lib/Components/Bidding/Components/Timer.tsx
alloy: Why is there a need for arbitrary props?

src/lib/Components/Bidding/Components/Timer.tsx
```diff
@@ -4,7 +4,13 @@ import moment from "moment-timezone"
...
 import React from "react"
 
-enum AuctionTimerState {
+// Possible states for an auction:
+// - PREVIEW: Auction is open for registration but artworks cannot be bid on. This occurs when the current time is before any auction's startAt.
+// - CLOSING: Timed auction has started
+// - LIVE_INTEGRATION_UPCOMING: Auction is open for pre-bidding, live portion has not started
+// - LIVE_INTEGRATION_ONGOING: Live auction is in progress
+// - CLOSED: Auction is over
+export enum AuctionTimerState {
```
alloy: Nit-pick: When enum values are all the same as their keys, I don’t see any benefit in using an enum over a string union.

Especially not if the value needs to be passed to an API that wants the actual value. I don’t know if that applies in this case, but as it sometimes is the case (eventually) I personally lean towards never doing it.

sweir27: These values don't need to be passed to any kind of API.

I personally found the enum to be _super helpful_ while writing this code (due to intellisense).

sweir27:
@alloy Not all artworks are in sales. Both of your proposed solutions sound okay, I just wasn't sure how to accomplish them. Can you point me in the right direction?

sweir27:
@kierangillen the entity header does look squashed 😞 appears to be happening on master as well.

src/lib/Scenes/Artwork/Components/CommercialInformation.tsx
```diff
@@ -1,62 +1,85 @@
...
+              const nextState = nextTimerState(state as AuctionTimerState, { liveStartsAt })
+              const { label, date } = relevantStateData(nextState, { liveStartsAt, startsAt, endsAt })
+              return { state: nextState, date, label }
+            }}
+            {...this.props}
+          />
+        </TimeOffsetProvider>
+      )
+    } else {
+      return <CommercialInformation {...this.props} />
```
zephraph: This is so much better!

src/lib/Components/Bidding/Components/Timer.tsx
sweir27: In this case, the component that's being wrapped needs a relay prop `CommercialInformation_artwork` passed to it  (but it could be anything).

joeyAghion:
@sweir27 so you're seeing the staleness we anticipated in https://artsyproduct.atlassian.net/browse/PLATFORM-1309 in the wild? If so, seems like we should implement what was discussed in that ticket and bypass typical caching for sales (or perhaps selectively, like for authenticated callers or sales that aren't yet closed). I wasn't racing to do that since I didn't realize it was a confirmed problem, but this changes things.

mzikherman:
I don't fully follow the cache staleness issue. Is it that once a sale state is updated, which should be reflected in the UI, that takes 60 seconds for the MP cache to clear, and thus the UI might show a link to a live sale, for up to 60 seconds after it's over?

If so, that might be behavior introduced when we switched MP caching away from the 'eagerly warming the cache' to a more standard one. I can imagine the former being faster to update.

That cache is configurable, so we might be able to set it better for sales being requested for an artwork, or whatever.

sweir27:
@mzikherman That's right!

src/lib/Components/Bidding/Components/Timer.tsx
alloy: Oh yes, absolutely that’s what we need, but a string union _should_ provide the same intellisense, did it not?

alloy:
> If so, that might be behavior introduced when we switched MP caching away from the 'eagerly warming the cache' to a more standard one. I can imagine the former being faster to update.

Perhaps both apply or we are conflating things, but I understood this to be about the cache-config `force` parameter that’s passed to Relay when performing the operation, which would make it ignore the response cache it has for the given operation+variables.

alloy:
> @alloy Not all artworks are in sales. Both of your proposed solutions sound okay, I just wasn't sure how to accomplish them. Can you point me in the right direction?

Would it be feasible, if not already the case, to wrap all of the sale related containers in a single refetch container? If so, then that refetch container could force-fetch _after_ loading the data from the response cache, meaning you can either:

* optimistically show the entirety of the artwork view from cache and update the sale part when the up-to-date data comes in
* pessimistically only show the non-sale parts of the UI until the up-to-date comes in

src/lib/Components/Bidding/Components/Timer.tsx
alloy: [Here’s a solution](https://www.typescriptlang.org/play/index.html?jsx=2#code/PQKhCgAIUgVALAppA5gewIYBtIEsDOkALmpBgMbmIAORZkASohXeWgLbVoB2i3dRAJ7VkGQhkjUATmmqQAFAGEArvhLtFaZfwAmaAO7dNnHnyIBKYvAxEoMfPC1YdZSjTrTZhfCPK4AZriILgBGgmTcePyIUv4UyEpaugZGHFy8-AAKMtT4ltRi+MGQYVaIdpIYUmYAdJAAgjo6uES4PNhY4URIldX8kGwmGXQOTi4F+PgAtN0yyijwEeH+ylLdMZI5hN02FeuQKmocmtpEeobG6WauVLT4NXbA4OC4JmuMzOR0-jLskABE1RY-2efGUf3qyi+bW4sFeMQAykQbMgAN5QSCZBgAUQAagBJbEAdUgAF4AVi8YSif8ADQYgAy+Nx2IA+viAHKwbEAcQY9Vg+IA8hzWQBVTKKIUAWU5PLJAKZLPZXN5-MFIvFkplcrpjOZbM53L5AuFopFPKFcoV-yVhtVJo15o5lt19MgBwZQoR1vJ-0UXp9Lr1HoD3uxABEbWGEZG9QBfZ64aKxeIHJJnFKXUxZLaQdEenSrGwwgBckG44JCMQxWAw1awAH5y2opMmUBjWuxEciiIhy5DoTw4d2pEiUeBE+AhCJIOO+9KMNwMCgYtkvAAeDGgCAe6CQAASMWQ+mQOkQ5Dr1SsNjKqD4MVw5EgAANDuoTskLmkc0R17kX02LxIFGZRnAxfdk0vZRzzvNQUUgdglxXGIAHJCDfDNzlSIYzH-fAXwePcQCeUNVA-LCsx-YZ8MgRAAA8+24HRCE-TNv1w3MvHAAA+BUC0gHcIJgCNEECXg7zfcjjkojirn4QCSBKUQPhYAZqOufQWngYTX3fGTTmw7MaK2QjdN0oV-ArNA+1pSBtCwRBJj06SNFknD5L-Uy8EIBjqCwJ8Wk6EDfACIIdDs7oCB8spdP8tN9CQa9YGERAEXINtaEgLSsBwZN-A2RLb32GdkCwNA0AAa0IALKtEZiyEmXAUG4QgWmIUh1l0qSjjcwyqM4ryvEU1Legwbs+ykIiPRIjF9L6r8PN-csmBYGpjLMFKRA3ea2KMjSuNyHjJ0gAAybcwF0+osHwUgKFuOgl3CTxchvR7r2TFzer2gbPPwwCGIeyB-DQKQrDQIpup+uTf3+uzEqfRYqmQfYXu2aw6BynBq0qSZghmeA5gWaboFIyAhXYFodtc6GlpMrw7LqwQ0Cs2mNsO-BjvAS9CjnXtEEXZdVykLcyO+9z2aG16GKYlj036mH6dyXi6MYvg5dWr51oOohRb5lFBZQqR8Op8WFbpvCtmOviBL6c8pHkSwBI9IS9xmuBUvSzLWCXQBMAjoapbqwAA3ZByFcoDXqS5BPqiwhV14Ntn3iqhIEERAiDskJlEe9GCAg93ovwV5-PCCQJmmWYtAWbLRBu0g9G4f3ICBXLwnIZGyBCLQBB6ZNmmD3Ai2wSP7kgIlyjd-ceGCjvr3rHvwaKbLo9fecBeQ4XAOi6t23soodBJmayY9NhWroVEDhpiWdbsmp77RyB4zJAuPTjmpH7EezuEq7gUka-WC5N5ri2DtG+g18JczdmfNQkA6wNhtLdbscD6yICwCCaBPBYFFikCWHgCoAAsAAmAuMCRj8wVIOVow54Rjn5utQMcpnhu2qEQVYkR5Cv0gKbCi5tJZcI9PAtBpJURCKwImKee4cF4O4CI6R1DuASMkcQWh68RHwT7EoyRqJ74fy2FovcwBjpu3MBiRMU5SrpnYKOPw2B8TcBBlIJCCjWRVCIPoUGlVowcBsbgOxDjQbOJhK4tYHipCVRBNzLBdAEQcEQLtcBnkVqfCIDUBEAAxRQes2Y61oqdfMZBQmePLMYXx-jHFBJ4CE9xnin4q3JI-UkNsMSsPYQoAuG4B7GKnqiNGNQxHPymPmPp8iYSDOGVsGoXYewogMRuYAXSMSmKnBudehthZzWvnwnWIjYndgSdswaSi3FhMqiI-0PiYi2KwPYipMjqmnP+ImIx4AgA) that makes `StateManager` generic in a way that it verifies the given countdown component takes the expected props _as well_ as verifying the additional props that the custom countdown component expects is passed to the `StateManager`. Let me know if it needs further explanation and/or if you want to pair!

sweir27:
@alloy I figured out that when refetching, we _were_ getting new data from metaphysics, so I updated the code to make sure it was handling that okay (the `StateManager` didn't handle new props after the initial render).

It also seems that the caching issue is scoped to Emission's example app, since we already have handling for re-fetching data when someone pulls to refresh or the view becomes visible. We can still make the change you're suggesting but I don't think it should block this work.

I didn't make the TypeScript changes you suggested because I didn't have time, but I'm happy to follow up with you on that.

src/lib/Scenes/Artwork/Components/__tests__/CommercialInformation-tests.tsx
```diff
@@ -1,16 +1,16 @@
...
 import React from "react"
 import { TouchableWithoutFeedback } from "react-native"
 import { ArtworkExtraLinks } from "../ArtworkExtraLinks"
-import { AuctionCountDownTimer } from "../AuctionCountDownTimer"
 import { BidButton } from "../CommercialButtons/BidButton"
 import { BuyNowButton } from "../CommercialButtons/BuyNowButton"
 import { CommercialButtons } from "../CommercialButtons/CommercialButtons"
 import { CommercialEditionSetInformation } from "../CommercialEditionSetInformation"
-import { CommercialInformation } from "../CommercialInformation"
+import { CommercialInformationTimerWrapper } from "../CommercialInformation"
```
alloy: For future reference, it might have read nicer to rename this locally to just `CommercialInformation`.

```suggestion
import { CommercialInformationTimerWrapper as CommercialInformation } from "../CommercialInformation"
```

alloy:
> We can still make the [caching] change you're suggesting but I don't think it should block this work.

Agreed, and as discussed the way to achieve it (by first rendering from cache and then updating from network) would require some dedicated research work, applies to all of [future] Emission, and is very UX related; so I think it makes sense for the Mobile Experience team to take that on. For instance, I know that @ds300 wants to eliminate spinners as much as possible. /cc @ashfurrow 

> I didn't make the TypeScript changes you suggested because I didn't have time, but I'm happy to follow up with you on that.

Let’s please do that 👍

artsyit:
:rocket: PR was released in v1.18.10 :rocket:

PR #1936, opened 2019-10-25
Rename enablePriceTransparency to AROptionsPriceTransparency for consistency

yuki24:
Just to get aligned with https://github.com/artsy/eigen/pull/2934 so we don't have to do extra key conversion for feature flagging in multiple platforms.

#trivial #skip_new_tests


artsyit:
:rocket: PR was released in v1.18.6 :rocket:

PR #1937, opened 2019-10-25
[ME-65] Migrate Artist page to use same sort as web

ashfurrow:
This is more than just sorting, see https://github.com/artsy/metaphysics/pull/2020. 

I chatted with Sam about what to do here, since it doesn't really make sense to separate the for-sale and not-for-sale artworks (because the new filtered artworks already sort based on merchandisability, which will already put for-sale works at the top). So we're removing the division between "For Sale" and "Other Works" to just have a single list:

Screen Shot 2019-10-25 at 17 04 33

I took a stab at converting the artwork grid tests to use `relay-testing-utils` but got stuck. Might make another try next week.


artsyit:
:rocket: PR was released in v1.18.5 :rocket:

PR #1945, opened 2019-10-30
Partner view tests

kierangillen:
Adds tests for new partner view components

#skip_new_tests


ds300:
:wave: we're looking into using the new relay test utils more and the next step is to try them on a pagination container. Might make sense to do that for this page, if you're down to pair on it some time?

src/lib/Scenes/Partner/Components/__tests__/PartnerArtwork-tests.tsx
```diff
@@ -0,0 +1,43 @@
...
+            }
+          }
+        }
+      `,
+      mockData: {
+        partner,
+      },
+    })
+
+  it("It renders the artworks", async () => {
```
ashfurrow:
```suggestion
  it("renders the artworks", async () => {
```

Note that we'll need to re-do the snapshots after this change. There are also a few others we'll want to change.

src/lib/Scenes/Partner/Components/__tests__/PartnerArtwork-tests.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const wrapper = await getWrapper(PartnerArtworkFixture as any)
+    const grid = wrapper.find(GenericGrid)
+    expect(grid.props().artworks.length).toBe(10)
+    expect(grid.length).toBe(1)
+    expect(wrapper.html()).toMatchSnapshot()
```
ashfurrow: Should this be `.html()` or `.toJSON()`?

kierangillen: We're using `html()` in other instances, which if there are diffs should be easier to track down the code changes. Both presumably would work though to capture any changes.

artsyit:
:rocket: PR was released in v1.18.29 :rocket:

PR #1946, opened 2019-10-31
Prevent concurrent builds on master

ds300:
If we have concurrent builds on master, at least one of the builds will fail due to race conditions in how `auto` works (i.e. adding commits to this repo and pushing to other repos too).

So i searched for a way to prevent concurrent builds on master.

We can configure circle to cancel previous builds on a branch whenever that branch is updated, but that might cancel a build in the middle of a critical deployment path leaving the repo in a weird state.

So what we really want is to make new builds wait for previous builds to finish first. Circle doesn't support this functionality out of the box, but in this support thread https://discuss.circleci.com/t/prevent-concurrent-builds-of-a-branch/1142 they suggest to use this tool https://github.com/bellkev/circle-lock-test but that required setting up `jq` which seemed tricky for us, so I just rewrote the logic in node.

I've tested this on the artsy/emission-ci-test repo.


package.json
```diff
@@ -150,6 +150,7 @@
     "klaw-sync": "6.0.0",
     "lint-staged": "6.1.1",
     "metro-react-native-babel-preset": "0.49.0",
+    "node-fetch": "^2.6.0",
```
alloy: I think these were all pinned by renovate, no?

scripts/await-previous-builds.js
```diff
@@ -0,0 +1,73 @@
...
+
+  if (CIRCLECI !== "true") {
+    throw new Error("awaitPreviousBuilds only works on CI")
+  }
+
+  if (!CIRCLE_TOKEN) {
+    throw new Error(
+      "Please add CIRCLE_TOKEN in env. Must have read access for the project. See https://circleci.com/docs/api/#authentication (make sure it's scope: all)"
+    )
+  }
```
alloy: Do we already expose this token? If not, will it be shared with forks? And, if so, what abilities does it grant and are we ok granting those others?

Also, seeing as this is specifically about doing releases, it seems to me that in reality this env variable would only need to be exposed for `master`, right?

scripts/await-previous-builds.js
```diff
@@ -0,0 +1,73 @@
...
+    CIRCLE_TOKEN,
+  } = process.env
+
+  const res = await fetch(
+    `https://circleci.com/api/v1/project/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}?circle-token=${CIRCLE_TOKEN}&limit=100`,
+    { headers: { Accept: "application/json" } }
+  )
+  if (res.status >= 400) {
+    throw new Error("Failed to fetch circle build data")
+  }
```
alloy: So is this going to end up failing the release? I guess that if it’s a `400` the idea is that this script is doing something wrong [and perhaps outdated per Circle’s latest API]?

scripts/await-previous-builds.js
```diff
@@ -0,0 +1,73 @@
...
+    CIRCLE_BUILD_NUM,
+    CIRCLE_PROJECT_REPONAME,
+    CIRCLE_PROJECT_USERNAME,
+    CIRCLE_TOKEN,
+  } = process.env
+
+  const res = await fetch(
+    `https://circleci.com/api/v1/project/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}?circle-token=${CIRCLE_TOKEN}&limit=100`,
+    { headers: { Accept: "application/json" } }
+  )
```
alloy: Similarly to the comment below, how feasible would it be for exceptions to be thrown from here for temporary issues, in which case we might prefer a retry over a failure as it would fail the release build?

package.json
ds300: yep, my bad, I pinned them again

scripts/await-previous-builds.js
ds300: yes. i updated the error message to mention how to create a project token.

scripts/await-previous-builds.js
ds300: This will be shared with forks, but it's only granting read access to the circle API for this repo, so that's safe to do, I think. See https://circleci.com/docs/2.0/managing-api-tokens/#overview

But yeah this key is only needed on master. Not sure how to share secrets only with particular branches, but it might be possible. Do you think it's worth figuring out?

scripts/await-previous-builds.js
ds300: I think it's no biggie. The developer can manually restart the build if that situation arises, which seems unlikely. Since this runs before any changes are made to the repo, on failures it won't be left in a bad state and manual retries should Just Work™

alloy: Hmm, right, I was thinking of Eigen builds where the dev having to restart a release build actually can be a biggie (due to deadlines/time-constraints).

scripts/await-previous-builds.js
alloy: Well, if it’s only read access then that seems fine to me. Perhaps good to just know in general, but that’s not specific to this PR. /cc @zephraph

alloy:
Aight, let’s :shipit:

artsyit:
:rocket: PR was released in v1.18.12 :rocket:

PR #1947, opened 2019-10-31
Add example of using relay’s mock generator for tests

alloy:
This is a stash of a quick attempt I still had lying around. Let‘s iterate on it.

Unsure if this massive query is the right place to do this. It definitely makes it easy to have the correct fixture data, but since it’s only being verified through a snapshot it’s a little hard to judge the result.

---

Things that could be improved upstream

check the following in 7 and fix if not already taken care of
- fix aliases. make sure you can actually return data for an aliased field from the parent type.
- do sanity check on returned data, to check it matches selections in the query.
- change the 'mock-value-for-field-"blah"' to include path rather than just field name

also
- export more types from react-test-utils/index.d.ts
- prevent passing unused variables in QueryRenderer


ds300:
shouldn't there be a corresponding change to GlobalMap-tests.tsx?

src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
```diff
@@ -101,4 +112,72 @@ describe("Artwork", () => {
 
     expect(tree.root.findAllByType(ContextCard)).toHaveLength(1)
   })
+
+  describe("Live Auction States", () => {
+    it("has the correct state for a work that is in an auction that is currently live, for which I am registered", () => {
```
alloy: The repetitive part of the description can move into a `describe` block:

```js
describe("Live Auction States", () => {
  describe("for a work that is in an auction that is currently live", () => {
    it("has the correct state when I am registered", () => {
      // ...
    })
  })
})
```

src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
```diff
@@ -101,4 +112,72 @@ describe("Artwork", () => {
...
+      const tree = ReactTestRenderer.create(<TestRenderer />)
+
+      environment.mock.resolveMostRecentOperation(operation => {
+        const result = MockPayloadGenerator.generate(operation, {
+          Artwork() {
+            return merge({}, ArtworkFixture, ArtworkFromLiveAuctionRegistrationClosed, RegisteredBidder)
```
alloy: I assume these fixtures have nested data (?), is that re-used by the mock generator for nested field selections?

src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
```diff
@@ -67,20 +67,41 @@ describe("Artwork", () => {
...
+
+    expect(environment.mock.getMostRecentOperation()).toMatchObject({
+      request: {
+        variables: {
+          input: {
+            artwork_id: slug,
+          },
+        },
+      },
+    })
```
alloy: Nice 👌

src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
```diff
@@ -67,20 +67,41 @@ describe("Artwork", () => {
...
   it("refetches on re-appear", () => {
     const tree = ReactTestRenderer.create(<TestRenderer />)
 
+    // resolve initial query
+    expect(environment.mock.getMostRecentOperation().request.node.operation.name).toBe("ArtworkTestsQuery")
+    environment.mock.resolveMostRecentOperation(operation => {
+      return MockPayloadGenerator.generate(operation)
+    })
+
+    // resolve 'viewed artwork' mutation
```
alloy: Probably meant for the code below this assertion?

src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
ds300: Yes.

Btw I'm not happy about using these fixtures here. I left them in for now because I got bogged down in trying to figure out exactly which properties on which types were needed to preserve the test behaviour, and without knowing the underlying auction logic it felt sinister to kinda reverse-engineer the artwork state like that. Might end up with something nonsensical.

src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
```diff
@@ -90,7 +111,10 @@ describe("Artwork", () => {
     tree.update(<TestRenderer isVisible={false} />)
     tree.update(<TestRenderer isVisible={true} />)
 
-    expect(environment.mock.getMostRecentOperation().request.node.operation.name).toBe("ArtworkRefetchQuery")
+    expect(environment.mock.getAllOperations()).toHaveLength(2)
+
+    expect(environment.mock.getAllOperations()[0].request.node.operation.name).toBe("ArtworkRefetchQuery")
+    expect(environment.mock.getAllOperations()[1].request.node.operation.name).toBe("ArtworkMarkAsRecentlyViewedQuery")
```
alloy: The above 3 assertions could be collapsed into 1:

```ts
expect(environment.mock.getAllOperations().map(op => op.request.node.operation.name)).toEqual(["ArtworkRefetchQuery", "ArtworkMarkAsRecentlyViewedQuery"])
```

src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
alloy: Gotcha. Good to know it works, but I agree about not leaving them in long-term.

I think we should get rid of fixtures due to complexity of building them _and_ maintaining them over time and code-changes/upgrades. What are your reasons?

ds300: I'm not too worried about maintaining them over time. Honestly on the BNMO test suite I found that the big artwork/order fixtures saved time when writing new tests and were not much of a problem to maintain. (maybe you disagree having gone through MPv2?)  Most of my problems with fully-fleshed-out fixtures has been making sure IDs + their relationships are consistent and meaningful, which I'm not sure this mocking framework helps with.

The reason I don't want to keep them in here is I want to find a lighter-weight way of composing well-named 'states' of types. e.g. 'artwork which is in an auction that has already begun but the registration is still open and I have not registered for it yet' is not gonna be fun to deal with if it's defined anonymously as overridden type resolvers in every single test. There are just too many fields in play.

alloy: > maybe you disagree having gone through MPv2?

Yeah, it’s been a major time-sink.

src/lib/Scenes/Artwork/__tests__/Artwork-tests.tsx
```diff
@@ -111,8 +111,10 @@ describe("Artwork", () => {
 
     expect(environment.mock.getAllOperations()).toHaveLength(2)
```
alloy: This one is also not needed with the below assertion, but if you prefer to be explicit that’s fine 👍

alloy:
Nice 👌 Let's merge this then and follow-up with a mutation example etc

artsyit:
:rocket: PR was released in v1.18.19 :rocket:

PR #1949, opened 2019-10-31
[ME-77] Marks artworks as recently viewed

ashfurrow:
I checked and [the existing code](https://github.com/artsy/eigen/blob/388d40eb0f1929465285c2f41e7a547c4fa85824/Artsy/View_Controllers/Artwork/ARLegacyArtworkViewController.m#L134) uses a slug, and tested that this works 👍


__mocks__/react-relay.js
```diff
@@ -9,6 +9,6 @@ module.exports = {
   createRefetchContainer: component => component,
   graphql: Relay.graphql,
   commitMutation: (environment, { onCompleted, onError }) => {
-    onCompleted()
+    onCompleted && onCompleted()
```
ashfurrow: `onCompleted` is typed as optional, so this matches that behaviour and allows for mutations to fire-and-forget.

artsyit:
:rocket: PR was released in v1.18.15 :rocket:

PR #1950, opened 2019-11-01
[ME-37] Fixes iOS Share Sheet invocation

ashfurrow:
Couldn't help myself, my curiosity got the better of me.

This bug runs deeper than Airdrop, but that's where we saw it first. Users sharing artworks to other iOS or macOS devices were sending a _text file_, which is *not* QWoA at all. So what was going on?

Our previous work on sharing had us doing a lot of semantic construction before passing the content on to iOS to actually share. This is an anti-pattern; the iOS share API is designed to be given a URL, title, and message so that the OS can massage that data into whatever semantic representation makes sense for the share destination the user picks.

That is to say, instead of computing a string `title` with the artwork title, artist names, and URL all in that string, we should be sending an object with that data to let the OS render however it makes sense.

I've verified this works on-device.

### DIFF
--- a/src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
+++ b/src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
@@ -30,7 +30,7 @@ interface ArtworkActionsProps {
 export const shareContent = (title, href, artists: ArtworkActions_artwork["artists"]) => {
   let computedTitle: string | null
   if (artists && artists.length) {
-    const names = take(artists, 3).map(arist => arist.name)
+    const names = take(artists, 3).map(artist => artist.name)
     computedTitle = `${title} by ${names.join(", ")} on Artsy`
   } else if (title) {
     computedTitle = `${title} on Artsy`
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/Artwork/Components/ArtworkActions.tsx
```diff
@@ -26,22 +27,19 @@ interface ArtworkActionsProps {
...
+  let computedTitle: string | null
   if (artists && artists.length) {
-    const names = []
-    artists.forEach((artist, i) => {
-      if (i < 3) {
-        names.push(artist.name)
-      }
-    })
-    message = `${title} by ${names.join(", ")} on Artsy https://artsy.net${href}`
+    const names = take(artists, 3).map(arist => arist.name)
```
kierangillen: trivial: typo here on `arist => arist.name`

kierangillen:
@ashfurrow Feel free to merge, lgtm!

kierangillen:
And nice job with this fix we would have otherwise overlooked!

artsyit:
:rocket: PR was released in v1.18.21 :rocket:

PR #1951, opened 2019-11-02
Repalce bidAmountMinor with bidAmountCents

yuki24:
Following up on https://github.com/artsy/metaphysics/pull/2021.

We have replaced the `bidAmountCents` with `bidAmountMinor` which is more friendly for other international currencies. This PR replaces the usage of `bidAmountCents` and updates the MP schema accordingly.

#skip_new_tests #trivial


artsyit:
:rocket: PR was released in v1.18.16 :rocket:

PR #1956, opened 2019-11-04
Adds EmptyState, Horizontal Pagination and Artwork Counts to Partner also adds isCloseToEdge util

kierangillen:
- Adds `EmptyState` to each scrollview
- Creates a horizontal pagination container for current and upcoming shows
- Creates a `isCloseToEdge` method for horizontal pagination
- Adds artworks count to header

## Empty state
Screen Shot 2019-11-05 at 12 00 52 PM

#skip_new_tests


src/lib/utils/isCloseToEdge.ts
```diff
@@ -0,0 +1,53 @@
...
+ *      <Artworks />
+ *    </Flatlist>
+ *  )
+ */
+
+const PAGE_END_THRESHOLD = 1000
+
+type CallBack = () => void
+
+export function isCloseToEdge(onScrollEnd: CallBack, pageEndThreshold: number = PAGE_END_THRESHOLD) {
```
ashfurrow: Are we planning to add tests for this class? I'm comfortable merging as-is as long as we ticket the follow-up tests.

kierangillen: We can, looks like there are no tests yet for `isCloseToBottom` so they could be grouped together.

kierangillen: Added a ticket: https://artsyproduct.atlassian.net/browse/ME-99

ashfurrow:
The tests have passed but GitHub hasn't updated the build status. Merging.

artsyit:
:rocket: PR was released in v1.18.22 :rocket:

PR #1958, opened 2019-11-05
[ME-83] Adds empty View Controller for upcoming Search work

ashfurrow:
This is just an empty view. Next steps will be to add a lab option to Eigen to show this instead of the existing search, then we can begin iterating on this. #skip_new_tests


kierangillen:
#mergeongreen

PR #1963, opened 2019-11-12
hide artists list when no artists

ds300:
We found this tiny improvement during knowledge share.

#skip_new_tests


kierangillen:
Just need to add #skip_new_tests i believe

artsyit:
:rocket: PR was released in v1.18.26 :rocket:

PR #1966, opened 2019-11-14
ME-86 Search Autosuggest 

ds300:
https://artsyproduct.atlassian.net/browse/ME-86

search

I ported the text input from palette so we'll probably want to move it there at some point. There are some differences in the typing though so might be we need to create a TextInputIOS component.

#skip_new_tests


data/complete.queryMap.json
```diff
@@ -152,58 +156,5 @@
...
...    url(version: \"square140\")\n    }\n  }\n}\n\nfragment LocationMap_location on Location {\n  id\n  internalID\n  city\n  address\n  address_2: address2\n  postal_code: postalCode\n  summary\n  coordinates {\n    lat\n    lng\n  }\n}\n\nfragment FairBoothPreview_show on Show {\n  slug\n  internalID\n  counts {\n    artworks\n  }\n  partner {\n    __typename\n    ... on Partner {\n      name\n      href\n      slug\n      internalID\n      id\n      profile {\n        id\n        slug\n        internalID\n        isFollowed\n      }\n    }\n    ... on Node {\n      id\n    }\n    ... on ExternalPartner {\n      id\n    }\n  }\n  coverImage {\n    url\n  }\n  location {\n    display\n    id\n  }\n  artworks: artworksConnection(first: 4) {\n    edges {\n      node {\n        ...GenericGrid_artworks\n        id\n      }\n    }\n  }\n}\n\nfragment GenericGrid_artworks on Artwork {\n  id\n  image {\n    aspect_ratio: aspectRatio\n  }\n  ...ArtworkGridItem_artwork\n}\n\nfragment ArtworkGridItem_artwork on Artwork {\n  title\n  date\n  sale_message: saleMessage\n  is_biddable: isBiddable\n  is_acquireable: isAcquireable\n  is_offerable: isOfferable\n  slug\n  sale {\n    is_auction: isA
```
ashfurrow: Seems like quite a few deletions here, is that expected?

src/lib/Scenes/Search/AutosuggestResults.tsx
```diff
@@ -0,0 +1,109 @@
...
+    Sentry.captureMessage(e.stack)
+    console.error(e)
+    return []
+  }
+}
+
+export const AutosuggestResults: React.FC<{ query: string }> = ({ query }) => {
+  const [results, setResults] = useState<AutosuggestResult[]>([])
+  const throttledFetchResults = useMemo(
+    () => throttle(async (q: string) => setResults(await fetchResults(q)), 400, { leading: false, trailing: true }),
```
ashfurrow: Very cool, I didn't know about the `leading`/`trailing` options here 👍

src/lib/Scenes/Search/Input.tsx
```diff
@@ -0,0 +1,114 @@
...
+  background-color: ${props => (props.disabled ? color("black5") : color("white100"))};
+`
+
+const StyledInput: typeof TextInput = styled.TextInput<StyledInputProps>`
+  padding: 0;
+  margin: 0;
+  font-family: ${fontFamily.serif.regular};
+`
+// @ts-ignore
+StyledInput.displayName = "StyledInput"
```
ashfurrow: I'm curious about the story here?

data/complete.queryMap.json
ds300: Yeah. I regenerated the query map from scratch, which is safe to do. The queries removed here are no longer used in emission. The reason they were in there is because relay-compiler will treat this file as append-only, rather than deleting old unused queries. We don't need to rely on that logic because we implemented our own query-preservation logic layer that gets applied only to the metaphysics repo (when deploying a new version of emission).

If there were a relay-compiler option to overwrite the previous file then we'd want to use that in emission, but since there isn't I guess we occasionally have this kind of churn. I'll add that to my list of things to fix in the relay ecosystem ☑️

src/lib/Scenes/Search/Input.tsx
ds300: the `displayName` property is used by React in dev mode to give meaningful names to components in stack traces and inspectors and so forth. I can't imaging us needing a better name for this particular component, but this line was already present in the palette version of this input component and I didn't have the presence of mind to delete it.

src/lib/Scenes/Search/__tests__/AutosuggestResults-tests.tsx
```diff
@@ -0,0 +1,175 @@
...
+}))
+
+jest.unmock("react-relay")
+
+// tslint:disable-next-line:no-empty
+jest.mock("react-native-sentry", () => ({ captureMessage() {} }))
+
+jest.mock("lib/NativeModules/SwitchBoard", () => ({ presentNavigationViewController: jest.fn() }))
+
+const env = defaultEnvironment as ReturnType<typeof createMockEnvironment>
```
ds300: This type exists in @types/relay-test-utils but it's not exported so I've made a note to fix that too.

src/lib/Scenes/Search/AutosuggestResults.tsx
```diff
@@ -0,0 +1,162 @@
...
+      {displayLabel.slice(0, i)}
+      <Serif size="3" weight="semibold">
+        {displayLabel.slice(i, i + query.length)}
+      </Serif>
+      {displayLabel.slice(i + query.length)}
+    </Serif>
+  )
+}
+
+const FadeIn: React.FC<{ delay: number }> = ({ delay, children }) => {
```
ds300: Added a basic fade-in transition for the list items, to give it some pizzazz ✨ 

pizzazz

ds300:
@ash this should be 🍏

src/lib/Scenes/Search/AutosuggestResults.tsx
ashfurrow: I must be getting better at React animations, because this is starting to make sense!

artsyit:
:rocket: PR was released in v1.18.32 :rocket:

PR #1969, opened 2019-11-15
Partner updates

kierangillen:
- Fixes bug where spinners would keep rendering after all content had been loaded
- Changes header to use small `Follow` button in corner
- Uses the correct count for partner locations
- Follow count now updates when follow button clicked

#skip_new_tests

Screen Shot 2019-11-15 at 4 12 36 PM


artsyit:
:rocket: PR was released in v1.18.33 :rocket:

PR #1971, opened 2019-11-19
Collection Header

ashleyjelks:
This PR adds a default header to the iOS Collections landing page.

[Links to GROW-1421](https://artsyproduct.atlassian.net/browse/GROW-1421)

**Header Image When Supplied By Marketing Team:**
Screen Shot 2019-11-19 at 3 05 46 PM

**Default Header Image Based on Most Marketable Artwork in Collection:**
Screen Shot 2019-11-19 at 1 50 07 PM


src/lib/Scenes/Collection/Components/CollectionHeader.tsx
```diff
@@ -1,15 +1,39 @@
...
+import { Box, color, Serif } from "@artsy/palette"
+import OpaqueImageView from "lib/Components/OpaqueImageView/OpaqueImageView"
 import React from "react"
+import { Dimensions } from "react-native"
 import { createFragmentContainer, graphql } from "react-relay"
 
-export const CollectionHeader = props => {
-  const { title, description } = props.collection
+interface CollectionHeaderProps {
+  collection: {
```
ashfurrow: Instead of defining these props manually, we can import the generated types based on the fragment that's defined later on in the file, from the `__generated__` directory. Here's an example:

https://github.com/artsy/emission/blob/97727e4b07b7162fceb07e518ca255366f1dc5c1/src/lib/Scenes/Artwork/Components/ArtworkHeader.tsx#L2

ashleyjelks: Good catch, just updated it!

artsyit:
:rocket: PR was released in v1.18.35 :rocket:

PR #1976, opened 2019-11-21
ME-110 - Partner QA

kierangillen:
- https://artsyproduct.atlassian.net/browse/ME-110
- Fixes all issues brought up in visual QA

Screen Shot 2019-11-21 at 3 07 19 PM

#skip_new_tests


src/lib/Components/StickyTabPage/StickyTabScrollView.tsx
```diff
@@ -14,9 +14,6 @@ export const useStickyTabContext = () => {
   return useContext(StickyTabScrollViewContext)
 }
 
-// how fast you have to be scrolling up to show the header when not near the top of the scroll view
-const SHOW_HEADER_VELOCITY = 10
```
ashfurrow: This idea didn't pan out, or?

src/lib/Scenes/Partner/Components/PartnerFollowButton.tsx
```diff
@@ -78,7 +78,12 @@ export class PartnerFollowButton extends React.Component<Props, State> {
     )
   }
 
-  handleShowSuccessfullyUpdated() {
+  handleShowSuccessfullyUpdated(partnerFollowed) {
+    const { setFollowersCount, followersCount } = this.props
+    if (setFollowersCount && followersCount) {
+      partnerFollowed ? setFollowersCount(followersCount - 1) : setFollowersCount(followersCount + 1)
```
ashfurrow: Nice catch!

src/lib/Scenes/Partner/Components/PartnerShows.tsx
```diff
@@ -12,34 +13,42 @@ import { PartnerShowsRailContainer as PartnerShowsRail } from "./PartnerShowsRai
...
+class ShowGridItem extends React.Component<ShowGridItemProps> {
+  onPress = () => {
+    const { show } = this.props
+    SwitchBoard.presentNavigationViewController(this, `/show/${show.slug}`)
+  }
+
+  render() {
+    const { show, itemIndex } = this.props
+    const showImageURL = show.coverImage && show.coverImage.url
+    const styles = itemIndex % 2 === 0 ? { paddingRight: space(1) } : { paddingLeft: space(1) }
```
ashfurrow: I know this is unrelated to this change, but I recall discussing (with @alloy?) that we prefer to stick with margins, rather than padding. Or I might be mis-remembering?

src/lib/Components/StickyTabPage/StickyTabScrollView.tsx
kierangillen: It's better if it just appears when the user comes back to the top of the view.

src/lib/Scenes/Partner/Components/PartnerShows.tsx
kierangillen: In some cases they can be used interchangeably as long as the padding exists on the wrapper, which is the case here. Having a blanket rule about using padding instead of margin or vice versa seems problematic, they both have their uses and I think it can just come down to personal preference.

ashfurrow: Cool, thanks for the explainer :bow:

artsyit:
:rocket: PR was released in v1.18.37 :rocket:

src/lib/Scenes/Partner/Components/PartnerShows.tsx
alloy: What we discussed is that components should not define whitespace around themselves, but leave that up to the parent composing it. Otherwise you inevitably end-up battling styles.

PR #1977, opened 2019-11-21
ME-108 - Updates Artist page to use new design system and spec

kierangillen:
- https://artsyproduct.atlassian.net/browse/ME-108

#skip_new_tests


src/lib/AppRegistry.tsx
```diff
@@ -74,13 +74,9 @@ interface ArtistProps {
...
-const Artist: React.SFC<ArtistProps> = track<ArtistProps>(props => {
-  return {
-    context_screen: Schema.PageNames.ArtistPage,
-    context_screen_owner_slug: props.artistID,
-    context_screen_owner_type: Schema.OwnerEntityTypes.Artist,
-  }
-})(props => <ArtistRenderer {...props} render={renderWithLoadProgress(Containers.Artist, props)} />)
+const Artist: React.SFC<ArtistProps> = props => (
+  <ArtistRenderer {...props} render={renderWithLoadProgress(Containers.Artist, props)} />
+)
```
ashfurrow: Is this change, to remove tracking, intended?

src/lib/Components/Artist/ArtistHeader.tsx
```diff
@@ -33,20 +30,39 @@ class Header extends React.Component<Props, State> {
...
+      <Box px={2} pt={3}>
+        <Serif style={{ textAlign: "center" }} size="5">
+          {artist.name}
+        </Serif>
+        <Spacer mb={0.5} />
+        {(count || bylineRequired) && (
+          <>
+            <TextWrapper style={{ textAlign: "center" }}>
+              {bylineRequired && <Sans size="2">{this.descriptiveString()}</Sans>}
+              {count &&
```
ashfurrow: If `count` is non-nil, but `bylineRequired` is, then this could cause a crash because the expression will be evaluated as `count`, I think? We should add `!!` in front of non-boolean variables when we want to explicitly evaluate them as booleans.

src/lib/AppRegistry.tsx
kierangillen: Yeah I moved the screentrack here https://github.com/artsy/emission/pull/1977/files#diff-3a6742a6612df63240e3e2279fb6fc86R22 so that I could also track both the `slug` and `internalID`

ashfurrow: Nice 👍

src/lib/Components/Artist/ArtistHeader.tsx
kierangillen: Cool, just added some coercion here.

artsyit:
:rocket: PR was released in v1.18.40 :rocket:

alloy:
Great work @kierangillen 👏

PR #1981, opened 2019-11-26
[MX-114] Infrastructure for hiding back button on Partner view

ashfurrow:
#skip_new_tests.

### DIFF
--- a/src/lib/Components/StickyTabPage/StickyTabPage.tsx
+++ b/src/lib/Components/StickyTabPage/StickyTabPage.tsx
@@ -1,7 +1,7 @@
 import { Box, color, Flex, Sans, space, Spacer } from "@artsy/palette"
 import { Schema } from "lib/utils/track"
 import { useScreenDimensions } from "lib/utils/useScreenDimensions"
-import React, { useRef, useState } from "react"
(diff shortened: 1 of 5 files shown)


src/lib/Components/StickyTabPage/StickyTabPage.tsx
```diff
@@ -34,6 +35,22 @@ export const StickyTabPage: React.FC<{
...
+  const viewControllerRef = useRef(null)
+
+  const shouldHideBackButton = Animated.lessOrEq(headerOffsetY, -10)
+
+  Animated.useCode(
+    () =>
+      Animated.onChange(
+        shouldHideBackButton,
+        Animated.call([shouldHideBackButton], ([shouldHide]) => {
+          if (viewControllerRef.current) {
```
ashfurrow: Ah, gotcha. So since we don't need the reference to `self` as a `UIViewController`, do we need the `viewControllerRef` at all?

ds300: oh nope, good catch

artsyit:
:rocket: PR was released in v1.19.2 :rocket:

PR #1982, opened 2019-11-26
StickyTabs perf + UX improvements

ds300:
The most critical thing that's changed here is that we've now implemented the sticky tab page in terms of `FlatList`s which obviously gives us all the lazy rendering goodness involved in that. So far I've converted one page to use it, the `PartnerShows` page.

This is a big diff but the individual commits are fairly discrete and should make for a fun read in order.

Some follow up items:

- Investigate debouncing onEndReached (seems to get called multiple times in one render cycle of a page, which means that it submits multiple fetch requests 😬)
- Refactor other pages with potentially many items to use the flat list (Artworks on artist + partner pages, Overview on partner page)
- Investigate building the loading spinner into the page

#skip_new_tests


src/lib/Components/Artist/ArtistAbout.tsx
```diff
@@ -18,11 +19,11 @@ interface Props {
 class ArtistAbout extends React.Component<Props> {
   render() {
     return (
-      <Box py={3} px={2}>
```
ds300: The `StickyTabPageFlatList/ScrollView` has the standard padding built-in now.

src/lib/Components/StickyTabPage/SnappyHorizontalRail.tsx
```diff
@@ -1,29 +1,44 @@
...
+      setOffset(offset) {
+        if (currentAnimation.current) {
+          currentAnimation.current.stop()
+        }
+        currentAnimation.current = Animated.spring(currentOffset, {
+          ...Animated.SpringUtils.makeDefaultConfig(),
+          stiffness: 600,
+          damping: 120,
+          toValue: -offset,
+        })
```
ds300: Changed this to be controlled imperatively to avoid the react render cycle and give users *instant* feedback.

src/lib/Components/StickyTabPage/SnappyHorizontalRail.tsx
```diff
@@ -1,29 +1,44 @@
...
+    }),
+    []
   )
 
   return (
     <Animated.View
       style={{
         flex: 1,
-        width: width * 3,
+        width,
```
ds300: fixed this hard-coded value (was assuming 3 tabs)

src/lib/Components/StickyTabPage/StickyTabPage.tsx
```diff
@@ -30,33 +30,31 @@ export const StickyTabPage: React.FC<{
   headerContent: JSX.Element
 }> = ({ tabs, headerContent }) => {
   const { width } = useScreenDimensions()
-  const [activeTabIndex, setActiveTabIndex] = useState(Math.max(tabs.findIndex(tab => tab.initial), 0))
-  const [headerHeight, setHeaderHeight] = useState<null | number>(null)
+  const initialTabIndex = useMemo(() => Math.max(tabs.findIndex(tab => tab.initial), 0), [])
+  const activeTabIndex = useAnimatedValue(initialTabIndex)
+  const [headerHeight, setHeaderHeightNativeWrapper] = useState<Animated.Value<number>>(null)
```
ds300: moved all this state into reanimated to avoid re-renders when changing tabs.

src/lib/Components/StickyTabPage/StickyTabPage.tsx
```diff
@@ -71,58 +69,31 @@ export const StickyTabPage: React.FC<{
...
+          initialActiveIndex={initialTabIndex}
+          onIndexChange={index => {
+            activeTabIndex.setValue(index)
+            railRef.current.setOffset(index * width)
+            tracking.trackEvent({
+              action_name: tabs[index].title,
+              action_type: Schema.ActionTypes.Tap,
+            })
+          }}
+        />
```
ds300: Moved the tab bar into a separate file and reduced the coupling with this file 💆

src/lib/Components/StickyTabPage/StickyTabPageFlatList.tsx
```diff
@@ -0,0 +1,230 @@
...
+  /* key must be unique per-tab */
+  key: string
+  content: JSX.Element
+}
+export interface StickyTabFlatListProps
+  extends Omit<FlatListProps<any>, "onScroll" | "data" | "scrollEventThrottle" | "ListHeaderComponent" | "renderItem"> {
+  data: StickyTabSection[]
+}
+
+export const StickyTabPageFlatList: React.FC<StickyTabFlatListProps> = props => {
```
ds300: github shows this as a new file, but it's really a modification of the previous ScrollView file.

src/lib/Components/StickyTabPage/StickyTabPageFlatList.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+  const flatListRef = useRef<{ getNode(): FlatList<any> }>()
+
+  const lastIsActive = useAnimatedValue(-1)
+
+  // prevent this tab from manipulating the header position when it is not active
+  Animated.useCode(
+    () =>
+      // when the active state changes
+      Animated.cond(Animated.neq(lastIsActive, isActive), [
```
ds300: I tried `Animated.onChange(isActive)` here but it didn't work on one of the three tabs 🤔 a little worrying.

src/lib/Components/StickyTabPage/StickyTabPageFlatList.tsx
```diff
@@ -0,0 +1,230 @@
...
+                    "Please make sure that tab content is wrapped with a StickyTabPageFlatList or a StickyTabPageScrollView"
+                  )
+                }
+                flatListRef.current.getNode().scrollToOffset({ offset: -y, animated: false })
+                lockHeaderPosition.setValue(0)
+              }),
+              Animated.set(lockHeaderPosition, 0)
+            ),
+          ],
+          Animated.set(lockHeaderPosition, 1)
```
ds300: Before this was running in JS land in a `useEffect` while now the `isActive` state is in reanimated so it has to run on the native side. Same logic though.

src/lib/Components/StickyTabPage/StickyTabPageFlatList.tsx
```diff
@@ -0,0 +1,230 @@
...
+            notBouncingAtTheTop,
+            [
+              // bouncing at the bottom, keep the header fully hidden
+              Animated.set(headerOffsetY, negative(headerHeight)),
+            ],
+            [
+              // bouncing at the top, keep the header fully open
+              Animated.set(headerOffsetY, 0),
+            ]
+          ),
```
ds300: This is the only thing that really changed here. It fixes the error you were seeing where too much padding was appearing at the top.

src/lib/Scenes/Partner/Components/PartnerShows.tsx
```diff
@@ -55,62 +54,90 @@ export const PartnerShows: React.FC<{
...
+
+  const sections: StickyTabSection[] = useMemo(
+    () => {
+      if (!pastShows && !hasRecentShows) {
+        return [
+          {
+            key: "empty_state",
+            content: <TabEmptyState text="There are no shows from this gallery yet" />,
+          },
+        ]
```
ds300: `FlatList` actually has an `EmptyStateComponent` prop so we could use that and return an empty array here instead. no biggie though

src/lib/Scenes/Partner/Components/PartnerShows.tsx
```diff
@@ -55,62 +54,90 @@ export const PartnerShows: React.FC<{
...
+        setIsLoadingMore(true)
+        relay.loadMore(PAGE_SIZE, error => {
+          if (error) {
+            // FIXME: Handle error
+            console.error("PartnerShows.tsx", error.message)
+          }
+          setIsLoadingMore(false)
+        })
+      }}
+      refreshing={isLoadingMore}
```
ds300: Need to look into what this prop does. Doesn't appear to do anything.

src/lib/Scenes/Partner/Components/PartnerShows.tsx
```diff
@@ -55,62 +54,90 @@ export const PartnerShows: React.FC<{
...
-                </Flex>
-              </Box>
-            )}
-            <Spacer mb={3} />
-          </>
-        )}
-    </Box>
+    <StickyTabPageFlatList
+      data={sections}
+      onEndReachedThreshold={0.5}
```
ds300: this is a proportion of the screen height, apparently. so 0.5 is half the height of the screen.

src/lib/Scenes/Partner/Components/PartnerShows.tsx
```diff
@@ -55,62 +54,90 @@ export const PartnerShows: React.FC<{
...
+      onEndReached={() => {
+        if (isLoadingMore || !relay.hasMore()) {
```
ds300: Like i said in the PR description, we need to look into a different way to debounce these because FlatList calls these callbacks multiple times before the react render cycle has caused `isLoadingMore` to be `true`

src/lib/Components/StickyTabPage/SnappyHorizontalRail.tsx
```diff
@@ -1,22 +1,40 @@
...
-        stiffness: 600,
-        damping: 120,
-        toValue: -offset,
-      }).start()
-    },
-    [offset]
+  const currentOffset = useAnimatedValue(-initialOffset)
+  const currentAnimation = useRef<Animated.BackwardCompatibleWrapper>()
+
+  useImperativeHandle(
```
ashfurrow: Oh _what_, this is cool, didn't know about this hook.

src/lib/Components/StickyTabPage/StickyTabScrollView.tsx
```diff
@@ -13,6 +14,8 @@ export const useStickyTabContext = () => {
   return useContext(StickyTabScrollViewContext)
 }
 
+const AnimatedFlatList: typeof FlatList = Animated.createAnimatedComponent(FlatList)
```
ashfurrow: Oh nice, is this a common pattern?

ds300: Yep. I'm unsure how it works under the hood but it's supported behaviour for the core RN components at least.

src/lib/Scenes/Partner/Components/PartnerShows.tsx
```diff
@@ -114,10 +120,17 @@ export const PartnerShows: React.FC<{
     [partner.pastShows.edges, hasRecentShows, pastShows]
   )
 
+  const tabContext = useContext(StickyTabPageFlatListContext)
+
+  const tabIsActive = Boolean(useNativeValue(Animated.eq(tabContext.activeTabIndex, tabContext.tabIndex), 0))
```
ds300: might make sense to expose this tabIsActive value in the context rather than the index values. Not sure how useful they are in any other situation.

artsyit:
:rocket: PR was released in v1.19.5 :rocket:

PR #1983, opened 2019-12-02
Partner artworks to InfiniteScrollGrid

kierangillen:
- Uses `InfiniteScrollGrid` for the `Partner.artworks` which prepares us to move `InfiniteScrollGrid` to `FlatList` to prevent nested `ScrollViews`

#skip_new_tests


artsyit:
:rocket: PR was released in v1.19.2 :rocket:

PR #1985, opened 2019-12-03
Show sticky header based on scroll offset rather than velocity

ds300:
#skip_new_tests


artsyit:
:rocket: PR was released in v1.19.3 :rocket:

PR #1995, opened 2019-12-06
[MX-54] App Ratings Prompt

ashfurrow:
This PR adds support for the native iOS app rating dialogue: 

Screen Shot 2019-12-06 at 14 19 14

(It will be the _actual_ app automatically once integrated into Eigen.)

I've documented the logic for how often we ask for the review, as well as a quirky new Emission constant. [Apple's documentation on this API](https://developer.apple.com/documentation/storekit/skstorereviewcontroller/2851536-requestreview?language=objc) is sparse, to say the least. I found contradictory reports on how it works under the hood, but [this post](https://medium.com/@kavithakumarasamy89/skstorereviewcontroller-apple-way-to-request-review-and-rating-inside-ios-app-in-ios-10-3-453a6f897e9d) was most useful.

Since the API is iOS 11+ anyway, I increased the minimum deployment target of the Emission CocoaPod/Example app to iOS 12 (to match Eigen's). I've split the work into separate commits so that they're easier to review. 

This will require a matching PR on Eigen's side. #skip_new_tests


src/lib/NativeModules/Events.tsx
```diff
@@ -8,4 +8,25 @@ function postEvent(info: any) {
...
+  // space out how often we ask for a rating so users have gotten a sense of the app's value.
+  if (launchCount % 20 === 2) {
+    if (!hasRequested) {
+      hasRequested = true
+      AREventsModule.requestAppStoreRating()
+    }
+  }
+}
+// Whether we have requested during the current session or not.
+let hasRequested = false
```
kierangillen: This will get hoisted, but for legibility maybe add it above the function?

kierangillen:
Looks good to me, merge when ready.

package.json
```diff
@@ -45,7 +45,7 @@
     "update-metaphysics-and-eigen": "node scripts/update-metaphysics-and-eigen.js",
     "sync-schema": "curl https://raw.githubusercontent.com/artsy/metaphysics/master/_schemaV2.graphql -o data/schema.graphql; yarn prettier --write --parser graphql data/schema.graphql",
     "sync-cities": "curl https://raw.githubusercontent.com/artsy/metaphysics/master/src/schema/city/cityDataSortedByDisplayPreference.json -o data/cityDataSortedByDisplayPreference.json",
-    "version": "yarn install && yarn bundle && pushd Example && bundle install && bundle exec pod update Emission --no-repo-update && popd && git add Pod/Assets Example/Podfile.lock",
+    "version": "yarn install --ignore-engines && yarn bundle && pushd Example && bundle install && bundle exec pod update Emission --no-repo-update && popd && git add Pod/Assets Example/Podfile.lock",
```
ashfurrow: /cc @ds300

ashfurrow: /cc @ds300

ashfurrow:
This is failing to build because we're specifying a newer macOS machine version that doesn't have the Ruby version specified on `master` in `.ruby-version` (2.5.3). But when Circle's macOS VM clones the git repo, it has the `master` branch checked out by default. `chruby` then tries to use that version, which is not installed, so the "Checkout code" step of our CI is failing. 

I'm going to merge and monitor on master, and fix from there.

artsyit:
:rocket: PR was released in v1.19.7 :rocket:

ashfurrow:
Following up, this ended up working on master.

PR #1998, opened 2019-12-09
Run unit tests on master

ds300:
@ashfurrow noticed we were not running unit tests on master 😬 https://github.com/artsy/emission/commit/27129edb1d4aef158f1cfdd8de19f165e6d7c137#commitcomment-36311454

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -181,11 +181,7 @@ workflows:
   version: 2
   build-test:
     jobs:
-      - build-and-test:
-          filters:
-            branches:
-              ignore:
-                - master
+      - build-and-test
       - deploy:
           filters:
             branches:


ds300:
@ashfurrow actually unit tests are running on travis on master. Which begs the question: why do we run unit tests both on travis and circle for PRs? Why do we use both travis and circle in the first place?

ashfurrow:
@ds300 that's a good question! We have two unit test suites in this repo: one for native Objective-C code, and the other for our Jest tests. There are only four Objective-C unit tests, so the native CI is more to make sure that the app continues to compile than actually execute correctly. The benefit of isolating our Jest tests is that we avoid a lot of CocoaPods setup (though, since we've moved to the CocoaPods CDN, that benefit has shrunk a lot, so it might make sense to consolidate).

ds300:
ahhh of course, it's native vs JS.

Cool, well, then let's merge this now, and I'd argue there's not much reason to worry about consolidation since when the rearchitecture work kicks in we'll be doing a big pile of CI consolidation.

artsyit:
:rocket: PR was released in v1.19.8 :rocket:

PR #2001, opened 2019-12-09
[GROW-1423]: add featured artists component

xtina-starr:
Addresses: [GROW-1423](https://artsyproduct.atlassian.net/browse/GROW-1423)

This adds the featured artists component for the collections page in iOS. I wanted to get some feedback on what I have so far.


src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
```diff
@@ -0,0 +1,118 @@
+import { Box, Button, EntityHeader, Flex, Sans } from "@artsy/palette"
+import SwitchBoard from "lib/NativeModules/SwitchBoard"
+import React from "react"
+import { TouchableHighlight, TouchableWithoutFeedback } from "react-native"
+import { createFragmentContainer, graphql } from "react-relay"
+
+interface FeaturedArtistsProps {
+  featuredArtists: any
```
ashfurrow: I think this should be typed as `FeaturedArtists_featuredArtists` instead of `any`, right? You can import it from the `__generated__` directory.

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
```diff
@@ -0,0 +1,118 @@
...
+export class FeaturedArtists extends React.Component<FeaturedArtistsProps, FeaturedArtistsState> {
+  state = {
+    showMore: false,
+  }
+
+  handleTap = (context: any, href: string) => {
+    SwitchBoard.presentNavigationViewController(context, href)
+  }
+
+  getFeaturedArtistEntityCollection = (artists: any) => {
```
ashfurrow: We try not to use `any` type much, since it introduces a lot of uncertainty into the code. Since Emission code can't be rolled back if there is a bug, it's extra-important to avoid bugs, and our types help us with that 👍 In this case, there are two options. I'll outline them both, and then discuss which one I recommend.

- Since this is `merchandisableArtists ` used in the fragment below, once we change the `featuredArtists` prop to be `FeaturedArtists_featuredArtists`, then we can move from `any` to `FeaturedArtists_featuredArtists["artworksConnection "]["merchandisableArtists"]`. 
- `getFeaturedArtistEntityCollection()` is only used once, so instead of defining this logic in its own function, we can inline inside the `render()` function. This means we don't need to explicitly type `artists` at all, since it'll be inferred from the `featuredArtists` prop.

I recommend the second approach because it's less explicit typing. In general, I recommend a longer `render()` function that doesn't call out to single-purpose functions like this, since the compiler's type inference will handle most things for you. This is especially a good idea for this function, since it doesn't really have any logic in it – we could replace the call to `this.getFeaturedArtistEntityCollection(artists)` with the `artists.map(...` call in this function and get the same behaviour with less code/types.

Let me know what I can clarify here 👍

pepopowitz: For what it's worth, I like having functions broken out of `render`, because it helps me understand everything that's happening in it in one quick pass.

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
```diff
@@ -0,0 +1,118 @@
...
+    return artists.map((artist, index) => {
+      const hasArtistMetaData = artist.nationality && artist.birthday
+      return (
+        <Box width="100%" key={index} pb={20}>
+          <TouchableWithoutFeedback onPress={() => this.handleTap(this, `/artist/${artist.slug}`)}>
```
kierangillen: No need to create the `handleTap` method when the Switchboard call can be done inline here: ` onPress={() => SwitchBoard.presentNavigationViewController(this, `/artist/${artist.slug}`)}`

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
ashfurrow: That's totally valid, too! I mainly prefer fewer single-use functions because it's less explicit types to write out. Also, it kind of encourages me to refactor a component into smaller components (if the render function is getting too larger). 

In any case, I really only care strongly about the `any` 👍

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
```diff
@@ -0,0 +1,118 @@
...
+            />
+          </TouchableWithoutFeedback>
+        </Box>
+      )
+    })
+  }
+
+  render() {
+    const { featuredArtists } = this.props
+    const artists = featuredArtists.artworksConnection.merchandisableArtists
```
kierangillen: This would be better if we could `import { get } from "lib/utils/get"` and then:

```const artists = get(this.props, p => p.featuredArtists.artworksConnection.merchandisableArtists)
const hasMultipleArtists = artists && artists.length > 1```

File: src/lib/Scenes/Collection/Components/FeaturedArtists.tsx [hunk 5]
```diff
@@ -0,0 +1,118 @@
...
+    const hasMultipleArtists = artists.length > 1
+    const artistCount = 3
+    const remainingCount = artists.length - artistCount
```
kierangillen: `artists` could be `null` here so we would want to guard for that case here.

kierangillen: Maybe just `if(!artists){ return null }`

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
```diff
@@ -0,0 +1,118 @@
...
+  getFeaturedArtistEntityCollection = (artists: any) => {
+    return artists.map((artist, index) => {
+      const hasArtistMetaData = artist.nationality && artist.birthday
+      return (
+        <Box width="100%" key={index} pb={20}>
+          <TouchableWithoutFeedback onPress={() => this.handleTap(this, `/artist/${artist.slug}`)}>
+            <EntityHeader
+              imageUrl={artist.imageUrl}
+              name={artist.name}
+              meta={hasArtistMetaData ? `${artist.nationality}, b. ${artist.birthday}` : undefined}
```
kierangillen: I'd check the typing on `meta` here, I'd assume we'd probably want to pass `null` or `""` depending on its type instead of `undefined`

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
ashleyjelks: Here's an example of that for reference! https://github.com/artsy/emission/blob/3adc3a7803fe5091bef60577eed9869ec26c1c31/src/lib/Scenes/Collection/Components/CollectionHeader.tsx#L2

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
ashleyjelks: @ashfurrow is it correct to say that one of these ways is more commonly found in iOS dev than the other??

ashleyjelks:
@xtina-starr I'm thinking about file structure for Collections and how we can maintain some order/organization around it that won't require complex refactors/moving around of files down the road.

I think Collections should follow the same organization and file naming conventions as the other components in Emission.

Can we re-name the files here to start with `CollectionComponentName.tsx` and also move them to this folder `src/lib/Scenes/Collection/Screens` e.g. `src/lib/Scenes/Collection/Screens/CollectionFeaturedArtists.tsx`??

Since we already have CollectionHeader.tsx and CollectionHeader.test.tsx I'll make a small PR  with the new file re-orgnization  so it can be merged and any new development of Collection compents can be added to the `src/lib/Scenes/Collection/Screens` directory.

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
ashfurrow: @ashleyjelks That's a good question. I'm not sure it's correct to say generally, or even just within Emission. But after a few years of working with React+TypeScript, it is my recommendation based on the trade-offs (it's not a super-strongly-held opinion, though).

My primary concern is typing things as `any`. So as long as we're avoiding those, I'm happy 👍 This is a pretty subtle discussion, and I'm happy to hop on a call to discuss/pair more – just let me know!

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
pepopowitz: [Updated](https://github.com/artsy/emission/pull/2001/files#diff-cf72bf49360aa2b6be4c6407c4b8ce68R56)! 

```
    const artists = get(this.props, p => p.collection.artworksConnection.merchandisableArtists)
    if (!artists) {
      return null
    }

    const hasMultipleArtists = artists.length > 1
```

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
pepopowitz: [Updated](https://github.com/artsy/emission/pull/2001/files#diff-cf72bf49360aa2b6be4c6407c4b8ce68R56)! 

```
    const artists = get(this.props, p => p.collection.artworksConnection.merchandisableArtists)
    if (!artists) {
      return null
    }

    const hasMultipleArtists = artists.length > 1
```

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
```diff
@@ -0,0 +1,116 @@
...
+    fragment FeaturedArtists_collection on MarketingCollection {
+      # TODO: size:9 is not actually limiting to 9 items. We need to figure out
+      #  why the back-end is not respecting that argument.
+      artworksConnection(aggregations: [MERCHANDISABLE_ARTISTS], size: 9, sort: "-decayed_merch") {
+        merchandisableArtists {
+          slug
+          internalID
+          name
+          image {
+            resized(width: 100) {
```
pepopowitz: @ashfurrow @kierangillen We noticed while writing tests that we were using the full `imageUrl`. Having been bitten fairly recently by unresized images on web, we thought it would be appropriate to resize here also. Is there any reason not to resize, and is the width we chose reasonable for an entity header image on mobile?

xtina-starr: bump @ashfurrow.

src/lib/Scenes/Collection/Components/__tests__/FeaturedArtists-tests.tsx
```diff
@@ -0,0 +1,93 @@
+import { Theme } from "@artsy/palette"
+import { FeaturedArtists_collection } from "__generated__/FeaturedArtists_collection.graphql"
+import { renderRelayTree } from "lib/tests/renderRelayTree"
+import React from "react"
+import { graphql } from "react-relay"
+import { CollectionFeaturedArtistsContainer as FeaturedArtists } from "../FeaturedArtists"
+
+jest.unmock("react-relay")
+
+const CollectionFixture: FeaturedArtists_collection = {
```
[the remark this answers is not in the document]
xtina-starr: Can this be addressed in a subsequent PR? Perhaps when we tackle https://artsyproduct.atlassian.net/browse/GROW-1717 and https://artsyproduct.atlassian.net/browse/GROW-1716.

src/lib/Scenes/Collection/Components/FeaturedArtists.tsx
ashfurrow: Thanks for the bump! I'm curious about the undersized images issue, but let me answer your questions:

> Is there any reason not to resize?

None that I can see. The `resized` resolver isn't used commonly in Emission yet, but I think that's because most of Emission predates this MP resolver.

> is the width we chose reasonable for an entity header image on mobile?

I don't think it's sufficient. The narrowest iPhone is 320 points wide (640 actual pixels). I would say 500 would be a better value here. 

_Ideally_, we would pass in the actual screen width by importing `Dimensions` from the `react-native` module and use `Dimensions.get("window").width`. This could be passed in to the fragment with an argument with `@arguments` wherever `FeaturedArtists_collection` is used. This way, the image would match exactly the screen size. This would also account for devices like iPad that have really high screen widths. Let me know if I can clarify/help/pair.

src/lib/Scenes/Collection/Components/__tests__/FeaturedArtists-tests.tsx
```diff
@@ -0,0 +1,93 @@
...
+    ],
+  },
+  " $refType": null,
+}
+
+describe("FeaturedArtists", () => {
+  const render = () =>
+    renderRelayTree({
+      Component: ({ marketingCollection }) => (
+        <Theme>
```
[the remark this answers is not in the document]
xtina-starr: If memory serves me correctly without the theme we were getting style component errors.

pepopowitz:
Closing in favor of #2032, which rebased & moved this work forward.

PR #2002, opened 2019-12-09
Fix sticky tab header positioning when content is not so high

ds300:
@kierangillen This fixes that issue you saw during QA

Might want to view the diff with whitespace changes turned off, looks like the recent prettier upgrade changed the indentation of this whole block.

#skip_new_tests


kierangillen:
👍

artsyit:
:rocket: PR was released in v1.19.12 :rocket:

PR #2004, opened 2019-12-09
Points to our fork of tipsi-stripe #trivial

ashfurrow:
We need to point to point to our fork of `tipsi-stripe` because we pulled in [this fix](https://github.com/tipsi/tipsi-stripe/pull/508) for [this issue](https://github.com/tipsi/tipsi-stripe/issues/408) (it'll be included in an upcoming release of `tipsi-stripe`, so I'm fine still pointing to my own fork). Previously, we did this in Eigen, but was removed in the RN upgrade infrastructure PR:

https://github.com/artsy/eigen/pull/2962/files#diff-4a25b996826623c4a3a4910f47f10c30L101

This was causing the bug (app crashes when entering credit card number) to re-occur.

This adds it back, using that new infrastructure mechanism. I took a look at the other RN pods that we were pointing to forks on, and it appears all the changes we'd made are now included in their upstreams, so we're good to go 🙌

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -79,11 +79,6 @@ PODS:
   - Folly (2018.10.22.00):
     - boost-for-react-native
     - DoubleConversion
-    - Folly/Default (= 2018.10.22.00)
-    - glog
-  - Folly/Default (2018.10.22.00):
-    - boost-for-react-native
-    - DoubleConversion
     - glog
   - glog (0.3.5)
   - INTUAnimationEngine (1.4.2):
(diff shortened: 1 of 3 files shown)


ashfurrow:
Took off the MOG label to apply this change first: 

https://github.com/ashfurrow/tipsi-stripe/commit/aeb742baf1a8329ba371d9c94c7650104a112807 

We were pointing Yarn to my fork, but it was pointing CocoaPods back to the main source 😭 I verified the code change was included in the Objective-C this time (the bug doesn't present in the Emission Example app).

artsyit:
:rocket: PR was released in v1.19.11 :rocket:

PR #2006, opened 2019-12-11
Moves collection screens to new a sub dir of collections

ashleyjelks:
This PR adds a Screens subdirectory under Collections so that Collections components and views in Emission are organized similarly to other Screens (e.g. Show, Fair, Artwork...).


artsyit:
:rocket: PR was released in v1.19.16 :rocket:

PR #2007, opened 2019-12-11
MX-117 Retry failed persisted queries with full text

ds300:
https://artsyproduct.atlassian.net/browse/MX-117

We often find that the beta builds fail to work on production because a query map update hasn't been deployed to metaphysics production yet.

This fixes that by retrying requests that failed due to a 404 on the query ID. On the retry request we pass the full query text to MP instead of the query ID.

Note that this won't help with query changes that depend on additions or fixes to MP.

I tested this on the artwork page.


ds300:
@ashfurrow  great suggestion! It totally slipped my mind. 🍏

artsyit:
:rocket: PR was released in v1.19.15 :rocket:

PR #1999, opened 2019-12-09
Adding more tests for search UI

ds300:
Adding tests as follow-up for #1994 

#skip_new_tests

PR #2008, opened 2019-12-11
Updates Danger rule for adding tests for new files

ashfurrow:
I was reviewing #1999 and noticed the "skip new tests" hashtag to avoid the Danger rule. We've been doing that a lot, and it got me thinking about the purpose of the rule itself.

According to the comments in the Dangerfile, the rule is really intended only for _new files_, but _modified files_ were being included in the check too. A comment above alludes to the fact that they're most interchangeable – but in my opinion, the check for new tests is not one of those times.

I've changed the Dangerfile to only check for new files, while leaving existing variables so that future rules can use them.

After merging this PR, the new-tests rule won't triggered spuriously, so we will all need to adjust our barometers for when it makes sense to add "skip new tests" to our PR descriptions 👍


artsyit:
:rocket: PR was released in v1.19.13 :rocket:

PR #2011, opened 2019-12-13
[MX-75] fix scroll to bottom in conversations

ds300:
https://artsyproduct.atlassian.net/browse/MX-75

The problem was just that we were calculating the height of the flatlist in a really naive way.


artsyit:
:rocket: PR was released in v1.19.19 :rocket:

PR #2015, opened 2019-12-17
Adds collection artwork preview component

ashleyjelks:
This PR adds the artwork preview grid to the collections view.

[Links to GROW-1473](https://artsyproduct.atlassian.net/browse/GROW-1473)

We have a mob pairing session scheduled to work on adding tests, but I think this PR is at least in a good state to have the code reviewed for the component itself.

Kapture 2019-12-17 at 15 34 43

#skip_new_tests


ashleyjelks:
> It looks like the actual CollectionArtworkPreview component is not included in this PR. Is that accidental, or because it is a component that already existed?

Definitely not intentional!

src/lib/Scenes/Collection/Screens/CollectionArtworkPreview.tsx
```diff
@@ -0,0 +1,40 @@
...
+  collection: CollectionArtworkPreview_collection
+}
+
+export class CollectionArtworkPreview extends React.Component<Props> {
+  render() {
+    const { collection } = this.props
+    if (!collection) {
+      return null
+    }
+    const artworks = collection.artworks.edges.map(({ node }) => node)
```
pepopowitz: Should we use the `get` util against this long chain of properties, to avoid app crashes if for some reason something came back from metaphysics funny?

src/lib/Scenes/Collection/Screens/CollectionArtworkPreview.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export class CollectionArtworkPreview extends React.Component<Props> {
+  render() {
+    let artworks
+    const { collection } = this.props
+    if (!collection) {
+      return null
+    }
+    const collectionArtworks = get(collection.artworks, artwork => artwork.edges, [])
```
pepopowitz: Does the `get` helper err out if the first argument is undefined? If so, or if you don't feel like looking, a quick way to avoid that happening here would be to move `.artworks` to the second argument, like this:

```suggestion
    const collectionArtworks = get(collection, collection => collection.artworks.artwork.edges, [])
```

src/lib/Scenes/Collection/Screens/CollectionArtworkPreview.tsx
```diff
@@ -0,0 +1,40 @@
...
+        <GenericGrid artworks={artworks} />
+      </>
+    )
+  }
+}
+
+export const CollectionArtworkPreviewContainer = createFragmentContainer(CollectionArtworkPreview, {
+  collection: graphql`
+    fragment CollectionArtworkPreview_collection on MarketingCollection {
+      artworks: artworksConnection(sort: "-merchandisability", first: 6) {
```
anandaroop: Should be `-decayed_merch` _n'est-ce pas_? Slack convo [here](https://artsy.slack.com/archives/C9SATFLUU/p1576685422107800)...

ashleyjelks: Yes, but I want to follow up with this in a sep PR bc it requires a change in Reaction and Emission and for now I think the data should be consistent.

src/lib/Scenes/Collection/Screens/__tests__/__snapshots__/CollectionHeader-tests.tsx.snap
```diff
@@ -1,64 +1,3 @@
...
-        },
-        Object {},
-      ]
-    }
-  >
-    &lt;p&gt;From &lt;a href="https://www.someartsyimage.net/artist/jean-michel-basquiat"&gt;Jean-Michel Basquiat&lt;/a&gt; to &lt;a href="https://www.someartsyimage.net/artist/shepard-fairey"&gt;Shepard Fairey&lt;/a&gt;, street artists are known for their creative expressions of rebellion.&lt;/p&gt;
-  </Text>,
-]
-`;
+exports[`renders properly 1`] = `"<view mb=\\"2\\"><aropaqueimageview imageurl=\\"http://imageuploadedbymarketingteam.jpg\\" height=\\"204\\" width=\\"750\\"></aropaqueimageview></view>"`;
```
pepopowitz: This looks different than before, but I still don't think it's capturing what you intended it to capture.

src/lib/Scenes/Collection/Screens/__tests__/__snapshots__/CollectionArtworkPreview-tests.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`renders properly 1`] = `"<view mb=\\"4\\" width=\\"100%\\"></view>"`;
```
pepopowitz: This looks different than before, but I still don't think it's capturing what you intended it to capture.

ashleyjelks: So this new snapshot is testing the Container and that the relay mock query generates the expected underlying view component. If you compare a similar snapshot in the Show Artworks preview, it's compiling to the same underlying view: https://github.com/artsy/emission/blob/344542288c6b2dbe85f3758224dae88cba853c10/src/lib/Components/Show/__tests__/__snapshots__/ShowArtworksPreview-tests.tsx.snap#L3

What do you think?

pepopowitz: I don't think it's a test worth keeping, if that's what the snapshot looks like. That's not going to catch any changes to the underlying component. I'd ditch the test, or find a way to test the underlying component itself.

src/lib/Scenes/Collection/Screens/__tests__/CollectionArtworkPreview-tests.tsx
```diff
@@ -0,0 +1,24 @@
...
+import React from "react"
+import { graphql } from "react-relay"
+import { CollectionFixture } from "../../Components/__fixtures__/CollectionFixture"
+import { CollectionArtworkPreviewContainer as CollectionArtworkPreview } from "../CollectionArtworkPreview"
+
+jest.unmock("react-relay")
+
+it("renders properly", async () => {
+  const tree = await renderRelayTree({
+    Component: (props: any) => <CollectionArtworkPreview collection={props} {...props} />,
```
xtina-starr: Do you we need the typing here? If so, it might be worth providing the exact type. `CollectionArtworkPreview_collection` I think???

src/lib/Scenes/Collection/Screens/__tests__/__snapshots__/CollectionArtworkPreview-tests.tsx.snap
ashleyjelks: I'm going to disable the test for now and follow up with @ds300 or @ashfurrow when they return next week from the break. In the meantime I think this PR is safe to merge!

artsyit:
:rocket: PR was released in v1.20.6 :rocket:

PR #2016, opened 2019-12-17
Updates shows to use coverImage

kierangillen:
- Displays the show's `coverImage` if there are now show images


artsyit:
:rocket: PR was released in v1.20.2 :rocket:

PR #2023, opened 2020-01-03
[AUCT-813] link to non-auction sale WIP

bhoggard:
I need help figuring out how to link the sale. It shows up as underlined text, but a click in the simulator does nothing.

Screen Shot 2020-01-03 at 3 25 54 PM


src/lib/Scenes/Artwork/Components/CommercialPartnerInformation.tsx
```diff
@@ -20,26 +25,31 @@ export class CommercialPartnerInformation extends React.Component<Props> {
...
-                <Sans size="3t" color="black60">
-                  {artwork.shippingInfo}
-                </Sans>
-              )}
-            {artworkEcommerceAvailable &&
-              !!artwork.priceIncludesTaxDisplay && (
-                <Sans size="3t" color="black60">
-                  {artwork.priceIncludesTaxDisplay}
-                </Sans>
+              {artwork.sale ? (
```
sweir27: Will this also show up for auction works? That doesn't seem ideal given we have other treatment for auction sales.

It would be worth adding a comment or logic so this _only_ works if the sale is not an auction.

Also, have you verified what the app's treatment of non-auction sales is? I'm not sure if it was designed to work for that use-case.

bhoggard: Good point. I believe we're not linking because the sale page doesn't look right for these. It shows number of bids, for example.

Image from iOS

bhoggard:
Closing this, as the app sale pages aren't designed for Buy Now sales.

PR #2027, opened 2020-01-09
Adds log out button to app

ashfurrow:
2020 Hackathon. See: https://github.com/artsy/eigen/pull/2977

PR #2028, opened 2020-01-09
Excludes mocks from requiring their own unit tests

ashfurrow:
We're seeing a Danger failure in this PR because I added a new mock, and Danger is insisting we write a unit test _for that mock_: https://github.com/artsy/emission/pull/2027

It's expecting that `src/lib/Scenes/Settings/__mocks__/LoggedInUserInfo.tsx` has a corresponding `src/lib/Scenes/Settings/__mocks__/__tests__/LoggedInUserInfo-tests.tsx` test file. This update excludes mocks from the Danger rule.

EDIT: Forgot to say that I tested this change locally using `DANGER_GITHUB_API_TOKEN=MY_PERSONAL_ACCESS_TOKEN yarn danger pr https://github.com/artsy/emission/pull/2027` and it works 👍


artsyit:
:rocket: PR was released in v1.21.2 :rocket:

PR #2030, opened 2020-01-10
Adds missing cursor to collection artwork pagination query

ashleyjelks:
The relay pagination container was missing the cursor required to fetch new artworks. This PR handles that bug fix.


artsyit:
:rocket: PR was released in v1.21.5 :rocket:

PR #2031, opened 2020-01-13
[MX-135] - Null checks for optional show.kind metadata

brainbicycle:
Added a null check similar to partner name check to prevent crash. 
This is what component looks like with showTypeString:
showComponentWith

and without:
showComponentWithout

### DIFF
--- a/src/lib/Components/Artist/ArtistShows/Metadata.tsx
+++ b/src/lib/Components/Artist/ArtistShows/Metadata.tsx
@@ -16,10 +16,11 @@ interface Props extends ViewProperties {
 class Metadata extends React.Component<Props> {
   render() {
     const partnerName = this.props.show.partner && this.props.show.partner.name
+    const showType = this.showTypeString()
     return (
       <View style={styles.container}>
         {!!partnerName && <Text style={styles.sansSerifText}>{partnerName.toUpperCase()}</Text>}
-        {!!this.showTypeString() && <Text style={styles.sansSerifText}>{this.showTypeString()}</Text>}
+        {!!showType && <Text style={styles.sansSerifText}>{showType}</Text>}
         <SerifText style={styles.serifText}>{this.props.show.name}</SerifText>
         {this.dateAndLocationString()}
         {this.statusText()}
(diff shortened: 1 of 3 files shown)


src/lib/Components/Artist/ArtistShows/Metadata.tsx
```diff
@@ -19,7 +19,7 @@ class Metadata extends React.Component<Props> {
     return (
       <View style={styles.container}>
         {!!partnerName && <Text style={styles.sansSerifText}>{partnerName.toUpperCase()}</Text>}
-        <Text style={styles.sansSerifText}>{this.showTypeString()}</Text>
+        {!!this.showTypeString() && <Text style={styles.sansSerifText}>{this.showTypeString()}</Text>}
```
ashfurrow: A minor piece of feedback, but instead of calling `showTypeString()` twice, we could store it in a local variable above the `return` statement. Purely a stylistic change, but feels nicer (to me, anyway).

brainbicycle: 👍

src/lib/Components/Artist/ArtistShows/__tests__/ArtistShow-tests.tsx
```diff
@@ -46,4 +46,30 @@ it("renders properly", () => {
...
+  }
+
+  const showNullKind = renderer
+    .create(
+      <Theme>
+        <ArtistShow show={showPropsNullKind as any} styles={showStyles} />
+      </Theme>
+    )
+    .toJSON()
+  expect(showNullKind).toMatchSnapshot()
```
ashfurrow: Nice test, thanks!

I think we should put this in its own `it` block with its own name, something like "it renders with a missing show kind" or something. I think there's also an opportunity to remove some duplicate code in the props we inject in.

If we move `showProps` up to the top-level in the test file, then we can use it to build out our `showPropsNullKind` props. Something like this:

```ts
const showPropsNullKind = {
  ... showProps, // this "spreads" out the showProps object so that showPropsNullKind has all the keys and values that showProps has
  kind: null // in the event that a key is repeated, the last one in the object "wins", so this effectively overwrites the `kind: "solo"` from showProps
}
```

The spread operator is one of my favourite language features from ES6 – [here's a blog post with more info](https://dev.to/torianne02/breaking-down-es6-spread-operator-k1m). Let me know what I can clarify!

brainbicycle: 👍 Good suggestions, let me know if this works!

artsyit:
:rocket: PR was released in v1.21.8 :rocket:

PR #2032, opened 2020-01-13
Featured artists module rebase

ashleyjelks:
The work for the Featured Artists module in iOS Collections is complete, it just needed a rebase against `master`  as many changes have been merged since the original `featured-artists-module` branch was created. I didn't want to push to @xtina-starr 's current branch so I created a new with the latest from `master` rebased against the work @xtina-starr has already completed in #2001  

Screen Shot 2020-01-13 at 1 01 58 PM
Screen Shot 2020-01-13 at 1 01 53 PM

PR #2033, opened 2020-01-14
Pass screen width in a variable to relay resize artworks query

ashleyjelks:
Follow up to #2032.

Instead of resizing the artworks images based on a static width of 500 px, this PR adds a variable to the Featured Artists Collections relay query to resize images based on the screen width.


src/lib/relay/QueryRenderers.tsx
```diff
@@ -280,24 +281,28 @@ interface CollectionRendererProps extends RendererProps {
 }
 
 export const CollectionRenderer: React.SFC<CollectionRendererProps> = ({ collectionID, render }) => (
-  <QueryRenderer<QueryRenderersCollectionQuery>
-    environment={environment}
-    query={graphql`
-      query QueryRenderersCollectionQuery($collectionID: String!) {
-        collection: marketingCollection(slug: $collectionID) {
-          ...Collection_collection
+  console.log(Dimensions.get("screen").width, "SCREEEEEEEEN WIDTH"),
```
pepopowitz: Rogue `console.log`. (I'd use the suggestion feature, but it is multiple lines and I don't think the suggestion feature supports that)

ashleyjelks: Good catch!

ashleyjelks:
> Very cool! This is just what I had in mind.
> 
> One thing that would make this really great is a unit test. [Here](https://github.com/artsy/emission/blob/22183b75cf09c9692bd6fa80d6b55b827a3e010c/src/lib/Scenes/Artwork/__tests__/hardware-tests.ts#L1-L3) is an example of how you can mock things in the `react-native` module (like `Dimensions` in this case) and [here](https://github.com/artsy/emission/blob/22183b75cf09c9692bd6fa80d6b55b827a3e010c/src/lib/Scenes/Search/__tests__/AutosuggestResults-tests.tsx#L162) is an example of how to test that a query variable is passed through. I wouldn't say that the unit test is _necessary_ before merging, since the code isn't _too_ complex, but the latter is an example of using [Relay's new Test Utils](https://relay.dev/docs/en/testing-relay-components), which we're trying to standardize on, so it could be useful as practice.

I think this is a great idea! I'll make a ticket to track the work.

artsyit:
:rocket: PR was released in v1.21.7 :rocket:

PR #2037, opened 2020-01-15
Exclude specified artists from featured artist list [FX-1717]

pepopowitz:
Collaboration with @anandaroop 

Addresses https://artsyproduct.atlassian.net/browse/FX-1717.

This PR adds the ability to remove specific featured artists from a collection. The feature [already exists in Reaction](https://github.com/artsy/reaction/blob/09c5a1b8b23b3b9ca0c1090c4c338784f57abb63/src/Apps/Collect2/Routes/Collection/Components/Header/index.tsx#L58) but needed to be moved over to Emission.


artsyit:
:rocket: PR was released in v1.21.8 :rocket:

PR #2042, opened 2020-01-16
Remove per-commit JS bundle options from Example app

ashfurrow:
Emission's Example app was originally intended to be used by PMs, designers, and stakeholders to see the absolute latest JS changes. The app could download the latest JS bundle, or a pull request bundle, or a commit bundle, from an S3 bucket.

The app hasn't been used like this for a while, for a few reasons:

- Emission's Example app wasn't high-fidelity enough to QA features. We're now opting to test only when bundled into an Eigen beta.
- Eigen's beta process is automated (after a manual kick-off)
- Emission library releases are now automated, further simplifying the Eigen beta deploy process.
- The [Emission Example app's automatic deploys](https://github.com/artsy/emission-nebula) have been broken for about eight months (see: https://github.com/fastlane/fastlane/issues/14798).

We have a bunch of infrastructure to support this old way of doing things, but have already deprecated the Example app automatic deploys. 

This PR removes infrastructure from the Example app to select a PR- or commit-specific JS bundle. It also removes the deploy job for these bundles (it was sending them to S3). Finally, it removes a Danger rule around the `package.json` `native-code-version`, which we relied upon to see if JS bundles were compatible with each other. 

One more small step towards merging Eigen+Emission!


artsyit:
:rocket: PR was released in v1.21.9 :rocket:

PR #2045, opened 2020-01-17
Fixes bug causing image to fail to render in collection header

ashleyjelks:
This PR fixes a bug where the collection header image URL was being reformated in Gemini both on the server and the client, thus creating an invalid URL.

[Links to FX-1742](https://artsyproduct.atlassian.net/browse/FX-1742)

Screen Shot 2020-01-17 at 3 23 03 PM

### DIFF
--- a/src/lib/Scenes/Collection/Screens/CollectionHeader.tsx
+++ b/src/lib/Scenes/Collection/Screens/CollectionHeader.tsx
@@ -24,7 +24,7 @@ export const CollectionHeader: React.SFC<CollectionHeaderProps> = props => {
   return (
     <>
       <Box mb={2}>
-        <OpaqueImageView imageURL={url} height={HEADER_IMAGE_HEIGHT} width={screenWidth} />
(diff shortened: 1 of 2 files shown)


artsyit:
:rocket: PR was released in v1.21.11 :rocket:

ashleyjelks:
> 👍
> 
> My understanding of this change is that we will no longer try to fetch a resized image. I'll merge this change, but should there be follow-up work to make it work with resized images? It seems like it would be a potential issue after launch if we are pulling in large images.

The image is resized in [Metaphysics](https://github.com/artsy/emission/blob/1543c66c792c45fa7d993de54287f4d600f850c2/src/lib/Scenes/Collection/Screens/CollectionHeader.tsx#L59) so we don't need to handle the resize a 2nd time in Emission. That boolean `useRawURL ` is what prevents the image component from requesting a resized url from Gemini.  [See OpaqueImageView](https://github.com/artsy/emission/blob/1543c66c792c45fa7d993de54287f4d600f850c2/src/lib/Components/OpaqueImageView/OpaqueImageView.tsx#L104).

PR #1781, opened 2019-08-08
Download pods from a cdn

sweir27:
🎉 🎉 

Followed the instructions here: http://blog.cocoapods.org/CocoaPods-1.7.2/

#trivial

PR #2046, opened 2020-01-17
Removes CocoaPods Specs repo setup

ashfurrow:
Since #1781, we have been downloading CocoaPods specs from the CDN instead of using the Specs repo. That means that this setup step is no longer required, but is still taking up CI time.

Also, Prettier made some changes to the yaml file, so, sure!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,12 +76,10 @@ jobs:
             - v9-pods-{{ checksum "package.json" }}
 
       - run:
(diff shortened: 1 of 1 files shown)


ashfurrow:
Going to test Peril, one sec... #mergeongreen

brainbicycle:
👍

artsyit:
:rocket: PR was released in v1.21.13 :rocket:

PR #2049, opened 2020-01-21
Remove nebula mentions from docs

brainbicycle:
Just some cleanup in docs now that nebula deploys are retired.


artsyit:
:rocket: PR was released in v1.21.16 :rocket:

PR #2050, opened 2020-01-21
[AUCT-744] iOS request condition report

bhoggard:
I don't think I can finish this PR this week, so I'm hoping @erikdstock or @yuki24 can take it over.

Screen Shot 2020-01-21 at 4 53 21 PM


brainbicycle:
Alright this is finally ready for review:

- Adds button to request condition report on biddable lots
- Adds a feature flag for this feature (also requires change in Eigen: https://github.com/artsy/eigen/pull/3011)


## Successfully requesting a condtion report 


requestCondtionReportSuccess


## Error on requesting a condition report


requestConditionError

**Notes:**
- The black background for the button loading state is not the desired behavior from palette: [Slack thread].(https://artsy.slack.com/archives/C9XJKPY9W/p1580313732000800) I am going to follow up with a PR there to fix this.

.circleci/config.yml
```diff
@@ -127,15 +127,15 @@ jobs:
...
-          echo 'nvm alias default v12.13.1' >> $BASH_ENV
+          name: Install Node 12 (Required for Yarn)
+          command: |
+            set +e
+            touch $BASH_ENV
+            curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.2/install.sh | bash
+            echo 'export NVM_DIR="$HOME/.nvm"' >> $BASH_ENV
+            echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
+            echo 'nvm install v12.13.1' >> $BASH_ENV
+            echo 'nvm alias default v12.13.1' >> $BASH_ENV
```
ds300: is this deliberate? I'm pretty sure this step is working in master, there was an indentation issue here a few weeks back but it was fixed.

.vscode/launch.json
```diff
@@ -0,0 +1,14 @@
...
+      "name": "Debug Jest Tests",
+      "type": "node",
+      "request": "launch",
+      "runtimeArgs": ["--inspect-brk", "${workspaceRoot}/node_modules/.bin/jest", "--runInBand"],
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
+      "port": 9229
+    }
+  ]
+}
```
ds300: :+1: good to have this available

src/lib/Components/Modal.tsx
```diff
@@ -1,4 +1,5 @@
 import { Button, Sans } from "@artsy/palette"
+import { TextAlignProperty } from "csstype"
```
ds300: cool, I didn't know about this package

src/lib/Scenes/Artwork/Components/ArtworkDetails.tsx
```diff
@@ -19,22 +21,34 @@ export class ArtworkDetails extends React.Component<ArtworkDetailsProps> {
     context_module: Schema.ContextModules.ArtworkDetails,
   }))
   render() {
+    const { artwork } = this.props
+
+    const enableLotConditionReport =
+      NativeModules.Emission &&
+      NativeModules.Emission.options &&
+      NativeModules.Emission.options.AROptionsLotConditionReport
```
ds300: we can use optional chaining here

```ts
const enableLotConditionReport = NativeModules.Emission?.options?.AROptionsLotConditionReport
```

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
```diff
@@ -0,0 +1,186 @@
...
+        <Button
+          mt={1}
+          size="small"
+          variant="secondaryGray"
+          loading={requestingConditionReport}
+          onPress={this.handleRequestConditionReportTap.bind(this)}
+        >
+          Request condition report
+        </Button>
+        <Flex height={0}>
```
ds300: is this Flex wrapper required? my understanding is that Modals do not affect layout

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
```diff
@@ -0,0 +1,186 @@
...
+    requestingConditionReport: false,
+    showConditionReportRequestedModal: false,
+    showErrorModal: false,
+    errorModalText: "",
+  }
+
+  requestConditionReport = () => {
+    const { artwork } = this.props
+    return new Promise(async (resolve, reject) => {
+      commitMutation<RequestConditionReportMutation>(defaultEnvironment, {
```
ds300: As a matter of hygiene, this should use the relay environment passed in to this component as a prop. See the example in this file: https://github.com/artsy/emission/blob/813a5705c5469f3cf8c4917470c069e658600e84/src/lib/Components/Artist/ArtistHeader.tsx#L125

Otherwise one day someone could change the environment passed into the QueryRenderer and then this component would be using a separate environment for mutations which could cause confusing issues since environments have independent caches and network layers etc.

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
```diff
@@ -0,0 +1,186 @@
...
+  showErrorModal: boolean
+  errorModalText: string
+}
+
+@track({
+  action_name: Schema.ActionNames.RequestConditionReport,
+  context_module: Schema.ContextModules.ArtworkDetails,
+})
+export class RequestConditionReport extends React.Component<RequestConditionReportProps, State> {
+  state = {
```
ds300:
```suggestion
  state: State = {
```
for arcane TypeScript reasons we need this annotation to make sure that the initial state value is correctly typed.

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
```diff
@@ -0,0 +1,186 @@
...
+  requestConditionReport = () => {
+    const { artwork } = this.props
+    return new Promise(async (resolve, reject) => {
+      commitMutation<RequestConditionReportMutation>(defaultEnvironment, {
+        onCompleted: data => {
+          resolve(data)
+        },
+        onError: error => {
+          reject(error)
+        },
```
ds300: i think these can be 

```ts
onCompleted: resolve,
onError: reject,
```

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
```diff
@@ -0,0 +1,186 @@
...
+          input: { saleArtworkID: artwork.saleArtwork.internalID },
+        },
+      })
+    })
+  }
+
+  @track({
+    action_type: Schema.ActionTypes.Fail,
+  })
+  presentErrorModal(errors: Error | ReadonlyArray<PayloadError>, mutationMessage: string) {
```
ds300: looks likie this `mutationMessage` value is always `null`, might as well delete it?

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
```diff
@@ -0,0 +1,186 @@
...
+  })
+  presentSuccessModal() {
+    this.setState({ showConditionReportRequestedModal: true, requestingConditionReport: false })
+  }
+
+  @track({
+    action_type: Schema.ActionTypes.Tap,
+  })
+  handleRequestConditionReportTap() {
+    this.setState({ requestingConditionReport: true })
```
ds300: Not a big deal, but to be tidy I think you could manage this state value entirely inside the `requestConditionReport` method rather than in here and then later in the `present*Modal()` methods. You could set it to true at the beginning of `requestConditionReport `, and then set it to false in a `finally` clause on the promise, before the promise is returned.

src/lib/Scenes/Artwork/Components/__tests__/ArtworkDetails-tests.tsx
```diff
@@ -58,4 +59,89 @@ describe("Artwork Details", () => {
...
+        framed: null,
+        series: null,
+        publisher: null,
+        manufacturer: null,
+        image_rights: "Scala / Art Resource, NY / Picasso, Pablo (1881-1973) © ARS, NY",
+      },
+    }
+    const component = mountArtworkDetails(artworkDetailsInfo.artwork)
+    expect(component.text()).toContain("Condition")
+    expect(component.text()).toContain("Amazing condition")
```
ds300: ✨

.circleci/config.yml
brainbicycle: Good callout, looks like this branch started off more behind master then I realized, this change was unnecessary after merging

src/lib/Scenes/Artwork/Components/ArtworkDetails.tsx
brainbicycle: 👍 I was using convention here: https://github.com/artsy/emission/blob/master/docs/developing_a_feature.md but I agree optional chaining is better. I'll update here and I want to follow up with some clean up for lab options in general including this, we have a bunch that we aren't using or are incorrectly defined in Eigen.

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
brainbicycle: No you are right that is definitely cleaner, I don't think I can use finally unfortunately because it was introduced in [Ecmascript2018](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-promise.prototype.finally) and we are using [2015](https://github.com/artsy/emission/blob/master/tsconfig.json#L7) I am wondering what consequences of upgrading are? That said I will consolidate to **catch** and **then** instead.

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
brainbicycle: nice! ✅

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
brainbicycle: Yep, good catch

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
ds300: ah didn't realise it was so new. `async` functions support `try {...} catch (...) {...} finally { ... }` but in this case it's probably not worth rewriting.

src/lib/Scenes/Artwork/Components/__tests__/ArtworkDetails-tests.tsx
```diff
@@ -58,4 +59,89 @@ describe("Artwork Details", () => {
     expect(component.text()).not.toContain("Frame")
     expect(component.text()).not.toContain("Signature")
   })
+
+  it("shows condition description if present and not biddable", () => {
+    const artworkDetailsInfo = {
```
dleve123: minor: but is it possible to provide a type to this fixture? I think we would want to type with `RequestConditionReport_artwork`

ds300: good call!

I think the preferred way to do this these days is to use the `@raw_response_type` annotation [(e.g.)](https://github.com/artsy/emission/blob/36878c33dfe49933fb02436a9d11ddc2641c558c/src/lib/Scenes/Collection/Components/__tests__/FeaturedArtists-tests.tsx#L111) on the test query and then index into the generated `*RawResponse` type [(e.g.)](https://github.com/artsy/emission/blob/36878c33dfe49933fb02436a9d11ddc2641c558c/src/lib/Scenes/Collection/Components/__tests__/FeaturedArtists-tests.tsx#L13).

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
brainbicycle: The margin on containing view in our implementation of the modal is messing with layout: https://github.com/artsy/emission/blob/master/src/lib/Components/Modal.tsx#L54 This seems like a bug so probably want to remove but doing so broke a couple layout tests for other components and seemed outside scope of this PR. So I went with this workaround instead. 

requestConditionReportGap

ds300: ah ok, makes sense then

src/lib/Scenes/Artwork/Components/__tests__/ArtworkDetails-tests.tsx
brainbicycle: 👍 I added a type for the fixtures, here the type is `ArtworkDetails_artwork` and there isn't a test query (I didn't write this original test). However I am happy to refactor to use a test query if you want just let me know!

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
```diff
@@ -0,0 +1,185 @@
...
+  state: State = {
+    requestingConditionReport: false,
+    showConditionReportRequestedModal: false,
+    showErrorModal: false,
+    errorModalText: "",
+  }
+
+  requestConditionReport = () => {
+    const { artwork, relay } = this.props
+    return new Promise(async (resolve, reject) => {
```
dleve123: @brainbicycle Is it possible to provide a type argument to the returned promise, [similar to the comparable code in Reaction](https://github.com/artsy/reaction/blob/master/src/Apps/Artwork/Components/ArtworkDetails/RequestConditionReport.tsx#L53)?

I believe that doing so means that the [`data` on L85](https://github.com/artsy/emission/pull/2050/files#diff-2ddabf7101faf8d9fc2cb4fdd4255203R85) could retain it type, leading to further type safety downstream

src/lib/Scenes/Artwork/Components/__tests__/RequestConditionReport-analytics-tests.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { mockTracking } from "lib/tests/mockTracking"
+import React from "react"
+
+jest.unmock("react-tracking")
+
+import { RequestConditionReport } from "../RequestConditionReport"
+
+jest.mock("lib/NativeModules/Events", () => ({ postEvent: jest.fn() }))
+
+const artwork = {
```
dleve123: I imagine that this test artwork could not be identity to the one in the above spec https://github.com/artsy/emission/pull/2050/files#diff-a27d3021ebaac1cf475edd4295f15fedR79

which would be more consistent, and also have the same type safety benefits.

brainbicycle: 👍 Good point, added to this and `RequestConditionReport-tests.tsx"

src/lib/Scenes/Artwork/Components/RequestConditionReport.tsx
brainbicycle: Yes it is ✅

artsyit:
:rocket: PR was released in v1.21.40 :rocket:

PR #2052, opened 2020-01-22
[FX-1729] tracking for collections

pepopowitz:
A collaboration with @anandaroop. 

Addresses https://artsyproduct.atlassian.net/browse/FX-1729 (whose ticket number represents [the smallest number expressible as the sum of two cubes in two different ways](https://en.wikipedia.org/wiki/1729_(number))).

This PR is all about tracking within collections. The events that are to be tracked:

- [x] A "screenview" when you view a collection (completed previously)
- [x] A "tap" when you tap "Read more" in the collection description (completed previously, but the constants are updated in this PR)
- [x] ~A "success" event when you follow an artist~ (addressed by existing component)
- [x] ~A "success" event when you unfollow an artist~ (addressed by existing component)
- [x] A "tap" when you tap on "View more" in the featured artists list


src/lib/Scenes/Collection/Components/__tests__/FeaturedArtists-tests.tsx
```diff
@@ -87,11 +90,11 @@ const CollectionFixture: FeaturedArtists_collection = {
 describe("FeaturedArtists", () => {
   const render = (collection: FeaturedArtists_collection) =>
     renderRelayTree({
-      Component: ({ marketingCollection }) => (
+      Component: mockTracking(({ marketingCollection }) => (
```
anandaroop: Using the `mockTracking` pattern described [here](https://github.com/artsy/emission/blob/724c929428f57000c31d4d14444a9db8deb39203/src/lib/tests/mockTracking.ts).

This seems a good fit for this use case: relay container around a a decorated class component with has a decorated tracking method — whew!

src/lib/utils/track/schema.ts
```diff
@@ -310,6 +310,11 @@ export enum ActionNames {
   // dispatch this on search input focus and again any time they type
   ARAnalyticsSearchStartedQuery = "Searched",
   ARAnalyticsSearchItemSelected = "Selected result from search screen",
+
+  /**
+   * Collection page events
+   */
+  ViewMore = "viewMore",
```
anandaroop: Pretty generic name that I could foresee used on other screens, but I don't think we have a "common" section or anything like that, so I've stuck it here.

src/lib/utils/track/schema.ts
```diff
@@ -328,18 +333,20 @@ export enum ContextModules {
...
   Bibliography = "Bibliography",
+  Collection = "Collection",
+  CollectionDescription = "CollectionDescription",
   CommercialButtons = "CommercialButtons",
   ExhibitionHistory = "ExhibitionHistory",
+  FeaturedArtists = "FeaturedArtists",
   PartnerContext = "PartnerContext",
   Provenance = "Provenance",
-  Collection = "Collection",
-  CollectionDescription = "CollectionDescription",
```
anandaroop: Seems like this list was previously alphabetized, restoring the status quo here.

src/lib/Scenes/Collection/Components/__tests__/FeaturedArtists-tests.tsx
```diff
@@ -174,6 +177,18 @@ describe("FeaturedArtists", () => {
...
+      const tree = await render(CollectionFixture)
+      const viewMore = tree.find("EntityHeader").filterWhere(x => x.text().includes("View more"))
+
+      viewMore.simulate("click")
+
+      expect(Events.postEvent).toHaveBeenCalledWith({
+        action_name: "viewMore",
+        context_module: "FeaturedArtists",
+        context_screen: "Collection",
+        flow: "FeaturedArtists",
```
anandaroop: This adds a test for the one brand new tracking event introduced in this PR.

I fought with trying to track the Collection `screenview` event for a bit, but gave up 😭

src/lib/Scenes/Collection/Screens/CollectionHeader.tsx
```diff
@@ -34,8 +34,8 @@ export const CollectionHeader: React.SFC<CollectionHeaderProps> = props => {
           <ReadMore
             content={collectionDescription}
             maxChars={screenWidth > 700 ? 300 : 250} // truncate at 300 characters on iPads and 250 on all other devices
-            contextModule={Schema.ContextModules.Collection}
-            trackingFlow={Schema.OwnerEntityTypes.Collection}
+            contextModule={Schema.ContextModules.CollectionDescription}
```
pepopowitz: Note for posterity that these values needed to be changed according to @abhitip's specs.

artsyit:
:rocket: PR was released in v1.21.24 :rocket:

PR #2058, opened 2020-01-24
[MX-96] Remove redundant screen view tracking for Artist page

ds300:
https://artsyproduct.atlassian.net/browse/MX-96

companion for artsy/eigen#2995 — removes one final redundant screen view event.

PR #2062, opened 2020-01-24
Fixes issue with Artist tracking components #trivial

ashfurrow:
This is a follow-up to https://github.com/artsy/emission/pull/2058 . Without decorating the top-level component with either `@screenTrack()` (which generates tracking events itself) or `@track()` (which exposes the tracking capabilities to the component tree, but doesn't itself track events), then the component tree will fail to render. I've verified this still keeps fixes the duplicate artist screen tracks 👍 

Going to self-merge so we have a workable beta for the weekend.


artsyit:
:rocket: PR was released in v1.21.23 :rocket:

PR #2063, opened 2020-01-27
reduce volley client error message noise in sentry

ds300:
Was going through sentry and noticed a lot of noise from these volley client errors. i don't know why volley is failing so often, but the noise is due to the stack trace being included in the error message rather than as metadata. Sentry identifies errors by the message so different messages means different errors. This PR should group them by http status.


artsyit:
:rocket: PR was released in v1.21.27 :rocket:

PR #2067, opened 2020-01-30
[WIP] Collections filter modal 

ashfurrow:
Worked on in Knowledge Share.

The initial modal overlay for the filter artworks component.

https://artsyproduct.atlassian.net/browse/FX-1765

PR #2068, opened 2020-01-30
[WIP] Adds toggling filter button for iOS Collections

ashleyjelks:
https://artsyproduct.atlassian.net/browse/FX-1765

Adds an unstyled filter button for opening filter artworks modal that only renders when collection artworks grid is within the viewport.

Kapture 2020-01-30 at 18 23 04

### DIFF
--- a/src/lib/Scenes/Collection/Collection.tsx
+++ b/src/lib/Scenes/Collection/Collection.tsx
@@ -1,11 +1,13 @@
 import { Box, Separator, Spacer, Theme } from "@artsy/palette"
 import { Collection_collection } from "__generated__/Collection_collection.graphql"
+import DarkNavigationButton from "lib/Components/Buttons/DarkNavigationButton"
 import { CollectionArtworksFragmentContainer as CollectionArtworks } from "lib/Scenes/Collection/Screens/CollectionArtworks"
 import { CollectionHeaderContainer as CollectionHeader } from "lib/Scenes/Collection/Screens/CollectionHeader"
 import { Schema, screenTrack } from "lib/utils/track"
 import React, { Component } from "react"
-import { FlatList } from "react-native"
+import { FlatList, View } from "react-native"
 import { createFragmentContainer, graphql } from "react-relay"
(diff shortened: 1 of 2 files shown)


src/lib/Scenes/Collection/Collection.tsx
```diff
@@ -66,13 +72,25 @@ export class Collection extends Component<CollectionProps, CollectionState> {
...
-    const { sections } = this.state
+    const { isArtworkGridVisible, sections } = this.state
 
     return (
       <Theme>
-        <Box>
+        <View style={{ flex: 1 }}>
           <FlatList
+            onViewableItemsChanged={this.onViewableItemsChanged}
+            viewabilityConfig={this.viewabilityConfig}
```
ds300: :+1: good find!

src/lib/Scenes/Collection/Collection.tsx
```diff
@@ -25,8 +28,11 @@ interface CollectionState {
 export class Collection extends Component<CollectionProps, CollectionState> {
   state = {
     sections: [],
+    isArtworkGridVisible: false,
+  }
+  viewabilityConfig = {
+    viewAreaCoveragePercentThreshold: 5,
```
anandaroop: Sooo in plain English this means the artwork grid must cover 5% of the viewport before it is considered visible? (I _think_ so, but it was a little hard to tell from the gif bc of that yellow alert.)

ashfurrow: I _think_ so. [I found the docs](https://facebook.github.io/react-native/docs/flatlist#viewareacoveragepercentthreshold) and it appears to be the case. Seems like we want the filter button to appear only when this is the case, which makes sense 👍 In any case, a short comment would be super-helpful here!

ashleyjelks:
@anandaroop @ds300 @ashfurrow I [combined this PR](https://github.com/artsy/emission/pull/2073) with #2067 so I'm going to close this for now. But will add a comment about the visibility percentage you mentioned here!

PR #2070, opened 2020-02-03
Fix sentry imports

ds300:
@ashfurrow regarding the deepZoom ticket, I noticed [this](https://sentry.io/organizations/artsynet/issues/1492672868/?project=166818&query=is%3Aunresolved) on sentry after our QA session, which told me that the sentry import was not right. These two files were using a default import whereas all the other imports of sentry (known to work) use the named import. Weirdly, both work at dev time but apparently default import doesn't work when bundled.

🤷‍♂ 

So I patched the typings to prevent this happening again.


artsyit:
:rocket: PR was released in v1.21.32 :rocket:

PR #2075, opened 2020-02-05
Updates Search query to allowed entities.

ashfurrow:
Follow-up from yesterday's MX Knowledge Share and this Metaphysics PR: https://github.com/artsy/metaphysics/pull/2160

This change essentially adds an allowlist to the search resolver, rather than rely on the defaults from Metaphysics. Let me know what you think!


artsyit:
:rocket: PR was released in v1.21.37 :rocket:

PR #2077, opened 2020-02-05
[FX-1787] Use a larger base image for collection header

pepopowitz:
A collaboration with @ashleyjelks. 

Addresses https://artsyproduct.atlassian.net/browse/FX-1787.

Some collections were showing pixelated header images - especially those whose first artwork had a portrait aspect ratio. We identified that we were using the `medium`-sized image as a basis for our image resizing, which resulted in the pixelation. 

In this PR we change the query to request the `larger`-sized image as a basis for image resizing. This results in higher quality header images for _all_ collections, but especially the ones that were previously pixelated.

We verified that if an artwork does not have a `larger`-sized image, it would fall back to `medium` without us having to do anything. 

### Before this change:

image

### After this change:

image


src/lib/Scenes/Collection/Screens/CollectionHeader.tsx
```diff
@@ -54,7 +54,7 @@ export const CollectionHeaderContainer = createFragmentContainer(CollectionHeade
         edges {
           node {
             image {
-              url
+              url(version: "larger")
```
anandaroop: How do the dimensions of the `larger` image compare to what we need to display in the app?

Asking in case we want to run this through a resizer as well. 

(But I recognize this might be complicated by variation in phone/tablet screen sizes and Retina, etc)

anandaroop: Answering my own question by peeking into a Gemini console:

```ruby
gemini:production> pp Version.find_by(key: "larger")
#<Version:0x000055bfd71a1c28
 id: 20,
 height: 1024,
 width: 1024,
 quality: 81,
 resize_to: "limit",
 key: "larger",
 description: "Larger Images",
 created_at: Mon, 06 Oct 2014 22:11:05 UTC +00:00,
 updated_at: Wed, 12 Jun 2019 18:02:59 UTC +00:00,
 convert_to: "jpg",
 resample: true,
 min_resampled_size: 1025,
 resampled_resolutions: [72, 150],
 strip_metadata: true,
 deep_zoom: false,
 skip_resize: false,
 extract_visual_genome: false,
 watermark: false>
```

anandaroop: This^ sounds reasonable to me. I'll go ahead and merge — we can follow up if anyone thinks that's necessary.  Thanks for the fix 🙌 !

pepopowitz: We settled on `larger` through experimentation, so I appreciate you digging deeper on this.

artsyit:
:rocket: PR was released in v1.21.36 :rocket:

PR #2078, opened 2020-02-06
Preserve changes made manually to the MP query map

ds300:
yesterday we needed to manually edit an existing query in MP's persisted query map. Unfortunately after deploying emission it overwrote that manual change because that's how emission updates the MP query map during it's deploy job on CI.

here we change the logic so that it preserves the MP version as the source of truth.

see artsy/metaphysics#2160 #2075 and https://artsy.slack.com/archives/CN8S32FJR/p1581001886215700 for context


ashfurrow:
Travis is 🍏 but hasn't reported back yet. Merging.

artsyit:
:rocket: PR was released in v1.21.38 :rocket:

PR #2079, opened 2020-02-06
[MX-160] Prevent crash when camera roll permission denied

ds300:
https://artsyproduct.atlassian.net/browse/MX-160

### DIFF
--- a/src/lib/Components/Consignments/Screens/SelectFromPhotoLibrary.tsx
+++ b/src/lib/Components/Consignments/Screens/SelectFromPhotoLibrary.tsx
@@ -52,9 +52,8 @@ export default class SelectFromPhotoLibrary extends React.Component<Props, State
     }
   }
 
-  loadPhotos() {
-    // TODO: Need to update the RN 0.48 types on DT
-    const fetchParams: any = {
+  async loadPhotos() {
+    const fetchParams: CameraRoll.GetPhotosParams = {
       first: 20,
       assetType: "Photos",
       groupTypes: "All",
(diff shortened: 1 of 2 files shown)


src/lib/Components/Consignments/Screens/SelectFromPhotoLibrary.tsx
```diff
@@ -52,9 +63,8 @@ export default class SelectFromPhotoLibrary extends React.Component<Props, State
     }
   }
 
-  loadPhotos() {
-    // TODO: Need to update the RN 0.48 types on DT
-    const fetchParams: any = {
+  async loadPhotos() {
+    const fetchParams: CameraRoll.GetPhotosParams = {
```
ashfurrow: 🎉

src/lib/Components/Consignments/Screens/SelectFromPhotoLibrary.tsx
```diff
@@ -188,12 +202,7 @@ export default class SelectFromPhotoLibrary extends React.Component<Props, State
       <Theme>
         <ProvideScreenDimensions>
           <BottomAlignedButton onPress={this.doneTapped} buttonText="Done">
-            <ScrollView
-              style={{ flex: 1 }}
-              scrollsToTop={true}
-              onScroll={this.onScroll.bind(this)}
```
ashfurrow: I still haven't quite internalized the distinction between class functions and class properties that _are_ functions. So I see we changed `onScroll`:

```diff
-  onScroll(e) {
+  onScroll = (e: NativeSyntheticEvent<NativeScrollEvent>) => {
```

Is there a strong preference for one or the other, generally? Or does it depend on a case-by-case basis? Seems like not having to `.bind(this)` all the time would make `onScroll = ` preferred. I'd appreciate any clarity around nomenclature that I've gotten wrong, too 🙇

ds300: > Seems like not having to .bind(this) all the time would make onScroll = preferred.

yep. Forgetting .bind is something i've seen cause subtle bugs occasionally.

artsyit:
:rocket: PR was released in v1.21.39 :rocket:

PR #2081, opened 2020-02-07
[FX-1748] Fix line-height in ReadMore component

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-1748

This simply updates `ReadMore` to use font size `3` instead of `3t` (i.e. the <b>t</b>ightly-spaced variant)

### Phone

banda

### Tablet

ipadbanda

### Background

I was curious why this component was spec'd with `3t` size to begin with, so I did the Github and Jira archaeology to find:

- https://github.com/artsy/emission/pull/1654, which implements:

- https://artsyproduct.atlassian.net/browse/PURCHASE-1139, which includes the comp:

- https://app.zeplin.io/project/5ab94282c124db8d58365b47/screen/5cdb43f2b4655d1e7436ca1a

From there I can see that this component was 16/24 serif which does in fact [map to size 3](https://www.notion.so/artsy/Typography-42e0e8e4f4e34ccc84c7de7adcd1b519)

I figure this was an oversight that just persisted till now, so I think this change should be good to go, and @jeffreytr agrees!

### DIFF
--- a/src/lib/Components/ReadMore.tsx
+++ b/src/lib/Components/ReadMore.tsx
@@ -31,7 +31,7 @@ export const ReadMore = React.memo(
         ...basicRules.paragraph,
         react: (node, output, state) => {
           return (
-            <TextComponent size="3t" color={color || "black100"} key={state.key}>
+            <TextComponent size="3" color={color || "black100"} key={state.key}>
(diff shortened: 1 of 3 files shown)


ashleyjelks:
Great find!

artsyit:
:rocket: PR was released in v1.21.41 :rocket:

PR #2086, opened 2020-02-12
[MX-161] Return early if event data doesn't make sense

ds300:
https://artsyproduct.atlassian.net/browse/MX-161

Tiny bugfix for a rare situation where the event data we get after a gesture ends doesn't seem to make sense.


brainbicycle:
LGTM 👍

artsyit:
:rocket: PR was released in v1.21.47 :rocket:

PR #2093, opened 2020-02-18
Enables incremental compiler option #trivial

ashfurrow:
@damassi brought this up on Slack and it seems like a great idea: https://artsy.slack.com/archives/C02BC3HEJ/p1581803473357200

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,6 +4,7 @@
     "esModuleInterop": true,
     "baseUrl": "src",
     "experimentalDecorators": true,
+    "incremental": true,
     "lib": ["es2015"],
     "jsx": "react",
     "keyofStringsOnly": true,


ds300:
This unfortunately won't affect us because we use `tsc` with the `noEmit` option :(((

See here for details: https://github.com/microsoft/TypeScript/issues/30661

The only workaround people have found so far is to turn off `noEmit` and dump the compiler output somewhere out-of-the-way.

For me, `tsc` runs in 10s and pretty much only ever as part of the prepush hook, so I don't mind putting up with this until the TS team finds a solution. Maybe we can nudge our friend on the inside? 🙊 

If this is a big deal for other devs then I'm onboard to try out the workaround. What do you think?

ashfurrow:
Gotcha, thanks for that context @ds300 👍 Orta is focused on the TypeScript website at the moment, and based on how the TS team is structured, I don't think we'd be able to nudge the team very far right now. 

This wasn't a big deal for me, it only seemed like a quick win. I'm fine closing – thanks for the context! 🙏

